PR #93, opened 2025-10-18
fix: refactor HeldItemSaveData::instantiate to return Result

tommy-xr:
## Summary
- Resolves TODO: "Refactor return value to result" in `held_item_save_data.rs:24`
- Improves error handling in save/load system by replacing panic-prone `.unwrap()` calls
- Adds proper error types and comprehensive test coverage

## Key Changes
- **Error Type**: Added `HeldItemError::InvalidEntityId` for invalid entity ID cases
- **Return Type**: Changed `instantiate()` from returning tuple to `Result<(Option<EntityId>, Option<EntityId>, Option<EntityId>), HeldItemError>`
- **Error Handling**: `EntityId::from_inner()` failures now return proper errors instead of panicking
- **Caller Updates**: Modified `mission/mod.rs` to handle Result gracefully with fallback behavior

## Technical Details
- Uses `EntityId::from_inner().ok_or(HeldItemError::InvalidEntityId(id))?` pattern
- Maintains backward compatibility through `unwrap_or_else` with warning log
- Error type implements `std::error::Error` and `Display` traits

## Test Coverage
- **Red/Green TDD**: Verified test fails before fix, passes after
- `test_instantiate_with_valid_entities`: Verifies successful operation
- `test_instantiate_with_invalid_entity_fails`: Verifies error handling for invalid entity ID (0)
- `test_instantiate_with_no_entities`: Verifies empty case handling

## Impact
- **Robustness**: Prevents potential panics when save data contains invalid entity IDs
- **Debugging**: Provides meaningful error messages for invalid entity cases
- **Maintenance**: Removes TODO comment and improves code quality

This change makes the save/load system more resilient to corrupted or invalid save data.

### DIFF
--- a/shock2vr/src/save_load/held_item_save_data.rs
+++ b/shock2vr/src/save_load/held_item_save_data.rs
@@ -3,6 +3,21 @@ use shipyard::{EntityId, World};
 
 use super::EntitySaveData;
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum HeldItemError {
+    InvalidEntityId(u64),
+}
+
+impl std::fmt::Display for HeldItemError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            HeldItemError::InvalidEntityId(id) => write!(f, "Invalid entity ID: {}", id),
+        }
+    }
+}
+
+impl std::error::Error for HeldItemError {}
+
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct HeldItemSaveData {
     pub held_entities: EntitySaveData,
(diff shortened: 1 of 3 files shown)


shock2vr/src/save_load/held_item_save_data.rs
```diff
@@ -21,39 +36,110 @@ impl HeldItemSaveData {
         }
     }
 
-    // TODO: Refactor return value to result
     pub fn instantiate(
         &self,
         world: &mut World,
-    ) -> (Option<EntityId>, Option<EntityId>, Option<EntityId>) {
+    ) -> Result<(Option<EntityId>, Option<EntityId>, Option<EntityId>), HeldItemError> {
```
tommy-xr, line 67: This doesn't actually seem any better. The point of switching to a `Result` is that this could be now be `Result<(EntityId, EntityId, EntityId), HeldItemError>` - removing the nested options. In other words, we should be able to remove all or nothing.

tommy-xr: Thanks for pointing this out! The hands and inventory slots can legitimately be empty when we serialize, so we still need the outer Option to represent "nothing equipped". To make the Result meaningful, I added explicit remapping checks: if the save data claims an entity but we fail to remap it, we now bail with a slot-specific error instead of silently returning None. That keeps the valid empty-hand cases but catches the inconsistent ones.

PR #286, opened 2026-06-17
fix(ci): remove dead Shodan build/test steps

tommy-xr:
## Why

CI is failing on every branch (e.g. #285 on all 3 platforms) at the **"Build Shodan CLI"** step.

PR #272 deleted the `tools/shodan` crate and removed it from the workspace `Cargo.toml`, but it left the corresponding steps in `.github/workflows/build.yml`:

```yaml
- name: Build Shodan CLI
  run: cargo build --verbose --release -p shodan
- name: Run Shodan CLI tests
  run: cargo test --verbose -p shodan
```

Since the `shodan` package no longer exists, `cargo ... -p shodan` errors out, and the job dies before it ever reaches `dark_viewer`, `dark_query`, `debug_runtime`, etc.

## What

- Remove the two dead Shodan build/test steps from `build.yml`.

No source or behavior change — this only deletes a leftover CI step. `git grep shodan` over `.github/` and `Cargo.toml` is now clean.

## Impact

- #285 will pass after a rebase / re-run.
- #284 (currently "no checks" because its head commit is markdown-only and the workflow has `paths-ignore: '**/*.md'`) will be fine on its next code push.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,16 +64,6 @@ jobs:
         env:
           RUSTFLAGS: "-D warnings"
 
-      # Build and test Shodan CLI tool
-      - name: Build Shodan CLI
-        run: cargo build --verbose --release -p shodan
-        env:
-          RUSTFLAGS: "-D warnings"
-      - name: Run Shodan CLI tests
-        run: cargo test --verbose -p shodan
-        env:
-          RUSTFLAGS: "-D warnings"
-
       # Build and test dark_viewer tool
       # Note: Windows build for dark_viewer is currently disabled due to compilation issues
       # TODO: Fix dark_viewer Windows compatibility and re-enable
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -334,7 +334,7 @@ dependencies = [
  "quote",
  "regex",
  "rustc-hash 1.1.0",
- "shlex",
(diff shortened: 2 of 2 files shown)


tommy-xr:
## Second CI fix: Windows / Visual Studio 18

After the Shodan-step removal, Windows still failed — a separate, environment-driven break:

```
cmake-0.1.50/src/lib.rs:955: couldn't determine visual studio generator
```

The `windows-latest` runner image now ships **Visual Studio 18** (`C:\Program Files\Microsoft Visual Studio\18\Enterprise`, MSVC 14.51). `glfw-sys` builds through the `cmake` crate, and pinned `cmake 0.1.50` only maps VS generators up to 17 (2022), so it panics on VS 18 and every Windows build dies.

**Fix:** `cargo update -p cmake --precise 0.1.58` (lockfile-only; `cmake` is transitive via `glfw → glfw-sys`). `0.1.58` adds `VsVers::Vs18 => "Visual Studio 18 2026"`, and the update pulls `cc 1.2.45 → 1.2.64` + `find-msvc-tools 0.1.9` which provide the `Vs18` enum (`"18.0" → VsVers::Vs18`).

Verified the full enum chain in the resolved crates and that `desktop_runtime` still builds locally. The Windows generator path can only be confirmed by CI on this runner.

PR #334, opened 2026-06-29
Save created APK

lvonasek:


lvonasek:
I love to see this project is progressing. I was considering to add CI to have the APK builds available but I see it is too early for it.

Keep up the great work👍

tommy-xr:
Hi @lvonasek , thank you for the PR and the kind words! 

Still a ways from being playable but hopefully will get there eventually 😅 

Big fan of your work in VR - need to try out the CS 1.6 port!

Issue #386, opened 2026-07-04
launch-failure e2e assertion out of sync with new launch error format


tommy-xr:
`test/launch-failure.e2e.test.ts` asserts `/exited early with code/`, but #369 changed the launch failure message in `src/server.ts:325` to `process exited early (code N, signal S)`. The test now fails on main:

```
not ok 6 - launching a crashing mission rejects with the runtime's output
error: 'error should report the process exit'
actual: debug_runtime failed to start: Error: process exited early (code 101, signal null)
```

One-line fix: relax the assertion to `/exited early \(code/` (or `/exited early/`). Filing instead of fixing since the SDK launch path has active in-flight work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

tommy-xr:
Fixed by #378 (duplicate of #374): `test/launch-failure.e2e.test.ts` now asserts `/process exited early \(code/`, matching the reworded launch-failure message.

PR #387, opened 2026-07-04
fix(ai): scripted sequences survive alertness changes; Play waits for its clip

tommy-xr:
Diagnoses and fixes the rec1 "entity tapping on the window" opening scene varying every round.

**The scene** is fully data-authored: the spawn-side tripwire signals `MaleRec`, whose `PropAISignalResponse` runs *frob button → goto the window waypoint → play `cs 152` (the `cortez` tap/gesture clip) → goto "runaway" → frob `killcrew`*.

**Why it varied**:
- Any alertness level change unconditionally replaced `current_behavior` and queued a new animation over the playing clip — so the moment the actor noticed the player (he performs at a window you're standing behind), the show was cancelled at a player-dependent point and he fell into unseeded-RNG wander. Two identical-input runs are byte-identical; all observed variance was player-visibility-coupled or RNG-after-interruption.
- `Play` actions reported complete immediately, relying on the accident that sequence advancement is only polled on `AnimationCompleted`; any interruption path (damage react, alertness re-queue) could skip the beat entirely.

**Fixes**:
- Running scripted sequences are protected from alertness behavior swaps and animation stomps; alertness keeps updating in the background and applies when the sequence finishes (finished sequences now hand back to the alertness behavior instead of lingering forever). The original engine's response `priority` gating is noted as future refinement.
- `PlayAnimationScriptedAction` explicitly waits for its clip via a newly wired `AnimationCompleted` → `Behavior::handle_message` path (the hook existed, was never called). Motion-query failures also dispatch `AnimationCompleted`, so the wait can't deadlock.
- Permanent `debug!` breadcrumb for animation query resolutions — silent failures there have now hidden three separate bugs.

**Verification**: full Cortez sequence completes with the player present (walk in → ~14s tap/gesture performance → sprint to "runaway" → despawn), confirmed via trajectory sampling + animation-query logs in the debug runtime. Full e2e: 51/52 — the one failure is pre-existing on main (#386, stale assertion vs #369's new launch-error format).

### DIFF
--- a/shock2vr/src/mission/mission_core.rs
+++ b/shock2vr/src/mission/mission_core.rs
@@ -2434,40 +2434,54 @@ impl MissionCore {
                     );
                 }
 
-                Effect::SetAIProperty { entity_id, update } => match update {
-                    AIPropertyUpdate::Alertness { level, peak } => {
-                        self.world
-                            .add_component(entity_id, PropAIAlertness { level, peak });
-                    }
-                    AIPropertyUpdate::Mode { mode } => {
-                        self.world.add_component(entity_id, PropAIMode { mode });
-                    }
-                    AIPropertyUpdate::Behavior { name } => {
-                        self.world
-                            .add_component(entity_id, RuntimePropAIBehavior(name));
-                    }
-                    AIPropertyUpdate::TargetAwareness {
-                        last_known_pos,
-                        has_line_of_sight,
-                    } => {
-                        self.world.add_component(
-                            entity_id,
-                            crate::runtime_props::RuntimePropAITargetAwareness {
+                Effect::SetAIProperty { entity_id, update } => {
+                    // The AI may have been destroyed earlier in this same
+                    // effect batch (e.g. a scripted sequence whose final
+                    // action frobs a slay trap, while the AI also published
+                    // a behavior change this frame); adding a component to a
+                    // dead entity panics.
+                    let is_alive = self
+                        .world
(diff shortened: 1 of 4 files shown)


tommy-xr:
Cross-engine review (Claude subagent + Codex) applied in `aef19c3`:

- **[both engines] Stall-proofing**: per-action 30s watchdog + `AnimationCompleted` dispatch on clip-load failure — protecting sequences from alertness removed the accidental failsafe, so a stalled `Goto` (no timeout) could previously wedge the AI forever.
- **[both engines] Complete protection**: nonlethal damage escalation, re-fired `Signal`/`TurnOn`, and `AIWatchObj` proximity no longer cancel or restart a running sequence. Re-trigger verified in rec1: walking out/into the tripwire mid-performance yields exactly one `cs 152` query and one complete scene.
- Removed a duplicate behavior-message forward (the hook was already wired at the top of `handle_message`; the first commit's premise was wrong), the final action's `completion_effect` is no longer dropped (with a repeat-completion guard), and the finished-sequence handback moved to `update()` so it fires even when no further animation events arrive.

Remaining follow-ups noted by review, intentionally out of scope: generation-tagged `AnimationCompleted` to close a narrow stale-completion race at sequence start, honoring the authored response `priority` for interruption semantics, and an SDK e2e pinning the behavior. Full suite after fixes: 51/52 (the 1 failure is pre-existing on main, #386).

PR #391, opened 2026-07-04
fix(sdk-test): match the launch-failure error wording from #376

tommy-xr:
The full e2e suite has been failing on `launch-failure.e2e.test.ts` since #376: the SDK's early-exit error message became `process exited early (code N, signal S)` but the test still asserts `/exited early with code/`. One-line regex update.

Verified: `SHOCK2_E2E=1 node --test dist/test/launch-failure.e2e.test.js` fails before, passes after.


tommy-xr:
Superseded by #378, which landed the identical regex fix (`/process exited early \(code/`) as part of the launch-retry work — nothing left for this PR to do.

PR #460, opened 2026-07-11
refactor(physics): capsule player controller with explicit step probe — delete the two-pass move hack

### LINKED ISSUE #457: earth.mis: tram colliders block walking out of the tram (exit to gravshaft impossible without the old wall-climb exploit)
## Symptom

In `earth.mis`, the player spawns inside the UNN tram and cannot walk out: forward movement hard-blocks at z ≈ -11.1..-10.9 on an **invisible thin barrier** (top ~2.1 SS2 ft above the cabin floor, thinner than ~1 ft). The intended intro path (tram → boardwalk → gravshaft at `(≈10-12, 12.2, 14.4)`, recruitment doors at y≈24) is unreachable by walking.

On `main` this appears to work, but only via two stacked physics artifacts of the old character controller (removed by the capsule/step-probe refactor):
1. The two-pass move's per-frame up-bump (`MOVEMENT_STEP_SIZE * dt` ≈ 0.33 wu) let the flat-bottomed cuboid **ratchet up thin ledges of any height** (the box can rest on a thin rail top; a capsule slides off).
2. After climbing the barrier, the player walks to the gravshaft at feet ≈ 1.27 wu — **1.5 wu above the actual plaza floor** (world-geometry floor is at y = -1.2 → feet -0.24). i.e. the player crosses the plaza standing on invisible, misbuilt entity colliders, not the real floor.

## Root cause (tram entity colliders)

The tram is 7 entities (`Large Tramcar` + `Large Tram Front/Back/LS Front/LS Back/RHS/Roof`, all kinematic at ≈ (10.15-10.8, 2.1-2.3, -10.4)).

- `Large Tramcar` (template -3432, inherits **Moving Terrain**) has authored physics
  `PropPhysType: ORIENTED_BOUNDING_BOX`, `PropPhysDimensions { size: (-4.8, 0.4, 9.6), offset0: (0, -1.8, 0) }`
  — i.e. the original engine's tram physics is a **thin floor slab** offset 1.8 ft below the object center; the visible shell/seats are non-physical.
- In `entity_creator.rs` (`shock2vr/src/mission/entity_creator.rs:851-888`), the OBB path does
  `PhysicsShape::Cuboid(abs(size) * scale)` and computes `_offset_rotation` from `offset0` but **never applies it** (underscore-dropped) — so the slab is built centered on the object, not 1.8 ft down.
- The shell-piece entities also get their own colliders, producing the solid mid-cabin blockers and the elevated surfaces the old controller walked on.
- The tram is Moving Terrain (`TPathInit → Marker`), but it never moves in a fresh session (verified: 10 s no-input, tram body static on both main and the refactor branch) — so any "ride the tram" behavior is also unimplemented/untriggered.

## Suggested fix

1. Apply `offset0` (rotated by the entity rotation) as the collider's local translation in the OBB path of `entity_creator.rs`.
2. Audit which of the tram sub-entities should be physical at all — per the original data, Moving Terrain physics is just the slab; the shell pieces likely need no colliders (or trimesh, not bbox).
3. (Follow-up) Moving-terrain support so the tram can actually carry the player, and gravshaft lift behavior.

## Repro

```
cargo dbgr --mission earth.mis --port 8093
curl -X POST :8093/v1/control/input -d '{"right_hand.thumbstick": [0.0, 1.0]}'
curl -X POST :8093/v1/step -d '{"frames": 180}'
curl :8093/v1/player/position   # stuck at z ≈ -10.95 (capsule branch)
```

Found while verifying the character-controller capsule/step-probe refactor (`refactor/player-capsule-autostep`); see `projects/character-controller-review.md` for the full investigation notes.

tommy-xr:
## Summary

First slice of the character-controller cleanup (full review + plan in `projects/character-controller-review.md`, comparing our controller against the original engine's player physics). This slice replaces the player's collider shape, deletes the legacy movement hacks, and makes step climbing an explicit, documented mechanism — plus fixes a long-standing walking freeze this work uncovered (#457).

### What changed

- **Cuboid → capsule** (same 4.8 × 1.6 ft footprint; resizing toward the original's 6.0 × 2.4 player is deferred until crouch can shrink the collider). The original player is a stack of spheres — a rounded shape slides along the corners the box snagged on.
- **Deleted the two-pass move hack** (`MOVEMENT_STEP_SIZE` up-bump + gravity-undo, a rapier-0.19-era workaround). The up-bump also let the flat-bottomed box **ratchet up thin ledges of any height** — e.g. on `main`, "walking out of the earth.mis tram" actually climbs an invisible barrier and crosses the plaza hovering above the floor.
- **Explicit step probe** (`try_step_up`): when grounded walking is blocked, cast up (headroom) → forward (clearance past the riser) → down (find the tread). Max step = **2 SS2 ft**, matching the original's step probe; `snap_to_ground` at the same distance handles walking *down* stairs. Rapier's built-in autostep can't work here: a capsule touching a step edge above its bottom-sphere center classifies as a ceiling-ish contact, never a wall, so it never fires (verified against rapier 0.31 source + empirically).
- **Deleted `normal_nudge_factor = 0.1`** (1000× rapier's default): proven dead — it never fixed the resting-contact freeze below and broke stair-stepping in single-pass mode. Its historical effect was accidental synergy with the up-bump.
- **Fixed a resting-contact freeze** (#457): the player rests exactly `contact_offset` above its support, so every movement cast starts within the cast's target distance of its own floor. On a **yaw-rotated** cuboid (the earth.mis tram's Moving Terrain floor slab — the 90° quaternion is irrational in f32) the rotated top-face normal carries ~1e-6 tilt, making tangential walking read as "approaching": every solver iteration re-hits the same contact at toi=0 and the player freezes after one frame. Fix: walk and gravity as **separate `move_shape` passes** (standard move-then-fall, no up-bump) + **`PLAYER_REST_LIFT`** (0.01 SS2 ft per grounded frame) keeping the resting gap a hair above the cast target. Axis-aligned floors yield an exact (0,1,0) normal (tangential motion reports no hit), which is why everything else always worked.
- Dead-code cleanup surfaced by the investigation (stale pre-0.31 impulse block, misleading `_offset_rotation` in `entity_creator`).

### Behavior changes to be aware of

- The player can no longer climb arbitrary walls by pushing into them (up-bump ratchet is gone); steps over 2 ft honestly block until jump/mantle land (next slices).
- `snap_to_ground` = 2 ft: walking off ledges ≤ 2 ft snaps down instead of falling — documented tradeoff, revisited when gravity becomes an integrated velocity.
- The player stands at true ground height (the old controller hovered up to one bump above its support).

## Verification

- **Negative-first unit tests** (each fails on the old controller / without its fix):
  - `player_steps_up_stairs_but_not_tall_ledges` — 1.5 ft riser climbs (old max was ~0.8 ft), 3 ft ledge blocks
  - `player_walks_on_rotated_platform` — freezes without `PLAYER_REST_LIFT`
  - `player_climbs_climbable_wall_but_not_plain_wall` — ladder grip unchanged with the capsule
- **New e2e**: `earth-tram-exit.e2e.test.ts` — walks from the earth.mis tram spawn out the door, across the boardwalk, to the gravshaft base (frozen after ~1 frame before the fix)
- **Full SDK e2e suite green (82/82)** on the core capsule commit, including the every-mission load smoke test, real-mission ladder climbing, and validated player moves; final full-suite run on HEAD in flight (will confirm in a comment)
- 121 unit tests, `RUSTFLAGS="-D warnings"` check on shock2vr/desktop_runtime/debug_runtime
- Cross-engine adversarial review (Claude + Codex) — all agreed findings addressed (grounded-gated step probe, horizontal-only climb reach, slope-limit-consistent tread threshold)
- Interactive verification via debug runtime: medsci1 walking/wall-stops/falls, earth.mis tram→gravshaft traversal (the fix for #457 was root-caused with per-cast hit instrumentation, body COM inspection, and an A/B against a main worktree)

## Follow-ups (tracked in `projects/character-controller-review.md`)

1. Gravity as integrated velocity (32 ft/s²) — unlocks jump/mantle/fall damage
2. Jump (14 ft/s) — also needed to enter the earth.mis gravshaft (~3 ft lip)
3. Authentic movement speeds (walk 5.5 / run 11 vs our flat 25 ft/s)
4. Real crouch (collider shrink + speed + fire-ray height)
5. Gravshaft room-gravity lift doesn't fire; tram TPath never runs (separate features)

Fixes #457

### DIFF
--- a/shock2vr/src/mission/entity_creator.rs
+++ b/shock2vr/src/mission/entity_creator.rs
@@ -846,7 +846,6 @@ pub fn create_physics_representation(
             v_phys_type.get(entity_id),
         ) {
             let qrotation = pos.rotation;
-            let _offset_rotation = qrotation.rotate_vector(dimensions.offset0);
 
             let mut is_sensor = false;
             let scale_factor = v_scale
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -22,10 +22,24 @@ use physics_events::*;
 
 use self::debug_render_pipeline::DebugRenderer;
 
-const MOVEMENT_STEP_SIZE: f32 = 20.0;
+/// Player capsule dimensions (SS2 ft): total height and radius. Kept at the
+/// old cuboid's footprint (4.8 tall, 1.6 wide) - resizing toward the original
+/// engine's 6.0 x 2.4 player is deferred until crouch can shrink the collider.
+const PLAYER_HEIGHT: f32 = 4.8;
+const PLAYER_RADIUS: f32 = 0.8;
+
+/// Gap (SS2 ft) the character controller keeps between the player collider
+/// and world geometry (`KinematicCharacterController::offset`).
+const PLAYER_CONTACT_OFFSET: f32 = 0.1;
+
+/// Maximum ledge height (SS2 ft) the player steps up automatically, and how
+/// far below their feet the ground is snapped to when walking down. 2 ft is
+/// the original engine's step-probe height, so stairs climb and descend
+/// without a jump.
+const PLAYER_STEP_HEIGHT: f32 = 2.0;
 
 /// Horizontal reach (world units) for ladder detection: the player grips a
-/// climbable surface when their collider, inflated by this much on x/z,
+/// climbable surface when their collider, inflated by this much radially,
 /// overlaps it. Standing flush against a ladder leaves a small gap between the
 /// collider and the rungs, so the un-inflated shapes never intersect.
 const CLIMB_REACH: f32 = 1.0 / SCALE_FACTOR;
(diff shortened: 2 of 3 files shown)


tommy-xr:
Final full SDK e2e suite on HEAD: **83/83 pass** (the 82 existing tests — including the every-mission load smoke suite — plus the new `earth-tram-exit` regression test). ~15 min wall clock, zero failures.

PR #469, opened 2026-07-11
fix(anim): engage authored blends from a drained queue, cosine easing

tommy-xr:
## Summary

Fix #2 of the animation-smoothness stack (stacked on #464). Two changes in `AnimationPlayer`:

1. **Authored blends now engage at normal clip seams.** `queue_animation` only blended when it interrupted a *playing* clip — but AI clips virtually never interrupt: the completion handler queues the next clip one tick after the previous finishes, so the queue is drained and the authored `blend_length` was silently skipped. It now falls back to blending from `last_animation`'s final frame (the exact pose `get_transforms` was showing), the same fallback `play_animation` already had. Idle clips author 500 ms blends that never ran — every idle clip change hard-popped up to 1.5 world units.
2. **Raised-cosine blend easing** instead of linear (ease-in/ease-out) — a linear ramp starts and stops the pose correction abruptly, which reads as two small hitches bracketing the fade. Applies to interrupt blends too.

Deliberately **not** added: a synthetic minimum blend for zero-blend clips. Stride clips (`ogsrunl`/`ogsrunr`) author 0 ms by design — left/right strides line up pose-for-pose — and the reference behavior also hard-cuts same-schema strides. The measurement confirms walking is unaffected; its remaining seam delta is a phase-reset timing hiccup (next PR in the stack).

## Measured impact (medsci1 pipe hybrid, `npm run anim-smoothness`, deterministic fixed 60 Hz)

| idle | before | after |
|---|---|---|
| clip-cycle pops | 4 per 10 s, max **1.50** units | **0** — max frame delta 0.163 |
| mean max-joint jerk | 0.130 | 0.108 |

Walking metrics unchanged (max 0.7515 identical), as intended.

## Validation

- Two new unit tests: authored blend engages from a drained queue (fails without the fix); zero-blend clips still hard-cut. 25/25 `dark` lib tests pass.
- Full e2e + cross-engine review + visuals: running, results to follow

## Visuals

BEFORE (`origin/main`) vs AFTER (this PR), medsci1 idle pipe hybrid, captured headlessly via the debug runtime (deterministic fixed 60 Hz; identical clip-change frames on both builds). Watch the right arm at the `ogsidle4`→`oggidle1` seam (~8.2 s in): BEFORE snaps the pose in a single frame (measured 1.50-unit max-joint jump); AFTER cross-fades over 500 ms (max 0.16).

**Slow-motion (~7×) of the worst seam** — the single-frame snap is unmistakable on the left, smooth on the right:

seam-blend slow-motion before/after

**Real-time** (covers two idle transitions, ~5.5 s) — the BEFORE snap reads as a quick jerk, AFTER is continuous:

seam-blend real-time before/after

**Still, same sim frame mid-transition** — BEFORE has already hard-snapped the arm/knife up; AFTER is mid cross-fade:

seam-blend still before/after

<sub>Verified over HTTP: on AFTER, `GET /v1/entities/:id/animation` reports `blend != null` (duration 0.5 s) at every idle clip change; on `origin/main` the same field is `null` — the authored blend never engaged.</sub>

### DIFF
--- a/dark/src/motion/animation_player.rs
+++ b/dark/src/motion/animation_player.rs
@@ -104,19 +104,40 @@ impl AnimationPlayer {
             .animation
             .push_front((animation.clone(), AnimationFlags::PlayOnce));
 
-        let blend_state = if let Some((current_clip, flags)) = player.animation.first() {
-            let duration = animation.blend_length.as_secs_f32();
-            if duration > 0.0 {
-                Some(BlendState {
-                    from_clip: current_clip.clone(),
-                    from_frame: player.current_frame as f32,
-                    from_looping: matches!(flags, AnimationFlags::Loop),
-                    duration,
-                    elapsed: 0.0,
+        // Fade from the playing clip, or - when the queue already drained (the
+        // normal case for AI clips, whose completion handler queues the next
+        // clip one tick after the previous one finished) - from the frozen
+        // last-frame pose that get_transforms has been showing. Without the
+        // fallback a clip's authored blend_length only ever applied to
+        // interruptions, so e.g. idle cycling (500ms authored) hard-popped.
+        let duration = animation.blend_length.as_secs_f32();
+        let blend_from = player
+            .animation
+            .first()
+            .map(|(clip, flags)| {
+                (
+                    clip.clone(),
+                    player.current_frame as f32,
+                    matches!(flags, AnimationFlags::Loop),
+                )
+            })
+            .or_else(|| {
+                player.last_animation.as_ref().map(|clip| {
+                    (
+                        clip.clone(),
+                        clip.num_frames.saturating_sub(1) as f32,
+                        false,
+                    )
                 })
-            } else {
-                None
-            }
+            });
+        let blend_state = if duration > 0.0 {
+            blend_from.map(|(from_clip, from_frame, from_looping)| BlendState {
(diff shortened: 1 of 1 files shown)


tommy-xr:
Verification gates complete:
- **Full e2e: 96/97.** The one failure (`a killed monster dies and stays dead`, corpse drifted 0.75u) passes in isolation on the same build — suite-context flake, not a blend regression. Possibly a pre-existing borderline made more visible by #464 (monsters now carry genuine walking momentum into death instead of seam-stalling); worth watching, tracked separately if it recurs.
- **Cross-engine review (Claude + Codex):** no Critical/High. [AGREED] finding (duplicated blend-source computation) fixed via a shared `blend_from` helper; Low threshold mismatch (`> 0.0` vs `> f32::EPSILON`) aligned. Codex flagged a **pre-existing** edge for the record: a blend outliving a short PlayOnce clip keeps fading while the next queued clip becomes the target — only reachable with stacked queues, which no code path creates after #464; deferred deliberately.
- **Re-measured after review fixes:** idle and walking both 0 seam spikes (idle max frame delta 0.163 vs 1.50 before; walking unchanged).
- Visuals (idle clip-cycle pop vs fade, before/after): capture in progress, will be appended to the PR body.

PR #472, opened 2026-07-11
feat(render): stretchy-vertex skinning for creature meshes

tommy-xr:
## Summary

Creature (AI) meshes author blend regions between body parts as **stretchy segments** carrying one weight per vertex. Our loader parsed the weights and the stretchy flag but discarded both, binding every vertex rigidly to a single joint — so creature skin cracked and pinched at elbows, knees, shoulders, and neck while animating ("parts of the mesh look incorrectly mapped"). This was a pre-existing, documented gap (`research/model-differences.md` §1).

**Ground truth**: validated the format byte-for-byte against `GRUNT_P.BIN` — 29 segments in stretchy/rigid pairs per joint, and the 110-entry weight array covers *exactly* the vertices of stretchy segments. It also confirmed our rigid binding was already correct (the loader reads the segment table's joint ids directly).

**The fix**:
- Loader keeps the stretchy flag + weights; stretchy vertices emit a two-bone binding: the joint `j` at weight `w`, and palette slot `40 + j` at weight `1 − w`.
- Slot `40 + j` is the joint's **parent frame**: the parent's orientation anchored at the joint's own position (`Skeleton::expand_skinning_palette`). Sharing the pivot means the blend *rotates* about the joint rather than tearing between two pivots; the `w=0`/`w=1` endpoints match the parent/child frames exactly. (The reference behavior slerps the rotation per vertex; a weighted matrix blend with a shared pivot matches it at the endpoints and closely approximates it between — revisit only if extreme flex shows artifacts.)
- `SceneObject`/skinned material grow to an 80-slot palette; the 40-slot setter pads new slots with each joint's own transform, so ragdolls, GLB models, and static-object callers behave identically by construction.
- Out-of-range segment joints (latent i8→u32 cast hazard) now bind to the root with a warning instead of indexing garbage past the palette.

## Validation

- 27/27 `dark` unit tests, including new ones: palette parent-frame construction, rigid/stretchy binding emission (with weight clamping), and structure-level invariants against the real grunt mesh (one weight per stretchy vertex; all joints in palette range)
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p debug_runtime` clean; `cargo fmt`
- **Missions e2e: 23/23** — every mission loads (the loader touches every creature everywhere)
- In-game render check: hybrid draws fully intact with the 80-slot palette (wiring errors would explode/detach parts)
- Cross-engine review + bent-joint before/after visuals: running, results to follow

Note: the vertex shader's bone palette is now 80 mat4 (320 vec4 uniform components). Quest's Adreno GLES3 exposes far more than this; flagging in case a minimum-spec GLES device ever matters.

## Visuals

MedSci-1 pipe hybrid, idle animation (arm raises the pipe across its body), framed on the upper body from behind. Same deterministic request sequence on both builds (fixed 60 Hz stepping), so every frame is pose-matched — the only difference is vertex skinning.

stretchy skinning before/after — idle

<sub>Left = base (`main`), right = this branch. Watch the right shoulder/deltoid and the elbow as the arm bends: BEFORE the skin facets and pinches into a hard seam at the joint; AFTER the parent/child bones blend so the surface stays continuous. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`).</sub>

### Before -> after (peak elbow flex)

shoulder + elbow before/after

<sub>Money shot at maximum joint flex: BEFORE (left) shows the faceted deltoid and the hard elbow seam where each vertex rigidly follows one bone; AFTER (right) the shoulder rounds smoothly into the bicep and the elbow crease is continuous. The improvement is a surface detail on a low-poly model — subtlest on straight limbs, most visible at bent shoulders/elbows.</sub>

### DIFF
--- a/dark/src/model.rs
+++ b/dark/src/model.rs
@@ -60,13 +60,14 @@ impl AnimatedModel {
     }
 
     fn to_animated_scene_objects(&self, player: &AnimationPlayer) -> Vec<SceneObject> {
-        let skinning_data = player.get_transforms(&self.skeleton);
+        let pose = player.get_transforms(&self.skeleton);
+        let palette = Skeleton::expand_skinning_palette(&pose, &self.skeleton);
 
         self.scene_objects
             .iter()
             .map(|m| {
                 let mut new_obj = m.clone();
-                new_obj.set_skinning_data(skinning_data);
+                new_obj.set_skinning_palette(palette);
                 new_obj
             })
             .collect::<Vec<SceneObject>>()
@@ -82,14 +83,17 @@ impl AnimatedModel {
             }),
             &rpds::HashTrieMap::new(),
         );
-        let new_data = animated_skeleton.get_transforms();
+        let new_data = Skeleton::expand_skinning_palette(
+            &animated_skeleton.get_transforms(),
+            &animated_skeleton,
+        );
 
         let new_scene_objects = self
             .scene_objects
             .iter()
             .map(|m| {
                 let mut new_obj = m.clone();
-                new_obj.set_skinning_data(new_data);
+                new_obj.set_skinning_palette(new_data);
                 new_obj
             })
             .collect::<Vec<SceneObject>>();
--- a/dark/src/ss2_bin_ai_loader.rs
+++ b/dark/src/ss2_bin_ai_loader.rs
@@ -14,7 +14,7 @@ use engine::{
     scene::{SceneObject, VertexPositionTextureSkinnedNormal},
     texture::{AnimatedTexture, TextureTrait},
 };
-use tracing::trace;
+use tracing::{trace, warn};
 
 use crate::{
     SCALE_FACTOR,
@@ -29,6 +29,11 @@ use crate::{
     util::load_multiple_textures_for_model,
 };
 
+/// Number of base joint slots in the skinning palette; stretchy frames for
+/// joint `j` live at palette slot `MAX_JOINTS + j` (see
+/// `ss2_skeleton::expand_skinning_palette`).
+pub const MAX_JOINTS: usize = 40;
+
 #[derive(Clone)]
 pub struct SystemShock2AIMesh {
     // pub header: BinHeader,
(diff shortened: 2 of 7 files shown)


tommy-xr:
Cross-engine review synthesis (Claude + Codex, both adversarial):

**Verdict: core math and data flow confirmed clean by both engines** — parent-frame construction (shared-pivot blend), shader endpoint reductions, weight-index ordering, caller padding (ragdoll/GLB/static unaffected), normal renormalization, and the negative-index fix all verified.

Findings to address (fix pass incoming):
- **[Codex High] GLES uniform budget**: 80 mat4 = 320 vec4 vertex uniforms exceeds the GLES 3.0 guaranteed minimum (256), and Quest Adreno drivers commonly report exactly 256 — the shader could fail to link on the VR target (which we can't compile-check locally). Fix: pack bones as mat3x4 rows (240 vectors for all 80 slots, within the floor) with a single contiguous upload — which also eliminates the 80-separate-uniform-call overhead (Codex Low).
- **[AGREED] silent rigid fallback**: a stretchy vertex whose weight index falls outside the array silently renders rigid, masking malformed data (the real-mesh test only validates totals). Fix: per-segment range validation with a warning.
- [Claude Low] Hoist the palette-size constant to a single source of truth; simplify the palette copy; align debug_normal_material's palette declaration.

Holding the fix commit until the in-flight visuals capture finishes (it builds from this branch), then re-validating: unit tests, missions e2e, and a deterministic screenshot compare (the packing change must be pixel-identical).

tommy-xr:
Final gates after the review fix pass:
- **Missions e2e: 23/23** (re-run after the mat3x4 packing + validation changes)
- 27/27 `dark` unit tests; warning-free `-D warnings` compile
- **Packing proven render-identical**: same deterministic capture sequence before/after the mat3x4 change differs by 15 pixels at ≤8/255 max channel delta (float reassociation from the reordered blend arithmetic) — no visual change
- Vertex uniform budget now 240 vec4 for the full 80-slot palette + 12 for world/view/projection = 252, within the GLES 3.0 guaranteed minimum of 256

Ready for review.

PR #475, opened 2026-07-12
feat(anim): fractional-frame interpolation + phase carry across seams

tommy-xr:
## Summary

Fixes #3+#4 of the animation-smoothness stack (stacked on #469), both in the playback clock:

1. **Fractional-frame interpolation.** Poses interpolate between adjacent keyframes at the sub-frame position (per-joint rotation slerp, translation lerp, root transform included) instead of snapping to whole 30 fps keyframes. At 60 Hz, every other tick used to repeat the previous pose — a constant sawtooth under all motion. The blend-from pose also advances fractionally during crossfades. (Correcting the record: `research/model-differences.md` claims both engines snap to integer frames; observation of shipped behavior shows sub-frame interpolation — the doc will be updated separately.)
2. **Phase carry across seams.** A completing clip carries its overshoot past the final frame into the next, and re-queues preserve the sub-frame remainder, so the frame clock never phase-resets at a seam. Honest note: **no measurable effect at exactly 60 Hz with these clips** (their durations are whole multiples of the tick) — this matters at 72/90/120 Hz VR refresh rates and prevents cadence drift. Unit-tested either way.

## Measured impact (medsci1 pipe hybrid, deterministic 60 Hz)

| | before (post-#469) | after |
|---|---|---|
| idle mean max-joint jerk | 0.108 | **0.010** |
| idle max frame delta | 0.163 | **0.081** |
| walking mean max-joint jerk | 0.556 | **0.222** |
| walking mid-clip p50 | 0.133 (alternating 0 / 0.27) | 0.154 (uniform every tick) |
| walking stride-seam max | 0.752 | 0.801 |

Versus the original baseline: idle jerk is now **13×** lower, walking jerk **3.9×** lower. The stride seam is now the one dominant residual (0.80 vs 0.15 mid-clip) — the one-tick completion hold plus the hard stride cut. That's a game-loop-ordering follow-up (apply queued animation effects before posing), deliberately out of scope here. The auto-threshold now flags all 40 stride seams because mid-clip variance collapsed — a metric artifact of the improvement, not a regression (absolute seam magnitude ~unchanged).

## Validation

- 27/27 `dark` unit tests, including new: overshoot carry + cadence preservation across a queue seam; slerp-midpoint interpolation (45° at fraction 0.5)
- Full e2e + cross-engine review: running, results to follow

### DIFF
--- a/dark/src/model.rs
+++ b/dark/src/model.rs
@@ -80,6 +80,7 @@ impl AnimatedModel {
                 animation_clip: &animation_clip,
                 frame,
                 fraction: 0.0,
+                wrap: false,
                 cancel_root_motion: false,
             }),
             &rpds::HashTrieMap::new(),
--- a/dark/src/model.rs
+++ b/dark/src/model.rs
@@ -79,6 +79,7 @@ impl AnimatedModel {
             Some(AnimationInfo {
                 animation_clip: &animation_clip,
                 frame,
+                fraction: 0.0,
                 cancel_root_motion: false,
             }),
             &rpds::HashTrieMap::new(),
--- a/dark/src/motion/animation_player.rs
+++ b/dark/src/motion/animation_player.rs
@@ -147,9 +147,18 @@ impl AnimationPlayer {
         self.animation
             .first()
             .map(|(clip, flags)| {
+                // Fade from the pose actually on screen - the whole frame
+                // plus the sub-frame remainder - not the integer keyframe
+                // behind it (which would start the fade with a backward pop).
+                let time_per_frame = clip.time_per_frame.as_secs_f32();
+                let fraction = if time_per_frame > f32::EPSILON {
+                    (self.remaining_time / time_per_frame).clamp(0.0, 1.0)
+                } else {
+                    0.0
+                };
                 (
                     clip.clone(),
-                    self.current_frame as f32,
+                    self.current_frame as f32 + fraction,
                     matches!(flags, AnimationFlags::Loop),
                 )
             })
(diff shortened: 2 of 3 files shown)


tommy-xr:
Full e2e: **97/97 real tests pass**. The run initially reported 2 failures (cyber-modules, elevator-panel) — both are stale compiled artifacts: `dist/` survives branch switches and tsc doesn't remove outputs whose sources are gone, so compiled tests for #466/#467 features (not present on this stack's base) ran against a runtime that predates them (the failures are literally `undefined !== 0` on a field the branch doesn't have, and an assertion whose own message says 'on main labels are null'). Cleaned `dist/`; the suite globs only the branch's own tests now.

Cross-engine review synthesis + fix pass incoming: blend-from will start at the fractional (displayed) pose rather than the integer keyframe behind it [Codex High, real for mid-clip interrupts only]; carried overshoot capped to sub-frame phase via modulo [AGREED]; wrap-aware interpolation for looping clips [AGREED]; empty-keyframe guard + minor simplification [Claude Lows].

tommy-xr:
Review fix pass landed (924b525) and re-validated:
- blend-from now fades from the fractional displayed pose (Codex High — real for mid-clip interrupts; the drained-queue seam path was already exact per the Claude reviewer's trace)
- completion carries only the sub-frame remainder — whole hitch-frames drop rather than leaving logical time ahead of the rendered pose ([AGREED]; simplification: the advance loop already bounds the remainder, so the carry is just remaining_duration)
- looping clips interpolate across the wrap ([AGREED]; blend-from inherits via from_looping)
- empty-keyframe guard + redundant-modulo cleanup (Claude Lows)

29/29 dark unit tests (two new: fractional blend-from, wrap-vs-hold). Re-measured: identical to the PR-body numbers (idle jerk 0.0101, walking 0.2219, seam max 0.8006) — the fixes affect interrupt/loop/hitch paths the steady capture doesn't exercise. Full e2e was 97/97 on the real suite (see prior comment re: two stale-dist artifacts). Ready for review.

PR #476, opened 2026-07-12
fix(anim): ramp end_rotation across playback; break failed-query loop

tommy-xr:
## Summary

The final two items from the animation audit (stacked on #475):

1. **`end_rotation` ramp.** A clip's authored end direction was applied as a single yaw snap on the completion tick — a visible pop on every turn/gesture clip. It now arrives as per-tick increments proportional to playback traversed; the fractions telescope, so a full playthrough applies exactly the authored total (unit-tested: 90° over a multi-tick playthrough sums to 90.000). The `-0.5` scale on application predates recorded history (initial-commit era); it's preserved and annotated so clips leave entities facing exactly where they always have — re-deriving the correct factor needs reference footage and is out of scope.
2. **Failed-query loop guard.** A motion query resolving to nothing reports `AnimationCompleted` (so scripted `Play` actions never hang), but the AI's completion handler re-queries — an unresolvable query churned every frame, forever. The same failing query now reports completion once (keyed on query *items*: the selection strategy carries a per-request counter that defeated the first attempt at deduping — caught by live measurement); any different query or a success resets the guard, so behavior changes recover normally.

## Verification

- **Live guard measurement** (medsci1, 600 deterministic frames, `RUST_LOG=shock2vr=debug`): failed-query breadcrumbs **602 → 2** (one per distinct failure); the ~64 remaining breadcrumbs are legitimate successful queries.
- 30/30 `dark` unit tests (new: rotation increments across ticks totaling exactly the authored angle)
- Smoothness capture + full e2e + cross-engine review: running, results to follow (hybrid walk metrics should be byte-identical — its run clips author `end_dir 0`)

### DIFF
--- a/dark/src/motion/animation_player.rs
+++ b/dark/src/motion/animation_player.rs
@@ -64,6 +64,11 @@ pub struct AnimationPlayer {
     current_frame: u32,
     remaining_time: f32,
     blend_state: Option<BlendState>,
+    /// Playback position (frame units) through which the head clip's
+    /// end_rotation has been emitted - the anchor for the per-tick rotation
+    /// ramp. Reset to 0 whenever a new clip becomes the head, so the ramp
+    /// covers a carried seam remainder from position zero.
+    rotation_pos: f32,
     /// Pose with the clip's root motion cancelled (see
     /// `AnimationInfo::cancel_root_motion`). Set for first-person viewmodels,
     /// whose entity transform is re-anchored to the camera every frame.
@@ -80,6 +85,7 @@ impl AnimationPlayer {
             current_frame: 0,
             remaining_time: 0.0,
             blend_state: None,
+            rotation_pos: 0.0,
             cancel_root_motion: false,
         }
     }
@@ -93,6 +99,7 @@ impl AnimationPlayer {
             current_frame: 0,
             remaining_time: 0.0,
             blend_state: None,
+            rotation_pos: 0.0,
             cancel_root_motion: false,
         }
     }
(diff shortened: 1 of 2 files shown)


tommy-xr:
Cross-engine review synthesis + fix pass (93663e1 → f0aa52b):

- **[AGREED] loop-wrap rotation loss + carried-seam under-rotation**: fixed structurally — the ramp now anchors on an explicit `rotation_pos` (position through which rotation has been emitted) instead of deriving it from the frame counter. Loop wraps emit their wrapped segment (multi-cycle hitches included); a clip inheriting seam carry ramps from zero. Tests: 4 loop cycles = exactly 360°; chained carried-seam clips each total exactly 90°.
- **[AGREED] same-failing-Play-twice stall**: the guard now *defers* a suppressed completion to window expiry instead of dropping it — a sequence replaying the same failing action advances within 0.5 s (was: stall to the 30 s watchdog). Live: ~60 → ~2 failed queries/s per failing entity, and no requester can hang.
- **[AGREED] entity cleanup**: `remove_entity_single` clears the guard and the (pre-existing leak) `id_to_animation_player` entry — shipyard recycles ids.
- Claude reviewer verified the mid-clip/completion telescoping, event ordering/composition (N increments ≡ old single snap), and key stability.

36/36 dark unit tests; `-D warnings` clean; smoothness metrics byte-identical (idle jerk 0.0101 / walking 0.2219 / seam 0.8006 — hybrid run clips author end_dir 0); full e2e was 97/97. Both this branch and #475 are restacked onto current main.

PR #478, opened 2026-07-16
feat(ragdoll): graduate the multibody rig — sleep on settle, default rig, full joint diagnostics

tommy-xr:
## Summary

The multibody ragdoll rig's June blocker ("extremities won't settle" contact limit-cycle) no longer reproduces on main — most likely fixed as a side effect of #333's collider friction/restitution work (see the re-evaluation in #477). What remained was a ~0.5 rad/s single-extremity buzz that sat above rapier's default angular sleep threshold (0.5 rad/s), and one awake body keeps the whole jointed island awake — so the corpse never slept and the buzz never stopped.

This PR graduates the multibody rig:

- **Raised angular sleep threshold (1.5 rad/s) on ragdoll bodies** — the settled island now sleeps ~6 s after death (velocities zero exactly; the buzz is gone). Contact or an applied impulse wakes the corpse, so it stays fully pokeable (verified). Rapier requires the *entire island* below thresholds for 2 continuous seconds, so a corpse in real motion can't freeze mid-fall.
- **Multibody is now the default rig** (no hip gap possible — translation is structurally not a DOF — and a visibly better settled pose than the impulse rig, which holds a 9.6 cm hip gap at rest). New `--experimental ragdoll_impulse` falls back to the legacy impulse rig.
- **Joint diagnostics cover the multibody set**: `/v1/physics/joints` and the `--debug-physics` anchor overlay now enumerate multibody joints too, with a new `joint_type` field (`"impulse"` / `"multibody"`).
- **New debug endpoint `POST /v1/physics/bodies/:id/impulse`** — apply a world-space impulse to a dynamic body (zero impulse = pure wake), closing part of the ragdoll harness gap noted in `projects/ragdoll.md`. Used here to verify wake-on-poke headlessly.

## Verification (headless, debug_ragdoll, fixed 60 Hz)

| Check | Result |
| --- | --- |
| Default rig routing | 19 multibody joints, 0 impulse; max separation **0.0** |
| `ragdoll_impulse` fallback | 19 impulse joints |
| Settle + sleep | velocities → exactly 0 at t≈6 s; `is_sleeping: true` |
| Wake on poke | impulse → motion resumes (`is_sleeping: false`), re-sleeps ~12 s later |
| Zero-impulse wake | `{"impulse":[0,0,0]}` wakes the body (rapier skips zero impulses, wake is explicit) |
| Settled pose | flat prone corpse (screenshot below), vs the impulse rig's folded-legs-in-air pose |

`RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean. Cross-engine review (/xreview: Claude + Codex) found no Critical/High issues; both engines' findings addressed in the second commit.

Stacked note: reads best after #477 (the re-evaluation doc); merge order doesn't matter technically.

## Visuals

multibody ragdoll: kill → collapse → settle → sleep → impulse poke

<sub>`debug_ragdoll` with the multibody rig as default: the pipe hybrid is killed, collapses into a ragdoll, settles into a flat prone corpse and falls asleep (`/v1/ragdoll/metrics` reports `max_linear_speed == 0`), then a `POST /v1/physics/bodies/:id/impulse` poke on a thigh body wakes it — it shifts and re-settles to sleep. Settle plays accelerated; collapse and poke are ~real-time. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

| Mid-collapse | Settled sleeping corpse |
| --- | --- |
| mid-collapse | settled sleeping corpse |

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -444,10 +444,12 @@ The project supports experimental flags for gating in-progress features during d
 - **`ragdoll`**: spawn a physics ragdoll on creature death (`SlayEntity`) instead of
   just removing the entity. Without it, death is unchanged.
 
-- **`ragdoll_multibody`**: make ragdolls use reduced-coordinate (multibody) joints
-  anchored at the limb articulation points — limbs can't separate (no hip gap), but
-  the rig is experimental (extremities can jitter on floor contact). Without it,
-  ragdolls use the stable impulse-joint rig. See `projects/ragdoll-settling-followup.md`.
+- **`ragdoll_impulse`**: make ragdolls fall back to the legacy impulse-joint rig
+  (heavier core, soft locked translation — settles but the hip visibly sags/separates
+  under load). Without it, ragdolls use the default reduced-coordinate (multibody)
+  rig — joints anchored at the limb articulation points, so limbs structurally can't
+  separate — with raised sleep thresholds so the settled corpse goes fully still
+  (and wakes again on contact/impulse). See `projects/ragdoll-settling-followup.md`.
 
 - **`loading_screen`**: show the animated loading screen during level transitions
   (`GlobalEffect::TransitionLevel` / `TestReload`) instead of switching instantly. The
--- a/projects/ragdoll-settling-followup.md
+++ b/projects/ragdoll-settling-followup.md
@@ -116,6 +116,9 @@ the impulse rig kept as the stable default:
   otherwise it uses the impulse rig (parent-origin anchor, heavy core, soft
   translation). Routing verified: default → 19 impulse joints; multibody flag → 0
   impulse joints, bodies stay connected (~1.6 m extent).
+  **2026-07-16 update: multibody is now the DEFAULT rig** (it settles and sleeps —
+  see the graduation notes below); the flag is retired and replaced by
+  `--experimental ragdoll_impulse`, which falls back to the legacy impulse rig.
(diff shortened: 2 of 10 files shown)


tommy-xr:
Two additional fixes pushed after end-to-end verification surfaced them (both found by the visuals capture noticing that different impulse magnitudes produced byte-identical simulations):

1. **Impulses on multibody links were silently ignored** — the reduced-coordinate solver recomputes link-body velocities from joint velocities every step, clobbering `apply_impulse`. `apply_body_impulse` now converts the impulse into a one-step force (the articulated forward dynamics *does* read the per-body force accumulator), cleared by the existing `clear_forces` cadence. Verified: 4× impulse → ~2× response (sublinear from articulated coupling). This mechanism is also what the upcoming killing-blow-impulse work will use.

2. **Rare spawn-transient runaway → NaN AABB → parry BVH panic** (~1-in-6 `debug_ragdoll` runs, nondeterministic across processes). `RagDoll::update` now clamps generalized-DOF velocities to a 30-unit cap once per frame, breaking the feedback loop. 10/10 soak runs now settle cleanly to an identical rest state (clamp engaged in 8 — the transient is common; it just usually self-recovered before NaN).

Issue #487, opened 2026-07-17
Crowd handling: converging AIs can gridlock body-to-body (no separation/yielding)


tommy-xr:
Observed while fixing #481: when multiple pursuing AIs converge into a tight space (e.g. two hybrids in the sealed crew-section pocket on medsci1, or a queue at a choke point), they push head-on into each other's capsules and can gridlock. The stall-recovery backout (added for #481) staggers and resolves most jams, but pairs in dead-end pockets can still lock up.

Classic fix: separation steering (small lateral bias away from nearby creatures, weighted into the path-follow heading) and/or yield priority (the AI closer to its goal keeps moving, the other side-steps). The alertness-churn e2e deliberately excludes mutual jams (creature within 2 units) from its freeze assertion and cites this issue.

Repro: medsci1, `--experimental nav_bridges`, teleport player to the open deck, DebugForceChase, wait ~40s; check the crescent-pocket pair with `GET /v1/ai/paths` live fields.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #485, opened 2026-07-16
nav_bridges follow-ups: obstruction-validated bridges, explicit crossing geometry, search accounting


tommy-xr:
Follow-ups from the cross-engine review of #483's experimental island bridging (`nav_bridges`):

1. **Obstruction validation**: bridge candidates are accepted from proximity + slope only; a candidate through thick geometry is possible (rare at the current gates, and stall-escape copes, but paths can hug walls there). Validate each synthesized crossing with a creature-width physics sweep at load (Rapier is available once the mission is built), dropping candidates that hit static geometry.
2. **Explicit crossing geometry**: bridge links carry `u32::MAX` edge vertices, so waypoint building falls back to the destination cell center; on large cells the executed segment can diverge from the measured gap. Store the entry/exit points (the closest-vertex pair the gap was measured from) and route through them.
3. **Search accounting**: one frame-budget slot can currently run up to three searches (A*, stressed retry, full-component Dijkstra via `find_path_toward`). Either budget individual searches or add a stats counter for partial-route Dijkstras so per-frame load is measurable.
4. **Dedup reachable-set logic**: `find_path_toward` and `find_closest_reachable_cell` both run `dijkstra_all` + nearest-cell selection; extract one helper returning the parent map + best cell.
5. **Locked-door gate e2e**: the live sync (cell-door object id → GlobalTemplateIdMap → door state) is covered only by a synthetic unit test; add an SDK e2e that locks a real door and asserts pursuing AIs stop routing through it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #488, opened 2026-07-17
fix(ai): eliminate mid-route freezes (whiskers, corner insets, stall recovery) - Fixes #481

### LINKED ISSUE #481: AI freezes (zero movement, Chase behavior, valid route) after alertness churn
## Symptom

After an alert/decay cycle followed by a re-alert (e.g. `DebugAlertAll` → 10s decay → `DebugForceChase`), several monsters end up **fully frozen**: `AIBehavior: Chase`, alertness High, a valid computed route in `GET /v1/ai/paths` (e.g. 31 waypoints), but **zero movement over 30+ seconds** — the position doesn't change at all, suggesting no locomotion clip is playing (zero root-motion velocity), not slow/stuck steering.

## Repro (debug runtime, medsci1)

```bash
cargo dbgr --mission medsci1.mis --experimental nav_bridges --port 8081
# teleport player to open deck
curl -X POST :8081/v1/player/teleport -d '{"x":-14,"y":0.5,"z":-30}'
curl -X POST :8081/v1/input/action -d '{"action":"DebugAlertAll"}'
curl -X POST :8081/v1/step -d '{"frames":630}'    # let alertness decay
curl -X POST :8081/v1/input/action -d '{"action":"DebugForceChase"}'
curl -X POST :8081/v1/step -d '{"frames":1830}'
# observe several OG- hybrids with unchanged positions but Chase behavior
# and non-empty routes in GET /v1/ai/paths
```

Without the decay phase (ForceChase on a fresh world), the same monsters chase and converge normally — so the churn (behavior swaps from escalate → decay → search/wander → forced re-escalate) appears to race a behavior/animation transition and leave the AI with no playing locomotion clip. Likely related to the clip-replacement seams handled in #464 / #469 / #476.

Bisected: reproduces with and without the heading-error locomotion-scale coupling, so it predates that change.

## Impact

Any gameplay sequence that rapidly cycles alertness (alarms, repeated stimuli) can beach monsters permanently. Also constrains e2e test design (force-chase.e2e.test.ts runs its convergence phase on a fresh world to avoid this).

tommy-xr:
## Problem

Monsters with valid routes froze permanently mid-route (running in place) under sustained pursuit — capping the convergence numbers in #483's demo. Issue #481.

## Diagnosis

Added live steering introspection (`GET /v1/ai/paths` now reports the waypoint actually being followed, path length, target, stall seconds — plus an SDK wrapper), then interrogated frozen specimens frame-by-frame. Four stacked mechanisms, each individually verified:

1. **Stairs read as walls**: knee-height avoidance whiskers hit descending steps' risers (vertical faces pass the normal filter). → cast from body center.
2. **Avoidance vs path heading war**: the main whisker's fixed +180° mitigation reversed the AI into a perpetual flip-flop against path steering; avoidance also *preempted* the path (first in chain). → deflect along the hit normal, and reorder every chain so path steering leads (the route already avoids static geometry); avoidance guards only the no-route fallback.
3. **Corner wedging**: taut-path corner insets (1.5 dark ft) were under the widest creature capsule radius — routes grazed door frames and the body wedged pushing full-speed into the corner. → EDGE_CLEARANCE 3 ft; waypoint arrival height tolerance 4→7 ft (waypoint heights are cell-floor heights, body centers ride ~4.2 ft above).
4. **Stall loops**: the stall escape re-pathed into the identical wedge. → on stall, back out toward the previous waypoint (jittered ~1s) before re-pathing; also staggers/resolves most mutual crowd jams.

Bonus: island bridges are now **vetted against static geometry** (torso-height ray between cell centers), closing the bridges-through-railings item from #485.

## Verification

- New e2e `alertness-churn.e2e.test.ts` encodes the freeze signature (far from its *own* route end, alone, not moving): **fails on the pre-fix build (3 frozen), passes with the fixes**. Residual mutual crowd jams in sealed dead-end pockets are excluded and tracked as #487.
- Full SDK e2e suite: 102/103 (the one failure, corpse-drift tolerance in monster-death, reproduces on clean `main` — pre-existing flake from the ragdoll momentum work).
- Unit: 157 green, incl. new regression tests for the arrival-height gate and edge-clearance collapse.
- Convergence numbers comparison vs #483 follows in a comment.

### DIFF
--- a/shock2vr/src/game_scene.rs
+++ b/shock2vr/src/game_scene.rs
@@ -724,6 +724,15 @@ pub struct DebugAiPathEntry {
     pub outcome: String,
     /// Waypoints of the computed route (empty for Failed)
     pub waypoints: Vec<[f32; 3]>,
+    /// Index of the waypoint the steering is CURRENTLY following (its live
+    /// path may lag the latest computed route above)
+    pub live_next_waypoint: Option<usize>,
+    /// Length of the live path being followed
+    pub live_path_len: Option<usize>,
+    /// World position currently steered toward
+    pub live_target: Option<[f32; 3]>,
+    /// Seconds without progress toward the current waypoint
+    pub live_stall_seconds: Option<f32>,
 }
 
 /// Snapshot of the pathfinding service's monotonic query counters.
--- a/shock2vr/src/mission/mission_core.rs
+++ b/shock2vr/src/mission/mission_core.rs
@@ -717,10 +717,32 @@ impl MissionCore {
         }
 
         let nav_bridges = game_options.experimental_features.contains("nav_bridges");
+        // Vet each synthesized island crossing against the static level
+        // geometry: a torso-height ray between the two cell centers must be
+        // clear, so bridges can't route AIs into railings or thin walls
+        // (which left them stalling at the seam forever - issue #481).
+        let bridge_validator = |from: cgmath::Vector3<f32>, to: cgmath::Vector3<f32>| -> bool {
+            let delta = to - from;
+            let distance = delta.magnitude();
+            if distance <= f32::EPSILON {
+                return true;
+            }
+            physics
+                .ray_cast2(
+                    Point3::new(from.x, from.y, from.z),
+                    delta / distance,
+                    distance,
+                    crate::physics::InternalCollisionGroups::WORLD,
+                    None,
+                    true,
+                )
+                .is_none()
+        };
         let pathfinding_service = abstract_mission.path_database.as_ref().map(|db| {
             Arc::new(PathfindingService::with_nav_options(
                 Arc::new(db.clone()),
                 nav_bridges,
+                Some(&bridge_validator),
             ))
         });
         // Steering strategies path through this unique; MissionCore keeps its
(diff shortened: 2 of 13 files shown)


tommy-xr:
## Convergence numbers: main (pre-#483) → #483 → this PR

Same protocol throughout: medsci1, player teleported to the open deck at (-14, 0.5, -30), 6 hybrids tracked once per second.

| Metric | main (pre-#483) | #483 (merged) | this PR |
|---|---|---|---|
| **Frozen alone mid-route after alertness churn** (the #481 signature, refined predicate) | n/a (no routes to freeze on) | **3** (3–5 across runs) | **0** |
| Fleet approach, 90s pinned chase (sum of distance deltas) | −16.6 (held-alert; no pin exists) | −34.4 best case / −14..−20 when the freeze hit the mid cohort | **−35.1, freeze-immune** |
| Churn-preceded 30s window (worst case for #483) | — | 1 of 6 closing | all route-holders moving or arrived |
| Long-route arrival (start ~30 units) | dead-ends at 16.6 (wall) | arrives when not frozen (closest 2.8) | arrives reliably (closest 3.0) |

The headline: #483's *best-case* fleet numbers were already good, but a per-run lottery decided whether mid-distance hybrids froze mid-route (whisker/wedge/stall-loop mechanisms, now individually fixed). This PR makes the best case the only case: the churn e2e fails on main (3 frozen) and passes here (0), across every run tried.

Remaining static hybrids are all accounted for: furniture-sealed storage room (route `Failed` by design), and the sealed crescent pocket pair (crowd jam in a dead-end closet — #487).

PR #507, opened 2026-07-18
feat(physics): detect and attribute non-finite rigid-body state before the broad-phase

### LINKED ISSUE #506: [functionality] Systemic crash: parry3d bvh_binned_build panic on tripwire sensor exit (earth.mis + station.mis, 5/5 repro) — blocks all traversal
## Domain / kind
`[functionality]` · **bug** (engine crash). **Campaign blocker**: kills the runtime within seconds of normal walking on both earth.mis and station.mis — gates every mission downstream.

## Signature (identical in 5/5 reproductions)
```
event: EndIntersect { sensor_id: EId(...), entity_id: EId(0.0) }   // player exits a "New Tripwire" sensor
shock2vr::scripts::trap_new_tripwire: sensor end intersect ... has_keys_now: false had_keys_before: true trip_flags: ENTER | EXIT | PLAYER
thread 'main' panicked at parry3d-0.25.3/src/partitioning/bvh/bvh_binned_build.rs:58:28:
index out of bounds: the len is 8 but the index is N     // N seen: 8, 24, 9
```
rapier3d 0.31.0 / parry3d 0.25.3. Crash logs preserved locally: `/tmp/s2q-8211-crash1.log` … `crash5.log` (play-through campaign iteration 0).

## Reproductions (5/5, two missions)
- earth.mis: sensor EId 580, twice (west street ~(-11, 21, 46); the z=27.4 corridor)
- earth.mis: sensors 598+579 on the recruitment entrance ramp at (9.6, 22.2, 32.9). Note EId 579 alone was crossed harmlessly earlier in the same run; the fatal frame had TWO simultaneous EndIntersects.
- station.mis: sensor EId 763, walking from the start room toward the Mission Postings door.
- It is NOT the earth→station level transition itself (that fires fine). It is `New Tripwire` (template −2502) sensor exits generally, during normal walking.

Deterministic repro: `cargo dbgr --mission station.mis --port 83xx`, walk from the start room toward the Mission Postings door (thumbstick forward + step). Run with `RUST_BACKTRACE=1` to get the full stack — the campaign logs don't have one.

## Prior art in-repo (important)
`shock2vr/src/physics/mod.rs` ~448-472 already documents this exact panic class: zero/non-finite collider AABBs panic rapier 0.31's BVH broad-phase build ("index out of bounds", nondeterministic because it depends on bin layout around the degenerate AABB), fixed then by `sanitize_collider_size` / `sanitize_collider_radius` (e.g. medsci1 "Lift 1 Walls" zero-radius sphere). The new crashes mean **some path still introduces a degenerate/non-finite AABB that bypasses the sanitizers** — plausibly something mutated on tripwire exit (SwitchLink TurnOff effects: doors starting to close, FX toggles, sound traps), or a kinematic/teleported body with a non-finite position, or a shape type the sanitizers don't cover (trimesh/compound/capsule/scaled).

## Fix expectations
- Find the actual degenerate collider/body (backtrace + log which collider AABB is non-finite/zero at build time — a temporary assertion/scan over collider AABBs before the BVH build is a fine diagnostic).
- Fix at the source (sanitize/skip that collider or fix the bad transform), not by pinning/patching parry — though DO check parry3d changelog ≥0.25.4 for an upstream binned-build OOB fix; an upgrade may be warranted *in addition to* input sanitization.
- Negative-first SDK e2e: station.mis, walk start-room → Mission Postings door, assert the runtime survives N steps (fails on main today, 5/5).
- Regression: full `missions.e2e.test.ts` + the new `gravshaft.e2e.test.ts` (PR #505).

Found by the play-through campaign (session data.json + pt-*.png in /tmp/claude; issue supersedes #504 as iteration-0 blocker).

### LINKED ISSUE #508: [functionality] Creature animation feeds NaN joint transforms to kinematic hitboxes (masked by #507's physics guard)
Found while fixing #506 (PR #507). The physics-side guard now *contains* this, but the producer is still there and should be fixed at the animation source.

## Symptom

On a station.mis session (walking the start-room / Mission Postings area), the new `sanitize_rigid_body_state` guard logged ~2,500 repairs in ~100 s:

```
WARN shock2vr::physics: [physics] entity Some(EId(921.0)): non-finite kinematic target pose
     Isometry { rotation: [NaN, NaN, NaN, NaN], translation: [NaN, NaN, NaN] } - reset to current pose
```

- The offending entities (e.g. EIds 842/844/850/851/852 and 921/924/925/929 in that run) are anonymous `Entity_N` bodies with no scripts/template - i.e. **creature hitboxes** (`shock2vr/src/creature/hit_boxes.rs` builds one kinematic body per joint and poses it from `RuntimePropJointTransforms` every frame).
- Each affected creature contributes ~10 NaN hitbox bodies, positioned where nearby NPCs/droids stand.
- Onset correlates with failed motion queries in the same frames, e.g.:
  `Unable to find animation for queries: [MotionQuery { creature_type: 0, items: [Tag("__null_animation__", false), Tag("prodroid", true)] ... }]`

So the chain is: motion query fails -> animation player / joint blending produces all-NaN joint matrices -> `hit_boxes.rs` derives NaN pos/rotation -> `set_position_rotation2` feeds a NaN kinematic target pose to rapier **every frame**.

Before #507 this poisoned the broad-phase BVH and eventually panicked parry's binned rebuild (the #506 crash). Now the hitboxes just freeze at their last finite pose - playable, but:
- those creatures' hitboxes stop tracking their animation (shots at a moving creature may miss/hit wrong), and
- the guard warn (once per entity) will keep appearing in every session.

## Fix expectation

Find where joint transforms go NaN when a motion query fails (zero-length clip / division by zero duration / uninitialized blend?) and make the animation system emit the last-good or bind pose instead. `hit_boxes.rs` could also cheaply skip non-finite joint matrices, but the animation output is the actual source (renderer skinning presumably consumes the same NaNs).

Repro: launch the debug runtime on station.mis, roam the start area near the NPCs/droids for a couple of minutes, grep the log for `non-finite kinematic target pose` and correlate with the `Unable to find animation` warnings.

tommy-xr:
## Problem (#506)

Walking around earth.mis / station.mis killed the runtime 5/5 with:

```
thread 'main' panicked at parry3d-0.25.3/src/partitioning/bvh/bvh_binned_build.rs:58:28:
index out of bounds: the len is 8 but the index is N   // N seen: 8, 9, 24, usize::MAX
```

Backtrace (reproduced in-game with `RUST_BACKTRACE=1`):

```
rebuild_range_binned  (parry3d 0.25.3 bvh_binned_build.rs:58)
  <- Bvh::optimize_incremental
  <- rapier3d BroadPhaseBvh::update
  <- PhysicsPipeline::step
  <- shock2vr PhysicsWorld::update
```

## Root cause

A single **non-finite rigid-body state at runtime** (NaN velocity, or a NaN kinematic
target pose) puts a NaN collider AABB into the broad-phase BVH. parry's binned rebuild
bins leaves by AABB center; a NaN center silently truncates the min/max fold that
computes the centroid range, so *valid* leaves land outside the range and the computed
bin index goes out of bounds - possibly many frames after the poisoning, whenever the
incremental optimizer happens to rebuild the poisoned subtree (which is why the crash
looked nondeterministic and correlated loosely with tripwire exits: those were just the
loudest log lines in the fatal frame).

- The creation-time sanitizers (`sanitize_collider_size` / `_radius`) only guard collider
  **sizes at creation** - runtime state corruption bypasses them, and the current-AABB
  audit (`/v1/physics/colliders/validate`) reports clean because the *current* pose is
  finite while the poisoned *leaf* persists in the tree.
- **Observed organic producer**: creature **hitboxes** - kinematic bodies driven by
  animation joint transforms that go NaN when motion queries fail - fed all-NaN kinematic
  target poses to physics *every frame* (~2,500 occurrences in 100 s across ~10 hitbox
  bodies of nearby NPCs/droids). Producer-side fix: #508.
- **Upstream**: dimforge/rapier#961 (open, 2026-07-16; same signature on parry 0.25/0.28,
  unfixed on parry master as of 0.29) - a parry upgrade alone cannot fix it. Verified
  with a 60-line standalone rapier 0.31 program: one NaN `linvel` on one of 50 bodies ->
  the exact panic ~40 frames later.

## This PR: detection & attribution only

Per review direction, this layer performs **no band-aid repairs** - it does not prevent
the crash by itself. `PhysicsWorld::report_nonfinite_rigid_body_state` runs immediately
before every physics step and logs an ERROR naming the exact entity and offending state
at the **first bad frame**, once per entity, so the parry crash a few frames later is
fully attributed to its producer:

```
ERROR shock2vr::physics: [physics] entity Some(EId(96.0)) (Dynamic body at [[98.56, -5.97, -46.39]]):
  non-finite rigid-body state: linvel [[NaN, -0.48, -0.015]] - this poisons the broad-phase BVH
  and will panic parry's binned rebuild within a few frames (#506; reported once per entity)
```

Checked fields: `linvel`, `angvel`, `pose`, and (for kinematic bodies) the `next
kinematic pose`. The actual crash is eliminated by fixing the producer (#508, stacked
on this branch); together they close #506.

## Verification

**Negative-first e2e** (`tools/shock2-sdk/test/nonfinite-body-report.e2e.test.ts`,
station.mis): poison **every dynamic body**'s velocity to NaN through the real debug
surface (two opposite overflow-to-inf impulses via `/v1/physics/bodies/:id/impulse`;
animated creatures get their velocity overwritten by animation locomotion each frame,
non-animated items keep the poison), step one frame, then assert:
- an ERROR report naming `linvel` fires (polled via `game.logs()`), and
- the poisoned state is **not** mutated (still non-finite afterwards).

Without this change: fails 1/1 (timeout waiting for the report). With it: **5/5** runs
pass.

**Regressions** (final build): `RUSTFLAGS="-D warnings" cargo check -p shock2vr
-p desktop_runtime -p debug_runtime` clean; `missions.e2e` 23/23; SDK unit tests pass;
`cargo fmt` applied.

Part of #506

### DIFF
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -408,6 +408,11 @@ pub struct PhysicsWorld {
     // Sensor Intersection List
     player_sensor_intersections: HashSet<EntityId>,
 
+    // Entities already reported by report_nonfinite_rigid_body_state, so a
+    // body fed bad state every frame (e.g. NaN animation joints driving a
+    // kinematic hitbox) is reported once instead of every frame.
+    reported_nonfinite_entities: HashSet<Option<EntityId>>,
+
     // Collision Events
     events: PhysicsEvents,
 }
@@ -1142,6 +1147,8 @@ impl PhysicsWorld {
 
             player_sensor_intersections: HashSet::new(),
 
+            reported_nonfinite_entities: HashSet::new(),
+
             events: PhysicsEvents::new(),
         }
     }
@@ -1192,11 +1199,88 @@ impl PhysicsWorld {
         debug_renderer.render()
     }
 
+    /// Report (once per entity, ERROR level) any rigid body whose state went
+    /// non-finite before it reaches the broad-phase. Detection only - no
+    /// state is mutated.
+    ///
+    /// A NaN velocity integrates into a NaN pose, which puts a NaN collider
+    /// AABB into the broad-phase BVH. parry's binned rebuild
+    /// (`bvh_binned_build.rs`) bins leaves by their AABB centers, and a NaN
+    /// center silently truncates the computed centroid range (NaN drops the
+    /// accumulated min/max), so valid leaves land outside it and the bin
+    /// index goes out of bounds - possibly many frames later, whenever the
+    /// incremental optimizer happens to rebuild the poisoned subtree
+    /// (issue #506; upstream: dimforge/rapier#961, still unfixed as of parry
+    /// 0.29). The creation-time size sanitizers above can't catch this: the
+    /// state goes bad at *runtime* (e.g. NaN animation joints driving a
+    /// kinematic hitbox, #508). This report names the exact entity and the
+    /// offending field(s) at the first bad frame, so the parry crash that
+    /// follows a few frames later is fully attributed; the fix belongs at
+    /// the producer's source, not here.
+    fn report_nonfinite_rigid_body_state(&mut self) {
+        fn finite_pose(pose: &Isometry<Real>) -> bool {
+            pose.translation.vector.iter().all(|c| c.is_finite())
+                && pose.rotation.coords.iter().all(|c| c.is_finite())
+        }
+
+        for (_handle, body) in self.rigid_body_set.iter() {
+            if !body.is_enabled() {
+                continue;
+            }
+
+            let linvel_bad = !body.linvel().iter().all(|c| c.is_finite());
+            let angvel_bad = !body.angvel().iter().all(|c| c.is_finite());
+            let pose_bad = !finite_pose(body.position());
+            let next_pose_bad = body.is_kinematic() && !finite_pose(body.next_position());
+
+            if !(linvel_bad || angvel_bad || pose_bad || next_pose_bad) {
+                continue;
+            }
+
+            let entity_id = body
+                .colliders()
+                .first()
+                .and_then(|c| self.collider_set.get(*c))
+                .and_then(|c| EntityId::from_inner(c.user_data as u64));
+            // A body fed bad state every frame (e.g. NaN animation joints
+            // driving a kinematic hitbox) is only reported once.
+            if !self.reported_nonfinite_entities.insert(entity_id) {
+                continue;
+            }
+
+            let mut bad_fields = Vec::new();
+            if linvel_bad {
+                bad_fields.push(format!("linvel {:?}", body.linvel()));
+            }
+            if angvel_bad {
+                bad_fields.push(format!("angvel {:?}", body.angvel()));
+            }
+            if pose_bad {
+                bad_fields.push(format!("pose {:?}", body.position()));
+            }
+            if next_pose_bad {
+                bad_fields.push(format!("next kinematic pose {:?}", body.next_position()));
+            }
+            tracing::error!(
+                "[physics] entity {:?} ({:?} body at {:?}): non-finite rigid-body state: {} - this poisons the broad-phase BVH and will panic parry's binned rebuild within a few frames (#506; reported once per entity)",
+                entity_id,
+                body.body_type(),
+                body.translation(),
+                bad_fields.join(", "),
+            );
+        }
+    }
+
     pub fn update(
         &mut self,
         desired_movement: Vector3<f32>,
         player_handle: &mut PlayerHandle,
     ) -> (Vector3<f32>, Vec<CollisionEvent>) {
+        // Attribute any non-finite body state to its entity before the step
+        // consumes it (see report_nonfinite_rigid_body_state) - by the time
+        // parry panics, the culprit is already named in the log.
+        self.report_nonfinite_rigid_body_state();
+
         /* Run the game loop, stepping the simulation once per frame. */
         profile!(scope: "physics", level: TRACE, "physics.step", {
             self.physics_pipeline.step(
--- a/tools/shock2-sdk/test/nonfinite-body-report.e2e.test.ts
+++ b/tools/shock2-sdk/test/nonfinite-body-report.e2e.test.ts
@@ -1,0 +1,105 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import { GameServer } from "../src/index.js";
+
+// End-to-end test for the non-finite body-state DETECTION layer (#506).
+//
+// A rigid body whose velocity goes non-finite integrates into a non-finite
+// pose, which puts a NaN AABB into rapier's broad-phase BVH - and parry's
+// binned rebuild (bvh_binned_build.rs) then panics with "index out of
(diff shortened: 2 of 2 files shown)


tommy-xr:
Layering note (from campaign review discussion): this PR is **containment + detection**, not the fix for the bad data. Its value is (a) the engine can no longer be killed by any single non-finite body state (upstream parry panic, dimforge/rapier#961), and (b) it turns a delayed, unattributable crash into a once-per-entity log line naming the offender — which is how the actual producer was localized to the creature-hitbox/animation path.

The root-cause fix is being done at the source in our own animation → `RuntimePropJointTransforms` → `hit_boxes.rs` pipeline (issue #508, PR stacked on this branch): a failed motion query must never emit NaN joint transforms (hold last-good/bind pose). Once that lands, the sanitizer here should be a silent safety net in practice — if its warning ever fires again, that's a new producer to hunt, not noise to ignore.

tommy-xr:
Rescoped per review direction: **detection/attribution only — no band-aid repairs.**

- `sanitize_rigid_body_state` -> `report_nonfinite_rigid_body_state`: all state mutation removed (no velocity zeroing, no kinematic-pose resets, no body disabling). It now only logs, once per entity, at ERROR level, naming the entity id, body type, position, the offending field(s) (`linvel`/`angvel`/`pose`/`next kinematic pose`) and their values — so when a producer exists, the (still-inevitable) parry crash a few frames later is fully attributed at the first bad frame. Sample:

  ```
  ERROR shock2vr::physics: [physics] entity Some(EId(96.0)) (Dynamic body at [[98.56, -5.97, -46.39]]): non-finite rigid-body state: linvel [[NaN, -0.48, -0.015]] - this poisons the broad-phase BVH and will panic parry's binned rebuild within a few frames (#506; reported once per entity)
  ```

- Test reworked to detection semantics (`nonfinite-body-report.e2e.test.ts`, replaces `bvh-nan-guard.e2e.test.ts`): poison every dynamic body's velocity to NaN via the debug impulse surface, step one frame, assert the ERROR report fires naming `linvel` **and** that the poisoned state was NOT mutated. Fails without the change (timeout waiting for the report), passes 5/5 with it. The station-walk survival assertion moved out — it belongs to the #508 producer-fix PR stacked on this branch.
  - Note for reviewers: poisoning *every* dynamic body (not just one) is load-bearing — animated creatures get their velocity overwritten by animation locomotion each frame (`mission_core.rs` `update_animations`), which silently clears the poison; only non-animated items keep it.

- Title/body updated; "Fixes #506" -> "Part of #506" (the #508 root fix + this layer together close it).

Re-validated: `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean, `cargo fmt` applied, `missions.e2e` 23/23, SDK unit tests green. Force-pushed as `20ea791`.

PR #534, opened 2026-07-23
fix(ai): fall back to authored idle stand motions

### LINKED ISSUE #382: Idle droids spin in a failed 'idlegesture' motion-query loop (12k+ warnings/minutes, no idle animation)
Found while investigating wound animations with the debug runtime on medsci1.

The maintenance droid's IdleBehavior queries \`idlegesture(required) + droid(optional) + maintenance(optional)\`, which matches nothing in the motion database (\`cargo dq motion 2 +idlegesture +droid\` → no results; the DB has idlegesture clips for humans/grunts but none reachable for droids). The failure path in \`mission_core\` dispatches \`AnimationCompleted\` so the script isn't left waiting — but for an idle droid that immediately re-queues the same failing query, producing a tight per-frame loop:

- **12,783** \`WARN Unable to find animation for query ... idlegesture, droid, maintenance\` in a few minutes of one session (one idle droid ≈ 60 warnings/sec)
- the droid stands with no idle animation the whole time

**Fix directions** (either/both): give IdleBehavior a fallback query for creatures without idlegesture clips (e.g. plain stand/locomote-stand tag), and/or rate-limit or de-duplicate the failed-query → AnimationCompleted → re-query cycle so a missing motion doesn't become a per-frame loop.

tommy-xr:
Fixes #382

## Summary

- add a behavior-level ordered motion-query list that defaults to each behavior's existing single query
- keep `idlegesture` as the preferred idle schema and retry with authored `stand` only when it has no match
- pass behavior alternatives through every AI animation start/requeue path, reusing the existing first-match-wins schema resolver
- add unit and debug-runtime regressions for the query order and the native `medsci1` Maintenance droid

This is data-driven rather than creature-specific: humans and other creatures with authored `idlegesture` clips keep selecting them, while droids (and any other creature without one) can use their authored stand set.

## Negative-first evidence

At base `7fa7fa3bd57e36834b717ca9c0b463018a00ccec`, with only the regression scaffold present and before the idle override/call-site implementation:

```text
$ RUSTFLAGS="-D warnings" cargo test -p shock2vr idle_prefers_gesture_then_stand -- --nocapture

assertion `left == right` failed
  left: [["idlegesture"]]
 right: [["idlegesture"], ["stand"]]

test result: FAILED. 0 passed; 1 failed
```

After the implementation, the same test passes. The runtime regression also discovers the mission entity by name on every launch, verifies its current clip is one of `mbtstd` / `mbtfake1` / `mbtfake2`, and verifies no `idlegesture` query warning is emitted.

Motion database checks independently pin the data:

```text
cargo dq motion 2 +idlegesture +droid +maintenance
No animations found

cargo dq motion 2 +stand +droid +maintenance
mbtstd, mbtfake1, mbtfake2

cargo dq motion 0 +idlegesture +human --limit 5
29 matching animations
```

## Validation

- `RUSTFLAGS="-D warnings" cargo test -p shock2vr` — 191 passed, 0 failed
- Node 22 `npm test` in `tools/shock2-sdk` — 121 discovered; 14 fast tests passed, 107 e2e-gated tests skipped
- Node 22 focused `motion-query-tags.e2e.test.ts` — 2/2 passed
- Node 22 full `npm run test:e2e` — 121/121 passed, 0 failed/skipped/cancelled (25m10s)

## Visual evidence

Maintenance droid authored idle animation

<sub>The idle Maintenance droid now advances through authored stand motions (`mbtfake1` → `mbtfake2` → `mbtstd`) instead of remaining frozen. Captured headlessly with deterministic fixed-timestep stepping.</sub>

### Before → after

Maintenance droid before and after

Before and after used separate runtime binaries (base `7fa7fa3` versus this change), independently discovered runtime entity IDs, and the same teleport/look/step sequence. Baseline stayed at `clip=null`, frame 0 across both samples; after advanced from `mbtfake1` frame 40 to frame 70. Hosted media and telemetry: [gist](https://gist.github.com/tommy-xr/45a96bbc901cbb440c83e505570081da).

## Review

xreview scope was `origin/main` plus the exact four-file change and all visual artifacts.

- Codex reviewer: no findings; visual and code paths verified; `VERDICT: SHIP`
- independent fallback reviewer: Critical 0, High 0, Medium 0, Low 0; visual motion and first-match behavior verified; `VERDICT: SHIP`
- synthesis: **SHIP — 0 agreed Critical/High findings**
- duplication evidence: 3,080 functions indexed; no actionable semantic duplication signal

The configured Claude Opus CLI was attempted for the second engine but was not authenticated in this environment (`loggedIn: false`), so the required second pass used an independent read-only reviewer agent instead. The token-clone subprocess was also unavailable; the semantic duplication pass completed.

### DIFF
--- a/shock2vr/src/scripts/ai/animated_monster_ai.rs
+++ b/shock2vr/src/scripts/ai/animated_monster_ai.rs
@@ -490,7 +490,7 @@ impl AnimatedMonsterAI {
             sync_effect,
             Effect::PlayAnimationBySchema {
                 entity_id,
-                motion_queries: vec![self.current_behavior.borrow().animation()],
+                motion_queries: self.current_behavior.borrow().animation_queries(),
                 selection_strategy,
             },
         ])
@@ -697,7 +697,7 @@ impl Script for AnimatedMonsterAI {
         let selection_strategy = self.next_selection(is_locomotion);
         let animation_effect = Effect::QueueAnimationBySchema {
             entity_id,
-            motion_queries: vec![self.current_behavior.borrow().animation()],
+            motion_queries: self.current_behavior.borrow().animation_queries(),
             selection_strategy,
         };
 
@@ -871,7 +871,7 @@ impl Script for AnimatedMonsterAI {
                         let selection_strategy = self.next_selection(is_locomotion);
                         Effect::PlayAnimationBySchema {
                             entity_id,
-                            motion_queries: vec![self.current_behavior.borrow().animation()],
+                            motion_queries: self.current_behavior.borrow().animation_queries(),
                             selection_strategy,
                         }
                     } else {
@@ -918,7 +918,7 @@ impl Script for AnimatedMonsterAI {
                 let selection_strategy = self.next_selection(is_locomotion);
                 return Effect::PlayAnimationBySchema {
                     entity_id,
-                    motion_queries: vec![self.current_behavior.borrow().animation()],
+                    motion_queries: self.current_behavior.borrow().animation_queries(),
                     selection_strategy,
                 };
             }
@@ -948,7 +948,7 @@ impl Script for AnimatedMonsterAI {
                 let selection_strategy = self.next_selection(is_locomotion);
                 Effect::PlayAnimationBySchema {
                     entity_id,
-                    motion_queries: vec![self.current_behavior.borrow().animation()],
+                    motion_queries: self.current_behavior.borrow().animation_queries(),
                     selection_strategy,
                 }
             } else {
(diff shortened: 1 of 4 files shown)


tommy-xr:
Required xreview completed on exact head `52a54a42` against current `origin/main`:

- **Codex:** no findings.
- **Claude Opus:** no Critical/High/Medium findings; Low notes only around opt-in e2e coverage, a latent future mixed attack/non-attack fallback concern, and a brittle mission-specific test lookup.
- The ordered fallback was verified against the existing first-match motion-query executor; all non-idle behaviors retain their prior single-query semantics.
- All five CI jobs are green, the branch is merge-clean, and the recorded full e2e run passed 121/121.

**Synthesized verdict: ship.** Marking ready for review.

PR #537, opened 2026-07-23
fix(debug): reject concurrent step requests

### LINKED ISSUE #358: debug_runtime: 503 'game thread may have crashed' is ambiguous - covers dropped step replies, not just thread death
## Symptom

Intermittently at the START of a full SDK e2e suite run (`SHOCK2_E2E=1 npm run test:e2e`), the first one or two tests fail with:

```
POST /v1/step failed with status 503: Game loop unavailable - the game thread may have crashed
```

- Reproduced 3/3 on full-suite runs in one session; **never** reproduces when the same test files run in isolation (0/7 attempts), including the exact failing pair.
- The failing runtime's captured logs show a fully successful init ("Game initialized successfully"), ~1s of normal stepping, then silence — **no panic, and neither of the loop's exit log lines** ("Game loop ended due to ...") appears.
- Launch logs showed `Blocking waiting for file lock on package cache` — another cargo (likely rust-analyzer) was competing at suite start, so the machine was under heavy load during the failures.

## Two harness gaps worth fixing

1. **The 503 message conflates two very different failures.** `game_loop_unavailable()` is returned both when `command_tx.send()` fails (game thread genuinely gone) *and* when `reply_rx.await` errs (the game loop dropped a reply sender while very much alive). Debugging the intermittent failure above cost significant time because the message asserts a crash the logs then contradict. Suggest distinct messages ("command channel closed" vs "reply dropped mid-processing") and an `ERROR` log in the runtime when a reply sender is dropped.

2. **`pending_step_reply` is silently overwritten** (`main.rs` line ~527: `pending_step_reply = Some(reply)`): if a second Step command ever arrives while one is in flight, the first request gets the misleading 503. Nothing intentionally does this today, but nothing prevents or logs it either. Suggest replying immediately with an error (or at least `tracing::warn!`) when a pending step reply is being replaced.

Instrumenting these would make the intermittent suite-start failure self-diagnosing next time it fires.

tommy-xr:
Fixes #358

## Summary

- Reject a second `/v1/step` request with `409 Conflict` while a step is already running.
- Preserve the original request's reply sender and step counters instead of overwriting them.
- Use a typed step result/error path and distinguish command-channel closure from a reply dropped during processing.
- Add a durable SDK e2e regression test covering overlap rejection, full completion of the original 10,000-frame request, and a successful later step.

## Negative-first evidence

With only the new e2e test added on base `7fa7fa3`, the focused Node 22 run failed:

```text
not ok 1 - concurrent step is rejected without interrupting the in-flight step
error: Missing expected rejection: the overlapping step should receive a clear conflict response
code: ERR_ASSERTION
```

The competing 2-frame request incorrectly returned success, reproducing the pending-reply overwrite. After the runtime change, the same test passes and asserts:

- the competing request receives HTTP 409 with `Step already in progress`;
- the original request reports exactly 10,000 frames advanced;
- a later 2-frame request succeeds at the expected next frame index.

## Validation

- `git diff --check origin/main...HEAD`
- `RUSTFLAGS="-D warnings" cargo check -p debug_runtime -p shock2vr`
- `RUSTFLAGS="-D warnings" cargo test -p debug_runtime -p shock2vr` — 190 passed, 0 failed
- Node 22 `npm test` — 121 discovered, 14 passed, 107 e2e skipped, 0 failed
- Node 22 focused concurrent-step e2e — 1 passed, 0 failed
- Node 22 `npm run test:e2e` — 121 passed, 0 failed, 0 skipped

## Review

- Native Codex exact-diff review: no Critical, High, Medium, or Low findings; ship verdict.
- Independent read-only fallback review: no actionable findings; ship verdict.
- Claude Opus review could not run: `claude auth status` reported `"loggedIn": false`, and the invocation failed with `Not logged in · Please run /login`.
- Dupfinder indexed 3,077 functions and found no actionable semantic duplication in the changed functions. Its token-clone backend was unavailable (`jscpd produced no report`), so that limitation is recorded here.

This is not a visual change; no PR visuals are required.

### DIFF
--- a/runtimes/debug_runtime/src/commands.rs
+++ b/runtimes/debug_runtime/src/commands.rs
@@ -15,7 +15,7 @@ pub enum RuntimeCommand {
     GetInfo(oneshot::Sender<FrameSnapshot>),
 
     /// Step the simulation forward by frames or time
-    Step(StepSpec, oneshot::Sender<StepResult>),
+    Step(StepSpec, oneshot::Sender<Result<StepResult, StepError>>),
 
     /// Take a screenshot of the current frame
     Screenshot(ScreenshotSpec, oneshot::Sender<ScreenshotResult>),
@@ -191,6 +191,13 @@ pub enum StepSpec {
     Frames { frames: u32 },
     /// Step by duration
     Duration { duration: String }, // Will be parsed with humantime
+}
+
+/// Error returned when a step command cannot be started.
+#[derive(Debug)]
+pub enum StepError {
+    /// Another step request is still advancing the simulation.
+    AlreadyInProgress,
 }
 
 /// Result of stepping the simulation
--- a/runtimes/debug_runtime/src/main.rs
+++ b/runtimes/debug_runtime/src/main.rs
@@ -517,7 +517,7 @@ fn run_game_blocking(
     // - `Screenshot` captures after the frame is fully rendered (below), not at
     //   command-receive time (which is before this iteration's render/swap and
     //   would read a stale/blank back buffer -> intermittent black screenshots).
-    let mut pending_step_reply: Option<oneshot::Sender<StepResult>> = None;
+    let mut pending_step_reply: Option<oneshot::Sender<Result<StepResult, StepError>>> = None;
     let mut frames_advanced_this_step = 0u32;
     let mut pending_screenshots: Vec<(ScreenshotSpec, oneshot::Sender<ScreenshotResult>)> =
         Vec::new();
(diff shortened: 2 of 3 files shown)


tommy-xr:
Required xreview completed on exact head `f632fd6e` against current `origin/main`:

- **Codex:** no findings.
- **Claude Opus:** confirmed the overlap guard has no accept-and-clobber race and preserves the active step. It noted (1) the new regression test is opt-in rather than CI-run, and (2) a Low-severity semantic question over returning 500 rather than 503 if the game loop drops a pending reply.
- The branch records a full 121/121 e2e pass, including the new concurrent-step scenario, so the opt-in CI note does not leave this change unverified.
- All five CI jobs are green and the branch is merge-clean.

**Synthesized verdict: ship.** Marking ready for review; the 500/503 distinction can be handled separately if we want to normalize unavailable-game responses.

PR #553, opened 2026-07-24
Fix Station service hall step-up

### LINKED ISSUE #499: [gameplay] Station service hall step blocks all three training years
## Play-through blocker

The Earth → Station campaign now initializes year 1, but the authored route into the three service lanes is physically blocked in the debug/desktop movement path.

At the service hall ramp (AIPATH waypoints `(14.04,-4.80,4.70) -> (13.38,-4.80,6.40) -> (13.38,-4.80,6.60) -> (12.80,-4.80,6.70) -> (12.00,-4.50,6.80)`), the player reaches about `(13.16,-3.80,6.44)` and cannot advance. Both ordinary locomotion and bounded collision-checked `/v1/player/move` stop there. The next path cell rises only 0.3 world units, within the configured 2 ft step height.

This prevents reaching the lower service lanes at x≈1 and therefore blocks all three Station years and deployment to MedSci.

## Suspected cause

`try_step_up` always probes upward by the full `PLAYER_STEP_HEIGHT` before moving forward. This low-overhead ramp appears to have enough clearance for the actual ~0.3u rise but not for a full 0.8u lift, so the headroom cast rejects a valid small step. This is an inference from the authored cells and the repeatable stop; verify with a mission regression test.

## Repro

1. Run the Station-init branch from #498 and load `frontier_station_initialized_20260718`.
2. Resume at the already-validated service-hall frontier `(14.0,-3.8,4.7)`.
3. Move collision-checked to `(13.38,-3.8,6.4)`.
4. Apply ordinary locomotion northwest toward `(12.0,-3.5,6.8)`.
5. Observe the player stops near `(13.16,-3.80,6.44)` and never steps onto the ramp.

## Acceptance

- Add a negative-first physics regression for a sub-max-height step with low overhead.
- Preserve rejection of too-tall steps and insufficient *actual* player headroom.
- Replay Station through the ramp and reach a real service-lane tripwire without teleporting past the blocker.

tommy-xr:
Fixes #499.

## What changed

- Treat only a zero-time, nearly lateral upward capsule contact as a tangential riser contact instead of false headroom.
- Preserve continuous-sweep safety: the exception is limited to capsules whose cylindrical span covers the full lift, and the fully lifted pose must be overlap-free.
- Keep crouched movement conservative because its shorter capsule cannot prove that invariant.
- Add a connected level-trimesh regression matching Station's authored face-normal error, plus real-ceiling and capsule-shape safety coverage.
- Add a `station.mis` SDK scenario that stages before the blocker, then crosses it using ordinary thumbstick input.

## Negative-first evidence

Before the production fix:

- Rust connected-trimesh regression failed: `a 0.6u connected-trimesh riser should be stepped up, rose 0.21549273`.
- The upward cast reported a zero-time almost-pure lateral contact (`normal1 ≈ [-1, -7e-8, -1e-7]`).
- Station ordinary locomotion stalled at `(13.344, -3.796, 6.487)` instead of reaching the upper tread.

## Validation

- `cargo test -p shock2vr physics::tests:: -- --nocapture` — 13 passed
- SDK `npm test`
- Focused Station scenario, including 3/3 reliability replay
- Full SDK e2e suite — 130 passed, 0 failed
- `xreview` code + visual passes — no remaining Critical/High/Medium findings

## Visual proof

Station service-hall step traversal

<sub>Same deterministic `station.mis` thumbstick sequence on `f9bf4e1` and `4b2019c`. Baseline bumps then stalls at `y=-3.796, z=6.487`; the fixed build reaches the upper tread at `y=-3.196` and advances to `z=8.001`. Camera yaw was adjusted only after locomotion for visibility.</sub>

### Before → after

Station service-hall step before and after

### DIFF
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -122,6 +122,18 @@ fn climb_redirect(desired: Vector<Real>, toward_ladder: Vector<Real>) -> Option<
     Some(desired_h - toward_ladder * into + Vector::y() * vertical * CLIMB_SPEED_SCALE)
 }
 
+/// Whether testing a vertical capsule only at the end of `lift` also covers
+/// every side contact its rounded cap could sweep through along the way.
+///
+/// This holds when the capsule's cylindrical segment spans the whole lift: a
+/// point touched by the moving cap then still intersects the full-radius
+/// cylinder at the endpoint. It does not hold for the shorter crouched player.
+fn lifted_pose_covers_swept_cap(shape: &dyn Shape, lift: Real) -> bool {
+    shape
+        .as_capsule()
+        .is_some_and(|capsule| 2.0 * capsule.half_height() >= lift)
+}
+
 /// Step-up probe (the original engine's stair-climbing approach: probe up,
 /// forward, then down from the blocked position). Called when the player's
 /// horizontal movement was mostly blocked; returns the extra translation that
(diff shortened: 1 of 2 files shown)


tommy-xr:
Closing after the final acceptance replay disproved this fix.

The focused scenario used the wrong post-approach thumbstick vector: `[0.957, 0.29]` drives mostly world +Z and climbs the booth-side/top surface to roughly `(13.346, -3.196, 7.834)`. That is not the issue's authored northwest route toward `(12.0, -3.5, 6.8)`.

With the correct northwest intent, `[0.29, 0.957]`, PR HEAD still advances only to about `(13.160, -3.796, 6.447)` and remains stalled through 48 frames at `(13.160, -3.796, 6.440)`—the original #499 blocker. A collision-validated move toward the authored post-ramp point also remains blocked.

The visual evidence therefore demonstrated the wrong adjacent ledge, not a real service-lane traversal, and the player cannot reach mission object 771 (the real lane-1 tripwire at x≈1) through the intended route.

No merge: this PR does not satisfy #499.

PR #594, opened 2026-07-25
fix: implement TrapUnlock so locked buttons can actually be unlocked

### LINKED ISSUE #590: TrapUnlock is a no-op, so PropLocked buttons can never be unlocked
Buttons that System Shock 2 unlocks via a `TrapUnlock` trigger can never be unlocked here, because `"trapunlock"` is mapped to a no-op script. Once such a button is `PropLocked(true)` it stays locked for the whole game.

### Cause

- `shock2vr/src/scripts/mod.rs:457` — `"trapunlock" => Box::new(NoopScript::new())`
- `shock2vr/src/scripts/script_util.rs:49-54` — `is_entity_locked` returns `true` unconditionally for a `PropLocked(true)` entity with **no** `PropKeyDst`:

```rust
match v_key_dst.get(entity_id) {
    Ok(key_dst) => !quest.can_unlock(&key_dst.0),
    // Locked with no key destination - nothing can unlock it
    Err(_) => true,
}
```

There is no key to satisfy and the one mechanism that would clear the lock does nothing, so the state is terminal.

### Confirmed in play, with a control

On eng1, frobbing each of these returns `hackfail` and nothing happens:

- `Elevator Button` obj **990** (elevator doors)
- grav-lift button obj **1787**
- storage-4 button obj **948**

For **1787** the unlock path was exercised directly: its own Unlock Trap (template 1213) was fired with `TurnOn` first, and the button still refused twice afterwards.

**Control:** the unlocked twin button **1871** works fine under the identical procedure, so this is the lock check and not a general button/frob problem.

### Scope note — this is narrower than first suspected

An earlier hypothesis blamed this gap for the cargo lifts too. That was **disproved**: cargo-lift call buttons are `TweqDepressable` + `BaseButton` with a direct SwitchLink and carry **no `PropLocked` at all**, so `is_entity_locked` never gates them, and a lift was ridden bottom → top successfully. Only genuinely `PropLocked` buttons are affected.

### Expected

Implement `TrapUnlock` faithfully: when triggered, it should clear the locked state of its linked target(s) (and `TrapLock`, if present, should set it), so a button gated behind an in-world unlock trigger becomes usable — through the game's own trigger/link flow, not a special case.

Found by the automated play-through loop (Engineering iteration 2).

tommy-xr:
Fixes #590

`"trapunlock"` was mapped to `NoopScript`, and `script_util::is_entity_locked`
returns `true` unconditionally for a `PropLocked(true)` entity with no
`PropKeyDst` ("Locked with no key destination - nothing can unlock it"). Every
button System Shock 2 hands out through an in-world unlock trigger was therefore
locked for the whole game - on eng1 the elevator-door (obj 990), grav-lift (1787)
and storage-4 (948) call buttons all answered `hackfail` forever, even after
their unlock trap had fired.

## How the unlock flows

Entirely through the game's own trigger/link path - no entity id, template or
mission is special-cased anywhere:

```
QB Filter 155 -> Once Router 878 -> Unlock Trap 1213 -> SwitchLinks -> buttons 990 / 1787 / 1210
```

`TrapUnlock` is instantiated from the authored script name in the normal
`create_script` registry, reacts to the `TurnOn`/`TurnOff` that arrive over
SwitchLinks like any other trap, and acts on the targets it reads from its own
SwitchLinks via the shared `script_util::get_all_switch_links`.

## Where the locked state lives

In Dark's own `P$Locked` property, written through a new `Effect::SetLocked`
(applied in `MissionCore::handle_effects`, the pattern `SetObjectState` /
`SetVisibility` already use). That is the *same* state `is_entity_locked`
already reads, so there is no second source of truth - and because `P$Locked` is
a registered Dark property, `save_load` serializes it with the rest, so the
unlock survives save/load for free.

## Both edges

`TurnOn` unlocks, `TurnOff` locks. Both are authored: rec2 drives Unlock Trap
148 from a button (unlock) *and* from Inverter 760 on the delayed dining ambush
(re-lock) for the same card slot 241.

I did **not** add a separate `TrapLock` script. With `TurnOff` handled it would
be redundant, and no shipped mission or gamesys template references the name
(checked with `cargo dq --filter 'S$traplock'` across all 23 missions) - it would
have been dead code.

## Verification

**Negative-first** (both levels fail without the change, pass with it):

- `cargo test -p shock2vr trap_unlock` - with `"trapunlock"` reverted to
  `NoopScript`, 3 of 4 tests fail (`assertion failed: !is_entity_locked(...)`,
  `the unlock must survive save/load`); restored, all 4 pass. The tests resolve
  the script through the real `ScriptWorld::create_script` registry, apply the
  emitted effect through the real `Effect::flatten` + `set_entity_locked`, and
  assert through the real `is_entity_locked`.
- `tools/shock2-sdk/test/trap-unlock.e2e.test.ts` (new) - with the no-op
  restored the e2e fails on `the unlocked button must not refuse`; with the fix
  it passes.

**Live, on eng1 via the debug runtime** (entities resolved by mission object id,
never a hardcoded runtime id):

| step | result |
| --- | --- |
| frob grav-lift button 1787 before the trap fires | refused (`hackfail`), lift stays at y = -13.0 |
| fire Once Router 878 (the authored chain), frob 1787 again | no refusal, lift moves to y = -22.5 |
| control: button 948, whose own unlock trap 838 was never fired | still refused |
| control: unlocked twin button 1871 | still works, lift moves |
| save + load, frob 1787 | no refusal, lift moves - unlock persisted |
| control after load: button 948 | still refused |
| `TurnOff` the trap, frob 1787 | refused again |

**Checks:** `cargo fmt --check --all`,
`RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`,
`cargo test -p shock2vr` (279 passed), and the full SDK e2e suite.

## Review

Ran the two-engine adversarial review. Both engines independently confirmed the
change uses the real in-world trigger/link flow and fakes nothing, and both
raised the same Medium - that `TurnOff` should lock - which is adopted here.
Codex's Low (drop the unused `TrapLock`) is adopted too. Remaining Lows
(doc duplication, driving the e2e through the Once Router instead of the trap)
are fixed in follow-up commits.

Deferred: `P$LockCnt` (Dark's nested lock refcount) is still unparsed; the port
treats the lock as a boolean, which is sufficient for all shipped wiring.

### DIFF
--- a/shock2vr/src/mission/mission_core.rs
+++ b/shock2vr/src/mission/mission_core.rs
@@ -3713,6 +3713,13 @@ impl MissionCore {
                         self.world.add_component(entity_id, PropObjState(state));
                     }
                 }
+                Effect::SetLocked { entity_id, locked } => {
+                    crate::scripts::script_util::set_entity_locked(
+                        &mut self.world,
+                        entity_id,
+                        locked,
+                    );
+                }
                 Effect::SetQuestBit {
                     quest_bit_name,
                     quest_bit_value,
--- a/shock2vr/src/scripts/effect.rs
+++ b/shock2vr/src/scripts/effect.rs
@@ -433,6 +433,15 @@ pub enum Effect {
         state: ObjectState,
     },
 
+    /// Persistently set Dark's `P$Locked` on one live object. This is the
+    /// lock state `script_util::is_entity_locked` consults, so a button or
+    /// door unlocked by a `TrapUnlock` becomes usable. Because Locked is a
+    /// registered Dark property, mission save/load serializes the result.
+    SetLocked {
+        entity_id: EntityId,
+        locked: bool,
+    },
+
     ResetGravity {
         entity_id: EntityId,
     },
--- a/shock2vr/src/scripts/effect.rs
+++ b/shock2vr/src/scripts/effect.rs
@@ -433,10 +433,9 @@ pub enum Effect {
         state: ObjectState,
     },
 
-    /// Persistently set Dark's `P$Locked` on one live object. This is the
-    /// lock state `script_util::is_entity_locked` consults, so a button or
-    /// door unlocked by a `TrapUnlock` becomes usable. Because Locked is a
-    /// registered Dark property, mission save/load serializes the result.
+    /// Persistently set Dark's `P$Locked` on one live object - the lock state
+    /// `script_util::is_entity_locked` consults. Emitted by the in-world lock
+    /// traps; see `scripts::trap_lock`.
     SetLocked {
         entity_id: EntityId,
         locked: bool,
--- a/shock2vr/src/scripts/mod.rs
+++ b/shock2vr/src/scripts/mod.rs
@@ -45,6 +45,7 @@ mod trap_destroyer;
 mod trap_email;
 mod trap_exp_once;
 mod trap_inverter;
+mod trap_lock;
 mod trap_new_tripwire;
 mod trap_off_filter;
 mod trap_on_filter;
@@ -131,6 +132,7 @@ use self::{
     trap_email::TrapEmail,
     trap_exp_once::TrapEXPOnce,
     trap_inverter::TrapInverter,
+    trap_lock::TrapLock,
     trap_new_tripwire::TrapNewTripwire,
     trap_on_filter::TrapOffFilter,
     trap_qb_filter::TrapQBFilter,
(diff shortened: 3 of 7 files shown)


tommy-xr:
Full SDK e2e suite on this head: **139/142 pass**, including the new `trap-unlock.e2e.test.ts`.

The 3 misses are unrelated:
- `monster-death` (corpse drift) and `spawn-drop-to-floor` (ECONNREFUSED on bind) both **pass when re-run in isolation** on this exact head - broad-run timing/port flakes.
- `earth-psionic-training` (`normal world-use should physically carry booster 288`) fails **identically on unmodified `origin/main`** (verified by checking out 3cf883e in the same worktree and re-running) - pre-existing, not from this change.

All 5 CI checks are green.

Issue #605, opened 2026-07-26
ACR3 (and possibly ACR2) chambers appear to have no walkable route on hydro2


tommy-xr:
Two of Hydroponics' four Anti-Contaminant Regulator chambers appear to have no walkable route to them, so their regulators cannot be reached by honest play.

### Observed (hydro2)

- **ACR3** (object 910): its cell is flagged `BLOCKING_OBB` / no-walk. Every descent attempted from the office walkway was refused, and the ladder-base level is walled to the east.
- **ACR2** (object 913, y ≈ −6, reached by a ladder *down* from the chemical storeroom): also not reachable — though note this one is confounded by #603 (ladder descent does not work at all), so it may be reachable once descending works.

During the playtest the regulator itself was tested by teleporting in (declared in the session at steps 6 and 11). The consume path **works** once you are there — vial consumed, `acr3` bit set, modules awarded — so this is purely a question of getting to the room.

### Why it matters

The deck's completion objective (`Note_3_3`) requires feeding all four regulators. ACR1 (hydro1) and ACR3 (by teleport) both accept vials correctly, so the mechanic is sound; access is the problem.

### What to investigate

Whether this is (a) genuine level geometry with an entrance the playtest failed to find, (b) the `BLOCKING_OBB` cell flag being treated as impassable when the original engine allows it (the `nav_bridges` experimental flag already makes blocking-OBB cells passable at a cost penalty — worth testing this deck with it), or (c) a missing/blocked connection like #602 on the same deck. The playtest could not distinguish these from inside the game.

Note ACR2 should be re-tested after #603 (ladder descent) is fixed, before assuming it is unreachable.

Found by the automated play-through loop (Hydroponics iteration 5).

tommy-xr:
**ACR2 is reachable — remove it from this issue's scope.**

Verified during the #603 descent fix: walking off the upper floor at `(36.0, -0.996, -53.6)` while looking down lands the player in the ACR2 chamber at y −6.996 — and this worked **identically before and after** that fix. ACR2's route uses a single tall collider rather than a rung stack, so it was never gated on the broken descent.

My write-up above speculated that ACR2 might become reachable once descent was fixed. That was wrong in both directions: it did not need the fix, and it was never actually unreachable — the iteration-5 session simply did not find the drop-down route.

[...]

PR #609, opened 2026-07-26
fix: let a gripped ladder be climbed down, not just up (#603)

### LINKED ISSUE #603: Ladder descent does not work - pushing down climbs up, pushing back free-falls
Ladders can be climbed up but not down. Descending either climbs back up or drops the player off the ladder into a free fall.

### Repro (hydro2)

The 11-rung Sector-C ladder stack at `(59.6, y -2…6, 22.2)`.

- **Ascent works** (PR #598): y −0.96 → 1.89 → 4.80 → 5.30 → 5.39, at both 0° and −60° pitch, with a clean dismount onto the walkway.
- **Descent does not.** Pitching down −60° and pushing *forward* climbs **up** again. Pushing *backward* detaches the grip entirely and the player free-falls 5.75 u (y 4.75 → −1.00).

There was also no fall damage on landing, which may be a separate gap worth noting.

### Why it matters

Several critical paths need a ladder *descent*, not just an ascent:
- medsci1's maintenance-conduit shaft down to the eng1 exit (`Rick Ladder 16`)
- hydro2's ladder **down** to the ACR2 chamber
- hydro1's two-level corridor room

The documented control scheme is "push toward the ladder to ascend, look down + push to descend", so the intended input is being given; the redirect just does not produce downward motion. `projects/climbing.md` already flags entering a descent from the top lip as untested — this is that gap, observed in play.

Follow-up to #596 / PR #598, which fixed the *angle* problem for ascent.

Found by the automated play-through loop (Hydroponics iteration 5).

tommy-xr:
Fixes #603

Based on `main` (PR #598 has landed there as 12ba580).

## Two separate things were wrong

**1. The reported input was inverted.** `head.look`'s `pitch` is **positive down**
in this codebase — `camera_rotation` builds the head quaternion so its `-Z` is
`-camera_forward`, and the desktop runtime feeds mouse `delta_y` straight into
`pitch`, so mouse-down increments it. Screenshots confirm: `pitch +60` shows the
floor, `pitch -60` the ceiling. The same rotation drives the movement vector, so
#603's `-60°` "pitching down" produced a movement vector with `y = +0.144` —
i.e. it was pushing **up** the ladder. That is the "pushing forward climbs up
again" half of the report. Documented on the `head.look` channel and in the
playtest skill so the next agent doesn't repeat it.

**2. With the correct pitch, the descent really was broken — the ladder's own
colliders blocked the climb they exist to enable.** PR #598's climb pass casts
the redirect through the same query filter as the walk, so the rungs are
ordinary solid geometry to it. A gripped player sits *flush* against the face
(the walk pass presses them to the resting contact offset before the grip takes
over and drops the horizontal input), which puts the **horizontal cap** of
whichever rung the capsule overlaps directly in the path of the vertical cast.

Instrumented on hydro2's Sector-C stack (eleven 0.8-spaced `Rick Ladder` rungs at
`(59.6, -2.0…6.0, 22.2)`) — the climb cast returned **exactly zero**:

```
CLIMBSTEP climb=[0, -0.05, 0]  applied=[0, 0, 0]  grounded=false  pos=[59.6, 3.274, 22.61]
```

## The fix

The climb cast now runs against its own `QueryPipeline` — the movement one
`with_filter`ed by a predicate that rejects colliders carrying the `CLIMBABLE`
membership. (A predicate, not a group mask: a ladder is also an `ENTITY`, so
masking the bit out of the group filter would not exclude it.) The walk and
gravity passes are untouched, so a ladder stays solid to anyone who is not
climbing it, and the climb still collides with real geometry — the floor at the
shaft's foot, a ceiling above — which is what keeps the
`CLIMB_MIN_PROGRESS_FRACTION` walk fallback working.

The exclusion is deliberately **global** rather than scoped to the gripped
ladder: stacked rungs are separate entities, so scoping it to the rung that
supplied the grip would reinstate the bug on the rung above or below it.

Also from `/xreview`: the climb's progress test now measures **signed vertical
travel** instead of the raw translation length. The redirect is purely vertical,
so any horizontal component in the result is the character controller sliding
off an obstruction — the blocked case, which must fall back to walking, not be
counted as a climb.

## Negative-first tests (`shock2vr/src/physics/mod.rs`)

Both fail on the parent commit and pass here:

| test | without the fix |
| --- | --- |
| `player_descends_a_stacked_rung_ladder` — a gripped player mid-shaft, flush against the rungs, looks down and pushes | covered **0.45 wu in 200 frames** (y 6.50 → 6.05) instead of riding down to the floor |
| `player_enters_a_descent_from_the_top_lip` — steps off a landing flush with the ladder top, then pushes back toward the ladder while looking down | the grip caught but could not move them down, so the push shoved them **back onto the landing** (x −1.66, y +7.40) |

All 287 `shock2vr` tests pass; `cargo fmt --check --all` and
`RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`
are clean.

## Live (debug runtime, same repro, same binary before/after)

**hydro2 Sector-C stack** — climb up, then look down (`pitch +60`) and hold forward:

| | before | after |
| --- | --- | --- |
| ascent, `pitch 0` | y −0.996 → **5.2**, and falls back to −0.996 the moment input is released | y −0.996 → **5.399**, holds when released |
| descent from the top | **5.392 → 5.392** — did not move at all in 150 frames of "look down + push" | 5.399 → 4.699 → 3.949 → … → **−1.00** at a steady 3.0 wu/s (= 6 × cos 60°), clean step-off |

**hydro2 top-lip / ACR2 (#605).** The shaft down to the ACR2 chamber
(`Rick Ladder 16` at `(33.65, ·, −53.6)`, regulator obj 913 at y −6) is
**reachable** — walking off the upper floor at `(36.0, −0.996, −53.6)` looking
down drops 2.2 wu, the grip catches, and the climb carries the player to the
chamber floor at y **−6.996**. This one behaves *identically before and after*
the fix (it is a single tall collider, not a rung stack), so ACR2 was **not**
gated on the descent bug — worth noting on #605.

**Top-lip entry** (`projects/climbing.md`'s untested case), eng1 Engine Core
landing at y −11.796: walking straight off the lip is a free fall (as in the
original — you walked into a hole away from the ladder), but stepping off and
pushing **back toward** the ladder while looking down grabs it after ~0.8 wu and
descends at the climb rate: −13.384 → −13.884 → −14.884 → −15.384 → floor
−15.558.

**eng1 ascent regression — unchanged, matching #598's numbers exactly:**

- Engine Core stack (tmpls 377/413/339/414/415 at `(0.10, −16.4…−13.2, −22.95)`):
  reaches the landing at y **−11.796** from every heading tested (0°, ±10°, ±20°,
  ±30°, ±35°, ±40°); dead-on rate −15.558 → −14.858 → −13.858 → −12.858 =
  **6.0 wu/s**.
- `Rick Ladder 16` (obj 1911): −15.596 → −14.896 → −13.896 → **−12.896** =
  **6.0 wu/s** dead-on; 5.9 / 5.6 wu/s at 10° / 20° off.

## Not fixed here

- **No fall damage on landing**, noted in #603. Out of scope — it is a damage
  question, not a climbing one, and nothing in this change touches it.
- Walking straight off a shaft's lip *away* from the ladder still free-falls.
  That matches the original engine's rule (you must be touching the climbable
  face and pushing into it); the working move is to press back toward the ladder.

### DIFF
--- a/.claude/skills/playtest/SKILL.md
+++ b/.claude/skills/playtest/SKILL.md
@@ -99,6 +99,15 @@ obstruction, not a step. Only the horizontal direction of the target is used;
 gravity decides the vertical, so aim at where you want to *stand*, not at a
 point in the air.
 
+**Ladders are not walked, they're climbed** — `/v1/player/move` never grips one.
+Aim the head with `{"head.look":[yaw_deg,pitch_deg]}` and hold
+`right_hand.thumbstick:[0,1]` (push *into* the ladder). **Pitch is positive
+DOWN**: `+60` looks at the floor and descends, `-60` looks at the ceiling and
+ascends — the same rotation drives the camera and the movement, so a negative
+pitch climbs *up* no matter what you meant. To enter a descent from the top of a
+shaft, step off the lip and then push back *toward* the ladder while looking
+down; the grip catches within a step's worth of falling.
+
 **Doors block the shapecast** — you can't move through a closed one. The pattern:
 move up to the door → it trips the tripwire (or `Frob` it) → `step` and wait for
 it to open (re-screenshot) → then `move` through. That's genuine door-by-door
--- a/runtimes/debug_runtime/src/main.rs
+++ b/runtimes/debug_runtime/src/main.rs
@@ -1651,7 +1651,12 @@ fn input_state_from_context(input: &InputContext) -> commands::InputState {
 ///
 /// Recognized channels:
 /// - `head.rotation`                : `[x, y, z, w]` quaternion
-/// - `head.look`                    : `[yaw_deg, pitch_deg]` convenience (forward = -Z)
+/// - `head.look`                    : `[yaw_deg, pitch_deg]` convenience (forward = -Z).
+///   **Pitch is POSITIVE DOWN** - `+60` looks at the floor, `-60` at the ceiling
+///   - matching the desktop runtime, where mouse-down increments pitch. The same
+///   rotation drives movement, so on a ladder `+pitch` descends and `-pitch`
+///   ascends; assuming the opposite is what made issue #603 read as "descent is
+///   broken".
 /// - `{left,right}_hand.trigger`    : number in [0, 1] (alias `trigger_value`)
 /// - `{left,right}_hand.squeeze`    : number in [0, 1] (alias `squeeze_value`)
 /// - `{left,right}_hand.a`          : number in [0, 1] (alias `a_value`)
--- a/runtimes/debug_runtime/src/main.rs
+++ b/runtimes/debug_runtime/src/main.rs
@@ -1655,8 +1655,8 @@ fn input_state_from_context(input: &InputContext) -> commands::InputState {
 ///   **Pitch is POSITIVE DOWN** - `+60` looks at the floor, `-60` at the ceiling
 ///   - matching the desktop runtime, where mouse-down increments pitch. The same
 ///   rotation drives movement, so on a ladder `+pitch` descends and `-pitch`
-///   ascends; assuming the opposite is what made issue #603 read as "descent is
-///   broken".
+///   ascends - a test that pitches `-60` to "look down" at a ladder climbs UP,
+///   which is how issue #603's repro was written.
 /// - `{left,right}_hand.trigger`    : number in [0, 1] (alias `trigger_value`)
 /// - `{left,right}_hand.squeeze`    : number in [0, 1] (alias `squeeze_value`)
 /// - `{left,right}_hand.a`          : number in [0, 1] (alias `a_value`)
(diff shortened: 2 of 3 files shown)


tommy-xr:
## Full SDK e2e suite: 140/142

`cd tools/shock2-sdk && npm run test:e2e` — including `missions.e2e.test.ts`, which
loads **all 23 missions** cleanly, and the pre-existing `flat climbing: pushing
into a ladder ascends it`, which passes.

Both failures are accounted for and neither is caused by this change (nothing here
is reachable from AI or inventory code — `step_player_movement` is called only from
`move_player` / `move_player_validated`, and the latter passes `climb: None`):

| failure | verdict |
| --- | --- |
| `no AI freezes mid-route after alertness churn (#481)` | **flaky under load** — passes in isolation on this branch (95.2 s, 1/1). AGENTS.md notes AI path queries run on a worker thread and jitter between runs. |
| `Earth Psionic Training reduces, refills, casts, and cleans up through real objectives` | **pre-existing on `main`** — fails identically with `shock2vr/src/physics/mod.rs` + `runtimes/debug_runtime/src/main.rs` reverted to `origin/main` and rebuilt: `AssertionError: normal world-use should physically carry booster 288`, same assertion, same line. Worth its own issue. |

PR #625, opened 2026-07-26
fix: preserve carried item archetypes across missions

### LINKED ISSUE #624: [gameplay] Cross-mission carried ammo cannot reload because source mission object IDs are resolved in destination hierarchy
## Summary

A concrete item carried across a level transition retains its positive source-mission object ID in `PropTemplateId`. Class-based gameplay in the destination then resolves that positive ID through the destination mission hierarchy, where the same numeric object ID can describe an unrelated object.

In the Engineering → Operations campaign, eng1 object 1496 is a Small Standard Clip descended from archetype -1358. In ops2, positive object 1496 is a Pellet Shot Box. Reload therefore rejects the carried standard clip even though it still has `AmmoScript` and `StackCount 6`.

## Reproduction

1. Obtain the concrete pistol (eng1 object 1495, Pistol archetype -17) and Small Standard Clip (eng1 object 1496, archetype -1358) normally.
2. Carry them through the authored transition into ops2.
3. Save and load in ops2.
4. Empty/wield the pistol and trigger the normal `Reload` input action.

Observed: pistol remains at Ammo 0 and reserve remains at StackCount 6.
Expected: pistol loads 6 rounds and the reserve clip is depleted.

Play-through evidence: `/tmp/claude/playthrough/ops-iteration-11c/report.html`; screenshot `pt-02-combat-end.png`.

## Root cause

`mission/reload.rs::load_from_reserve` classifies reserve items using `PropTemplateId` plus the current mission's `GlobalTemplateHierarchy`. For concrete objects, `PropTemplateId` is the positive mission object ID, not a globally stable archetype. Positive IDs are only meaningful in their source mission and can alias unrelated destination objects.

## Acceptance criteria

- Preserve a carried concrete item's stable canonical archetype identity across transitions and save/load without discarding its concrete mission-object identity.
- Class-based reload uses that stable identity and is not hardcoded to Operations or specific ammo IDs.
- Existing saves without the new identity metadata continue to deserialize.
- Negative-first coverage reproduces a source/destination positive-ID collision and proves the pistol loads 6 rounds/depletes reserve.

tommy-xr:
## Summary

- preserve each entity's nearest gamesys archetype separately from its concrete, mission-local `PropTemplateId`
- serialize/remap that canonical identity across level transitions and save/load, with a default for older saves
- use the stable class identity for reserve-ammo reload, psi-amp HUD classification, and projectile hit-spang hierarchy checks

Concrete mission object identity remains unchanged; this adds the missing globally stable class/provenance identity rather than rewriting object IDs or hardcoding Operations ammo.

## Regression coverage

The negative regression models the real eng1 → ops2 collision: carried positive object `1496` has canonical Small Standard Clip archetype `-1358`, while destination hierarchy aliases positive `1496` to unrelated Pellet Shot Box `-42`. Reload must load six pistol rounds and deplete the reserve.

Also covers canonical identity remapping through `EntitySaveData` and older-save serde defaults.

## Validation

- `cargo test -p shock2vr reload_uses_preserved_archetype_when_concrete_id_collides_across_missions`
- `cargo test -p shock2vr instantiate_restores_canonical_template_on_the_remapped_entity`

Fixes #624

### DIFF
--- a/shock2vr/src/hud/virtual_arms.rs
+++ b/shock2vr/src/hud/virtual_arms.rs
@@ -157,12 +157,7 @@ pub(crate) fn get_wielded_psi_power(world: &World) -> Option<(String, i32)> {
 
     // Is the wielded weapon the psi amp? Resolved via its template's class
     // tags, the same mechanism as `get_wielded_ammo_type`.
-    let template_id = world
-        .borrow::<View<dark::properties::PropTemplateId>>()
-        .ok()?
-        .get(weapon)
-        .ok()?
-        .template_id;
+    let template_id = crate::scripts::script_util::entity_class_template_id(world, weapon)?;
     let class_tags = world
         .borrow::<UniqueView<crate::mission::mission_core::GlobalTemplateClassTags>>()
         .ok()?;
--- a/shock2vr/src/mission/entity_creator.rs
+++ b/shock2vr/src/mission/entity_creator.rs
@@ -620,6 +620,19 @@ pub fn initialize_entity_with_props(
     ancestors.push(template_id);
 
     world.add_component(entity_id, PropTemplateId { template_id });
+    // A concrete object's positive ID belongs to this mission only. Preserve
+    // its nearest gamesys archetype separately so carried objects retain a
+    // stable class identity in later missions whose positive IDs may collide.
+    let canonical_template_id = ancestors
+        .iter()
+        .rev()
+        .copied()
+        .find(|ancestor| *ancestor < 0)
+        .unwrap_or(template_id);
+    world.add_component(
+        entity_id,
+        RuntimePropCanonicalTemplateId(canonical_template_id),
+    );
 
     for parent_id in ancestors {
         let maybe_parent_props = entity_info.entity_to_properties.get(&parent_id);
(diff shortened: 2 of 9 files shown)


tommy-xr:
Legacy-save caveat discovered during campaign replay audit: PR as currently written preserves canonical identity for newly-created/saved entities, but an existing save predating `canonical_template_ids` defaults that map empty. Such held entities instantiate from serialized properties without source-mission rehydration, so the old Operations frontier still falls back to colliding positive ID 1496. A conservative migration should resolve the saved effective `P$SymName` against a unique negative gamesys template before instantiation (frontier values are uniquely `Pistol`→-17 and `Small Standard Clip`→-1358); ambiguous/unmatched names must remain unresolved. Do not merge/replay the legacy frontier until follow-up migration is added.

tommy-xr:
Legacy-save migration added in bb50069. It derives an exact-name index containing only unique negative gamesys templates during the existing gamesys hydration pass, then backfills missing held canonical IDs from serialized effective P$SymName before instantiation. Ambiguous, unmatched, and case-mismatched names remain unresolved. The current frontier values migrate generically: Pistol→-17 and Small Standard Clip→-1358. Regression begins with an empty canonical map and destination ID collision, then verifies Reload loads 6 and depletes reserve.

PR #644, opened 2026-07-27
fix: persist translating door state across saves

### LINKED ISSUE #643: [save][gameplay] Open translating doors restore closed and seal the return route
## Reproduction

In ops4.mis, open translating door mission 676 on the authored return route, cross it, save beyond the doorway, restart the runtime, and load that save. The door restores at its closed pose/state instead of the saved open endpoint, sealing the reverse route.

## Root cause scope

Translating-door runtime movement updates physics/render state but does not persist the live door state/base location/pose needed by save/load. StdDoor initialization consequently restarts from authored closed data.

## Expected

Persist PropTranslatingDoor state and live base location as doors move and reach endpoints. A settled-open door restores exactly open; settled-closed restores exactly closed; a moving door restores at its saved position and resumes toward the appropriate destination. Keep the implementation generic—no mission or tripwire special cases.

## Acceptance

- Negative-first unit/save round trips for settled-open, mid-opening resume, and closed symmetry.
- Exact Ops4 e2e: traverse door 676, save beyond it, fresh runtime load, reverse-cross the still-open doorway.
- Warning-free core/runtime checks and full relevant SDK e2e validation.

tommy-xr:
Fixes #643.

## Summary

- persist each translating door's live `state` and `base_location` while it moves
- persist exact open/closed endpoint states when motion settles
- restore settled endpoints exactly and resume saved opening/closing motion toward its authored destination
- add focused unit coverage plus an Ops4 fresh-process save/load traversal for tripwire 677 / door 676

## Why

Door motion previously changed only the rendered model and physics body. Save serialization therefore retained the authored `PropTranslatingDoor`, and a fresh load initialized an open door back at its closed endpoint. On the reviewed Ops4 return path this sealed the doorway behind the player.

The fix keeps the serialized translating-door property synchronized with the existing runtime motion. It does not special-case Ops4 or its tripwire.

## Validation

- `cargo test -p shock2vr` — 312 passed
- `npm test` in `tools/shock2-sdk` — 20 passed, 132 opt-in e2e skipped
- exact `ops4-door-save.e2e.test.ts` — passed across two fresh runtime processes
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p debug_runtime` — passed

Visual evidence will be added after the deterministic capture finishes.

### DIFF
--- a/shock2vr/src/scripts/std_door.rs
+++ b/shock2vr/src/scripts/std_door.rs
@@ -1,7 +1,7 @@
 use cgmath::{InnerSpace, Vector3, Zero};
 use dark::properties::PropTranslatingDoor;
 use engine::audio::AudioHandle;
-use shipyard::{EntityId, Get, View, ViewMut, World};
+use shipyard::{EntityId, Get, View, World};
 use tracing::trace;
 
 use crate::{physics::PhysicsWorld, time::Time};
@@ -33,17 +33,13 @@ impl StdDoor {
             < (self.desired_position - trans_door.base_closed_location).magnitude2()
     }
 
-    fn persist_motion(
-        &self,
-        entity_id: EntityId,
-        world: &World,
-        state: i32,
-        position: Vector3<f32>,
-    ) {
-        let mut doors = world.borrow::<ViewMut<PropTranslatingDoor>>().unwrap();
-        if let Ok(door) = (&mut doors).get(entity_id) {
-            door.state = state;
-            door.base_location = position;
+    /// Door motion is persisted through the effect pipeline (applied by
+    /// `mission_core`) rather than written here - scripts stay pure.
+    fn persist_motion(entity_id: EntityId, state: i32, position: Vector3<f32>) -> Effect {
+        Effect::SetTranslatingDoorState {
+            entity_id,
+            state,
+            base_location: position,
         }
     }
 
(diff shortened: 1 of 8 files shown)


shock2vr/src/scripts/std_door.rs
```diff
@@ -33,6 +33,20 @@ impl StdDoor {
...
+        entity_id: EntityId,
+        world: &World,
+        state: i32,
+        position: Vector3<f32>,
+    ) {
+        let mut doors = world.borrow::<ViewMut<PropTranslatingDoor>>().unwrap();
+        if let Ok(door) = (&mut doors).get(entity_id) {
+            door.state = state;
+            door.base_location = position;
+        }
```
tommy-xr: Our principle with the scripts has been that they are 'pure functional' - they don't mutate the game world, just provide effects. We should refactor this to match that guideline (and let's do a check to see if that guideline actually is still followed in practice - we can search for ViewMut across the scripts probably)

tommy-xr: Good catch - fixed in d9acd154.

`StdDoor::persist_motion` no longer touches the world. There's a new `Effect::SetTranslatingDoorState { entity_id, state, base_location }` (following the `Effect::SetLocked` precedent - a script-emitted, persistently-set-one-property effect), applied in `mission_core`'s effect match next to `SetLocked`. `StdDoor` folds it into the effect it already returns at each persistence point via `Effect::Combined`, so the write still lands the same frame and the state values (3=opening, 2=closing, 1=open, 0=closed) and positions are unchanged. The unit tests that asserted on world state now either assert on the returned effect or apply the returned effect manually first (the pattern `energy_weapon.rs` tests use).

Audit of the rest of the guideline: `rg ViewMut shock2vr/src/scripts/` now shows only `script_util::debit_player_nanites` in production code, and that's invoked solely from `mission_core`'s Trainer/Replicator effect handlers (effect-application machinery, not a script). Everything else is test-only. `trigger_destroy.rs` and `internal_simple_health.rs` in this PR are already pure - they only return effects/messages.

Verified: `cargo fmt`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`, `cargo test -p shock2vr` (313 passed), and the PR's e2e tests `ops4-door-save` + `ops4-trigger-destroy` pass unchanged.

Issue #648, opened 2026-07-27
Tools other than dark_query still assume a loose (classic) install


tommy-xr:
Split out of #649 (`dark_query` 25th Anniversary support). That PR added
`shock2vr::data_files` — the data-file part of the runtime's mount list — and
routed `dark_query`'s gamesys / mission / `motiondb.bin` / motion-clip reads
through it, so `cargo dq` works against a 25th Anniversary install
(`DARK_ASSET_PATH=/path/to/ss2-25th`), where everything lives inside
`sshock2.kpf`.

These are still loose-file only:

- **`tools/bench` (`cargo bn path`)** — `load_path_database` does
  `File::open(data_root.join(mission))` (`tools/bench/src/main.rs:257`), and
  `resolve_missions --all` `read_dir`s the data root for `*.mis`
  (`tools/bench/src/main.rs:242`). On a 25AE install the first errors and the
  second silently yields an empty mission list. The single-mission case is a
  drop-in swap to `shock2vr::data_files::asset_paths`; `--all` additionally
  needs mission enumeration from the archive (the archive lists `data/*.mis`).
- **`cargo dq maps`** — reads the interface *resource family* from
  `res/intrface.crf` (`tools/dark_query/src/main.rs:643`). A 25AE install keeps
  that family in `sshock2.kpf` under `data/res/intrface/`, possibly overridden
  by the mod layers, so this needs the renderer's family mounts rather than the
  data-file ones. It currently fails with a clear message instead of panicking.
- **`tools/dark_viewer`** — the model/animation viewing path goes through
  `Game::init`, so it already handles both layouts; only `find_video_file`
  (`tools/dark_viewer/src/main.rs:192`) and the font viewer's
  `resolve_data_path` + `File::open` (`tools/dark_viewer/src/scenes/font_viewer.rs:28`)
  still stat the data root directly.
- **`tools/hitbox_analyzer`** — takes filesystem paths to `.bin`/`.cal`
  (`tools/hitbox_analyzer/src/main.rs:423,445`), so it only works against an
  unpacked `res/obj` tree.

The reusable piece for the resource families is
`shock2vr::build_25th_anniversary_mounts`, which is currently private and
requires a `Storage` bundle for its `BundleAssetPath` — factoring the bundle
mount out would let CLI tools build the full family list too.

Issue #577, opened 2026-07-25
earth-psionic-training e2e is frame-count fragile: settled Psi Boosters fail world-use pickup


tommy-xr:
`test/earth-psionic-training.e2e.test.ts` passes or fails depending on **how many sim frames elapse before the boosters are world-used**, not on the behaviour it claims to test. Once the three authored Psi Boosters settle (they drop from their authored `y ≈ 22.84` to `y ≈ 22.499` over the first ~30–45 frames), the test's fixed world-use staging picks up booster 275 but **not 288 or 289**:

```
booster 275: item=(224.776,22.499,189.437) player_after=(224.876,20.699,189.437) carried=True
booster 288: item=(225.176,22.499,189.437) player_after=(225.276,20.699,189.437) carried=False
booster 289: item=(224.374,22.499,189.437) player_after=(224.474,20.699,189.437) carried=False
```

→ `AssertionError: normal world-use should physically carry booster 288`.

The staging in `test/helpers/earth-world-use.ts` is relative to the item's *current* position, so the player/item geometry is identical in both cases — only the settled state of the boosters differs.

### Reproduction (no code change needed)

On plain `main`, drive the test's own sequence and insert 30 extra `POST /v1/step` frames after the entry tripwire and before the boosters — the pickups then fail exactly as above. Without the extra frames the same sequence passes. (Script used: teleport to each of the four cardinal approaches of tripwire template 320, `POST /v1/player/move` into it, step, then world-use templates 275/288/289 the way `earthWorldUse` does.)

### Why it matters

Anything that changes how many frames the entry-tripwire helper burns flips this test. It surfaced in #559's fix (PR #575): with faithful walking, `crossEarthTrainingTripwire` succeeds on its *third* cardinal approach instead of its first (the first approach walks off a ledge and the player falls, as a real player would), which costs 30 extra frames — enough to settle the boosters and break the pickup. The movement change is correct; the test is measuring frame count.

### Suggested fixes

- Make the world-use staging robust to a settled item (retry / re-stage, or aim from a position derived from the item's collider rather than a fixed pitch + offset), **or**
- Settle the boosters deliberately (step until their positions are stable) before the first pickup, so the test always runs against the settled state.

Worth checking first whether a *settled* booster is genuinely harder to world-use than a freshly-loaded one — if so, that is a gameplay bug rather than a test bug.

PR #649, opened 2026-07-27
fix(dark_query): read game data from a 25th Anniversary install

tommy-xr:
## Problem

`cargo dq` could not read a **25th Anniversary Edition** asset set, so all game-data research was silently forced onto a legacy unpacked install:

```console
$ DARK_ASSET_PATH=/Users/hacker/ss2-25th cargo dq entities command1.mis --limit 5
Error: shock2.gam not found. Checked these directories: ./Data, ../Data, ../../Data, .
```

## Root cause

`tools/dark_query/src/data_loader.rs` opened `<data_root>/shock2.gam` and `<data_root>/<mission>.mis` as **loose files**. A 25AE install has neither on disk — the gamesys, the missions and `motiondb.bin` all live inside `sshock2.kpf`. `paths::data_root()` already accepts such a directory (`sshock2.kpf` is one of its sentinels, added with #557), so the data root resolved fine; only the `File::open` failed.

The error message compounded it: it printed `paths::search_roots()` — the hardcoded `./Data, ../Data, …` candidate list — even though `DARK_ASSET_PATH` had resolved somewhere else entirely, so the message named directories that had nothing to do with the failure.

## Approach

The runtime has resolved these files through the asset-path layer since #557, so instead of adding a second KPF reader, the **data-file part of that mount list moves into `shock2vr::data_files`** and both the runtime and the CLI use it:

- `data_files::data_file_mounts(data_root)` — the two `sshock2.kpf` mounts that hold the gamesys/missions (`data/`) and `motiondb.bin` (`data/res/mschema/`), empty for a classic install.
- `data_files::asset_paths(data_root)` — those mounts plus the loose-file folder mount; a whole asset-path layer a CLI can build with no renderer and no bundle storage.
- `Game::init`'s 25AE mount list is **unchanged in content and order** — it now calls `data_file_mounts` for its last two mounts. `is_25th_anniversary_install()` moved into the same module and keeps its old signature via `paths::data_root()`.

`dark_query` reads the gamesys, missions, `motiondb.bin` and the `res/motions/*_.mc` clips through `data_loader::open_data_file`, which goes through that layer (mounts built once per process). The error now names the **resolved** data root and mentions `DARK_ASSET_PATH`.

`dq maps` is the one subcommand not covered: it reads the *interface resource family* from `res/intrface.crf`, which a 25AE install keeps inside the KPF behind the renderer's family mounts. It now fails with a message instead of panicking inside the archive reader — see #648 for that and the other tools.

## Verification

**Before** (base commit, 25AE install):

```console
$ DARK_ASSET_PATH=/Users/hacker/ss2-25th cargo dq entities command1.mis --limit 5
Error: shock2.gam not found. Checked these directories: ./Data, ../Data, ../../Data, .
```

**After** (same command) — identical to the legacy install's output, including the entity counts:

```console
$ DARK_ASSET_PATH=/Users/hacker/ss2-25th cargo dq entities command1.mis --limit 5
INFO Loaded 4541 entities from gamesys
INFO Loaded 917 entities from mission
INFO Merged total: 5458 entities
ID       | Type     | Names                                    | Template | Props | Links | Unparsed
---------+----------+------------------------------------------+----------+-------+-------+---------
-5223    | Template | sym:small shockwave                      | -3598    | 4     | 0     | Yes
-5220    | Template | sym:anim texture break                   | -363     | 4     | 0     | Yes
-4995    | Template | sym:Active Protocol Box                  | -2250    | 2     | 4     | Yes
-4964    | Template | sym:BlueForceField                       | -505     | 4     | 0     | Yes
-4858    | Template | sym:Big EMP Explosion                    | -382     | 6     | 1     | Yes

Showing 5 of 5458 entities (limited)
```

**Data agreement between the two installs.** Ran `entities --limit 400`, `entities --filter "*Door*"`, `templates 22`, `motion`, `motion-info`, `speech`, `sound` and `aipath` against both `DARK_ASSET_PATH=/Users/hacker/ss2-25th` and `DARK_ASSET_PATH=/Users/hacker/ss2-data-unpacked`; every output is byte-identical once log timestamps are stripped. The only exception is `templates`' *unparsed property* list, whose order comes from a `HashSet` and differs run-to-run **on the same install** (pre-existing; same 16 entries either way).

Per-subcommand, both installs:

| command | 25AE | classic |
| --- | --- | --- |
| `entities` / `templates` / `speech` / `sound` / `aipath` / `motion` / `motion-info` | ok | ok |
| `maps` | clear error (#648) | ok |

`motion-info humdie3` resolves its `.mc` clip out of the KPF and prints the same root-y curve as the legacy run.

**No runtime regression from the mount refactor.** Launched the debug runtime on `command1.mis` against **both** installs, stepped 30 frames and queried entities — both load, step and report the same objects (e.g. `Double Elevator Door`, `template_id` 142, identical position).

**Error message, after:**

```console
$ DARK_ASSET_PATH=/Users/hacker/ss2-25th cargo dq entities nosuch.mis
Error: nosuch.mis not found in the game data at /Users/hacker/ss2-25th - set DARK_ASSET_PATH to your System Shock 2 install if that is the wrong directory
```

**Automated test.** `shock2vr/src/data_files.rs` has three unit tests that need no game assets: they build a scratch classic layout (a loose file) and a scratch 25AE layout (a generated zip named `sshock2.kpf` holding `data/shock2.gam`, `data/command1.mis`, `data/res/mschema/motiondb.bin`) and assert the layer reads each file, plus that a missing name resolves to `None` rather than panicking. Negative test done first: with `data_file_mounts` stubbed to return no mounts (the old loose-file assumption), `anniversary_install_reads_data_files_out_of_the_archive` fails; with the mounts it passes.

Note that CI's test job runs `-p engine -p desktop_runtime -p dark_viewer -p dark_query -p debug_runtime -p debug_command` — **not** `-p shock2vr` — so these tests (and the other 321 shock2vr unit tests) do not run in CI today. They pass under `cargo test -p shock2vr` in ~6s with no assets; widening the CI package list looks worthwhile but is left out of this PR.

**Build + tests**

```
RUSTFLAGS="-D warnings" cargo check --all-targets -p shock2vr -p desktop_runtime -p debug_runtime \
  -p dark_query -p dark_viewer -p bench -p hitbox_analyzer -p debug_command   # clean
cargo test -p shock2vr   # 324 passed
cargo test -p dark       # 63 passed
cargo test -p dark_query # ok
cargo fmt --all -- --check  # clean
```

Plus the SDK e2e suite (`SHOCK2_E2E=1`).

## Review

`/xreview` (Claude subagent + Codex CLI) — no Critical/High findings from either engine; both independently confirmed the mount list is unchanged in content and order and that the archive's in-memory `Cursor` satisfies the same `Read + Seek` contract the chunk parsers use. Applied from their Medium/Low findings: `open_data_file` returns the reader instead of the engine's `RefCell` wrapper (dropping the `&mut *x.borrow_mut()` ceremony), `dq maps` errors instead of panicking, the now-orphaned `paths::search_roots()` is deleted, and the "no unpacked motion clip" message keeps the data root. Not changed, deliberately: a *corrupt* (as opposed to missing) archive still panics inside `ZipAssetPath`, which is pre-existing behavior shared with the runtime; and names are lowercased before lookup, matching what `AssetCache::get_raw_reader` and `mission_exists` already do.

### DIFF
--- a/shock2vr/src/data_files.rs
+++ b/shock2vr/src/data_files.rs
@@ -48,9 +48,8 @@ pub fn data_file_mounts(data_root: &Path) -> Vec<Box<dyn AbstractAssetPath>> {
 /// An asset-path layer that resolves only the raw data files - no textures,
 /// sounds or bundle storage, so a CLI tool can build it without a renderer.
 ///
-/// Use `data_root` as the base path when reading through it (that is what the
-/// loose-file mount resolves against), e.g.
-/// `AssetCache::new(data_root.to_string_lossy().into_owned(), asset_paths(data_root))`.
+/// Pass `data_root` as the `base_path` argument of `exists`/`get_reader`: that
+/// is what the loose-file mount resolves names against.
 pub fn asset_paths(data_root: &Path) -> Box<dyn AbstractAssetPath> {
     let mut mounts = data_file_mounts(data_root);
     mounts.push(AssetPath::folder(String::new()));
--- a/shock2vr/src/data_files.rs
+++ b/shock2vr/src/data_files.rs
@@ -1,0 +1,166 @@
+//! Resolving the raw game-data files - the gamesys (`shock2.gam`), the missions
+//! (`*.mis`) and `motiondb.bin` - in whichever install layout is present.
+//!
+//! A classic install keeps them loose at the data root. A 25th Anniversary
+//! Edition install has none of them on disk: everything lives inside
+//! `sshock2.kpf`, with the gamesys and missions under `data/` and `motiondb.bin`
+//! under `data/res/mschema/`.
+//!
+//! The renderer's mount list (`Game::init`) already handles both layouts; this
+//! module owns the data-file part of it so that CLI tools with no renderer (and
+//! no bundle storage) resolve exactly the same files the game does, instead of
+//! `File::open`ing the data root and failing on a 25AE install.
+
+use std::path::Path;
+
+use engine::assets::asset_paths::{AbstractAssetPath, AssetPath};
+
+use crate::zip_asset_path::ZipAssetPath;
+
+/// The single archive a 25th Anniversary Edition install ships its data in.
+const BASE_ARCHIVE: &str = "sshock2.kpf";
+
+/// Whether `data_root` is a 25th Anniversary Edition install rather than a
+/// classic one. The remaster ships its data inside `sshock2.kpf`; a classic
+/// install has loose `.crf` archives instead.
+pub fn is_25th_anniversary_install(data_root: &Path) -> bool {
+    data_root.join(BASE_ARCHIVE).exists()
+}
+
+/// The archive mounts that hold the raw data files, highest priority first.
+///
+/// Empty for a classic install, where those files are loose at the data root and
+/// the folder mount finds them.
+pub fn data_file_mounts(data_root: &Path) -> Vec<Box<dyn AbstractAssetPath>> {
+    if !is_25th_anniversary_install(data_root) {
+        return Vec::new();
+    }
+
+    let archive = data_root.join(BASE_ARCHIVE).to_string_lossy().into_owned();
+    vec![
+        // `motiondb.bin` moved from the data root to `res/mschema/`, and the
+        // missions + gamesys live under `data/`.
+        ZipAssetPath::with_prefix(archive.clone(), "data/res/mschema/"),
+        ZipAssetPath::with_prefix(archive, "data/"),
+    ]
+}
+
+/// An asset-path layer that resolves only the raw data files - no textures,
+/// sounds or bundle storage, so a CLI tool can build it without a renderer.
+///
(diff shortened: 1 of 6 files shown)


tommy-xr:
**e2e suite**: `SHOCK2_E2E=1 npm run test:e2e` → **153/155 pass**. The two failures reproduce **identically on `origin/main` with this branch checked out away** (verified by detaching to `15364911` and re-running just those two files), so they are pre-existing and unrelated to this change:

- `earth-psionic-training.e2e.test.ts` — "normal Cryokinesis projectile should damage the real Training Droid" (see also #577, same test's fragility)
- `world-aim.e2e.test.ts` — `POST /v1/load` → 500 `Failed to load 'world-aim-e2e' - corrupt or incompatible save`

Everything the change touches is covered by the passing 153, plus the direct 25AE/classic runtime smoke described in the PR body.

Issue #660, opened 2026-07-27
dark_viewer / cutscenes: 25th Anniversary videos (.ogv under cutscenes/<layer>/) are never found


tommy-xr:
Split out of #648, whose `tools/dark_viewer` bullet also named `find_video_file`
(`tools/dark_viewer/src/main.rs:192`). Unlike the other items in that issue, this
one is **not** a loose-vs-archive resolution problem — the videos are loose files
in *both* layouts, so `shock2vr::data_files` does not help — and it is not
dark_viewer-only:

- A classic install has `cutscenes/Intro.avi` at the data root, which
  `find_video_file` and the game's own `resolve_cutscene_path`
  (`shock2vr/src/scenes/mod.rs:323`, plus `is_cutscene_mission`, which keys off
  the `.avi` extension) both expect.
- A 25th Anniversary install ships **no `.avi` at all**. It has
  `cutscenes/original/intro.ogv`, `cutscenes/enhanced/{cs1,cs2,cs3,landing,
  shuttle1..3,starport,credits}.ogv` and `cutscenes/kex/*.ogv` — Theora video +
  Vorbis audio in an Ogg container, in per-layer subdirectories, with different
  basenames from the originals.

So on a 25AE install:

```console
$ cargo dv Intro.avi --debug-no-render
Could not find video file Intro.avi in any of the expected locations (searched under <data root> and current directory)
```

and the game cannot play any cutscene either.

Making this work needs more than a path tweak, which is why it is not part of
the #648 follow-up PR:

1. A shared resolver (in `shock2vr`, so the game and `dark_viewer` agree) that
   maps a requested cutscene to whatever the install actually ships: search the
   `cutscenes/<layer>/` subdirectories and accept the `.ogv` counterpart of a
   requested `.avi`. Which layer wins (`enhanced` vs `original`) is a product
   decision, not just plumbing.
2. `is_cutscene_mission` / dark_viewer's extension dispatch must accept `.ogv`.
3. The decode path needs verifying against Theora/Vorbis: `ffprobe` reports the
   files also carry a `Data: none` stream ("Unsupported codec with id 0 for
   input stream 0"), which naive stream selection can trip over.

Reproduce with `DARK_ASSET_PATH=/path/to/25th-anniversary-install`.

Issue #650, opened 2026-07-27
world-aim e2e: save then load fails with "corrupt or incompatible save"


tommy-xr:
Reproduces on `main` (verified at 15364911, no local changes):

```
cd tools/shock2-sdk
SHOCK2_E2E=1 node --test dist/test/world-aim.e2e.test.js
```

```
✖ built SDK aims at a live torso after rotated save/load and ordinary fire hits
  Error [HttpError]: POST http://127.0.0.1:8221/v1/load failed with status 500:
    Failed to load 'world-aim-e2e' - corrupt or incompatible save
    at dist/test/world-aim.e2e.test.js:78
```

The test saves and then immediately loads the same name inside one runtime, so this is not a stale save from an older build: `<data_root>/saves/world-aim-e2e.sav` is written by the run (12 KB) and rejected by the very next `/v1/load`. Either the writer is emitting something the reader rejects for this scene, or the version/compat check is too strict for it.

Found while running the full e2e suite for #649 (which is unrelated — it only touches data-file resolution for the `dark_query` CLI). The other failure in that same run is the already-tracked #577.

tommy-xr:
Verified conclusively on current main `55c39cac2a9b22d170372c494c1b0bbb8f50c822`. The responsible fix, `1cbd6879ce56842209d2ac795f858b84efc09a65`, is present via merged PR #699: https://github.com/tommy-xr/shock2quest/pull/699

Exact reported SHA `15364911c9ffd20119c92db97ea63f253a114386` reproduces the failure. Its freshly written 12,090-byte save is valid JSON but records `global_data.active_mission = "debug_psi"`; persistent load then returns HTTP 500 because the synthetic debug scene is not a mounted mission asset.

Current main passes the complete corrected authentic scenario with an isolated save root: `earth.mis` saves and reloads, Training Droid template 593 is rediscovered, its live torso is visibly acquired without fallback, and ordinary pistol fire reduces HP from 20 to 14. The focused E2E passed 1/1 in 32.396 seconds, and the SDK fast suite passed 26/26. PR #699 CI is green on Android, macOS, Ubuntu, Windows, and format.

Matched proof honestly shows the fixture change rather than implying a save-format product change: https://gist.github.com/tommy-xr/3125a2e6ce561100d89963ec0ac99964

No acceptance criterion remains unresolved: the built world-aim E2E now uses an authentic reloadable mission and completes save → load → aim → fire.

PR #661, opened 2026-07-27
fix(tools): read game data from a 25th Anniversary install in bench, dq maps and the font viewer

### LINKED ISSUE #648: Tools other than dark_query still assume a loose (classic) install

tommy-xr:
## Problem

#649 routed `dark_query` through the asset layer so `cargo dq` reads a **25th Anniversary Edition** install. The other tools still assumed loose files. The worst of them fails *silently*:

```console
# 25AE install: every mission lives inside sshock2.kpf
$ DARK_ASSET_PATH=/path/to/ss2-25th cargo bn path bench --all
mission         cells   links  found/miss   badcell  lookup p50    path p95 noroute p95  toward p95  inflate  turn deg  waypts bridges   build
$ echo $?
0
```

That is the whole output: `resolve_missions` `read_dir`s the data root for `*.mis`, finds none, and the command prints a bare header and **exits 0 having benchmarked nothing**. `AGENTS.md` tells agents to run `cargo bn path bench` before and after touching pathfinding, so a pathfinding change could be "measured" against zero missions and look fine.

The single-mission path at least says something:

```console
$ DARK_ASSET_PATH=/path/to/ss2-25th cargo bn path stats medsci1.mis
medsci1.mis: failed to open /path/to/ss2-25th/medsci1.mis
```

`cargo dq maps` and `dark_viewer`'s font viewer were loose-file only too.

## Approach

Everything reuses the `ZipAssetPath` / `AssetPath` layer — no second KPF parser.

**`tools/bench`** — missions open through `data_files::open_data_file`, and `--all` enumerates through `data_files::mission_names`, which lists the archive's `data/*.mis` as well as the directory. An empty mission list is now an error, as is a run where no mission could be read at all (`bench` and `stats` both), so a benchmark of nothing can no longer pass for a clean run.

**`data_files::open_data_file`** is `dark_query`'s `data_loader::open_data_file` helper (mounts built once per process — indexing the KPF is not free and `--all` opens 23 missions) moved into the module both tools share; `dark_query` now delegates to it instead of keeping its own copy.

**`cargo dq maps`** reads the *interface resource family*, not a data file: a classic install keeps it in `res/intrface.crf`, a 25AE one inside `sshock2.kpf` under `data/res/intrface/`, possibly overridden by a mod layer. The new `shock2vr::resource_family_paths(family)` returns the mounts for **one** family — the same archives, in the same precedence, that the game resolves that family through — so `dq maps` mounts `intrface` on either layout. On 25AE that means the mod stack is honoured (see below).

**`Game::init` is untouched.** The first cut factored its mount list into a shared builder; review pointed out that mounting everything makes any tool die inside `ZipAssetPath`'s `File::open(..).unwrap()` when an unrelated archive is missing, so the family-scoped helper replaced it and the runtime's mount list is not restructured at all. `mission_exists` drops its private copy of the archive scan (it now delegates to `mission_names`) but keeps its filesystem short-circuit, so its behaviour is unchanged.

**`dark_viewer`'s font viewer** takes the asset cache instead of `File::open`ing the data root: a path that exists on disk (as given or under the data root, both of which the old resolution accepted) still wins, otherwise a bare `MAINAA.FON` resolves out of the interface archive the way the game resolves it.

## Not in this PR

- **`find_video_file` / cutscenes → #660.** This one is *not* a loose-vs-archive problem: cutscene videos are loose files in both layouts, so `data_files` does not apply. A 25AE install ships no `.avi` at all — it has Theora/Vorbis `.ogv` under `cutscenes/{original,enhanced,kex}/` with different basenames. Supporting them means a shared resolver (the game's own `resolve_cutscene_path` / `is_cutscene_mission` have the identical gap, and fixing only the viewer would diverge them), extension dispatch for `.ogv`, and verifying the decoder against Theora plus the stray `Data: none` stream `ffprobe` reports. That is a feature, not path plumbing — filed as #660.
- **`tools/hitbox_analyzer` — deliberately unchanged, closing that bullet.** It is a "point me at a directory" analysis tool: it takes a filesystem path (default `Data/res/mesh`), globs `*.bin`, and pairs each mesh with its **sibling `.cal`** (`mesh_path.with_extension("cal")`). It never resolves a name through the data root, so there is nothing to route through the asset layer; making it archive-aware would mean re-inventing its input model (enumerate a family, resolve sibling files through mounts) for a one-off analysis tool that already works against any unpacked mesh directory on either install. The one real nit — its default `Data/res/mesh` bypasses `paths::data_root()` — is unrelated to the archive layout.

## Verification

Both installs, real data. Placeholders below stand in for the local install paths.

### bench — negative first

**Before** (base commit, 25AE install) — the silent-empty-list bug reproduced above: `bench --all` prints only the header and exits 0; `stats medsci1.mis` prints `failed to open /path/to/ss2-25th/medsci1.mis`.

**After**, same commands on the same install:

```console
$ DARK_ASSET_PATH=/path/to/ss2-25th cargo bn path stats medsci1.mis
=== medsci1.mis ===
cells: 4695  vertices: 9900  links: 27035
cell flags: unpathable=3 below_door=532 blocking_obb=366 moving_terrain=1
...
walk connectivity: 4325 pathable cells in 1683 components; largest: [1266, 260, 108, 104, 77]; isolated cells: 1523
```

**Byte-identical to the classic install's output** (`diff` of the two `stats` runs is empty), and `bench --all` benchmarks all 23 missions on both installs with **identical** cells/links/found-miss/inflate/turn/waypoint columns — only the latency columns differ, as they must:

```console
$ diff <(awk '{print $1,$2,$3,$4,$5,$10,$11,$12,$13}' bench-all-classic.txt) \
       <(awk '{print $1,$2,$3,$4,$5,$10,$11,$12,$13}' bench-all-25th.txt) && echo identical
identical
```

The empty/unreadable cases now fail loudly:

```console
$ DARK_ASSET_PATH=/path/to/empty-dir cargo bn path bench --all
Error: no .mis files found in the game data at ../../Data - set DARK_ASSET_PATH to your System Shock 2 install if that is the wrong directory   # exit 1
$ DARK_ASSET_PATH=/path/to/gamesys-only cargo bn path stats medsci1.mis
medsci1.mis: failed to open medsci1.mis in the game data at /path/to/gamesys-only
Error: no mission could be read   # exit 1
```

### dq maps

**Before**, 25AE: `Error: /path/to/ss2-25th/res/intrface.crf not found - 'maps' needs a classic install with loose .crf archives`. **After**: the map rectangles print on both installs.

On classic the output is **unchanged from the base commit** (the before/after diff is only the removed `Using intrface.crf archive: ...` line). Across installs the two outputs differ in exactly 3 of 20 rectangles — and that is **correct**: `mods/scp.kpf` ships its own `intrface/medsci1/english/p001ra.bin`, which the mod stack overrides with, i.e. what the game itself renders there. Confirmed by hashing the three copies:

| source | sha1 (first 12) |
| --- | --- |
| classic `res/intrface.crf` | `8f6e780013fb` |
| 25AE `sshock2.kpf` (base) | `8f6e780013fb` |
| 25AE `mods/scp.kpf` | `a1aa708bb4cc` |

A data root with no archives now gives a clean error instead of panicking inside the archive reader:

```console
$ DARK_ASSET_PATH=/path/to/gamesys-only cargo dq maps MEDSCI1
Error loading map data for MEDSCI1: Could not load revealed rects: MEDSCI1/english/P001RA.BIN
```

### dark_viewer font viewer

```console
$ DARK_ASSET_PATH=/path/to/ss2-25th     cargo dv MAINAA.FON --debug-no-render            # Scene creation succeeded.
$ DARK_ASSET_PATH=/path/to/ss2-classic  cargo dv MAINAA.FON --debug-no-render            # Scene creation succeeded.
$ DARK_ASSET_PATH=/path/to/ss2-classic  cargo dv res/fonts/BLUEAA.FON --debug-no-render  # Scene creation succeeded.
$ DARK_ASSET_PATH=/path/to/ss2-25th     cargo dv /abs/path/to/MAINAA.FON --debug-no-render  # Scene creation succeeded.
$ DARK_ASSET_PATH=/path/to/ss2-25th     cargo dv NOSUCH.FON --debug-no-render
Error creating scene: Could not find font NOSUCH.FON in the game data
```

Before this change the bare `MAINAA.FON` form failed on **both** installs (it was `File::open`ed relative to the data root).

### No runtime regression

`shock2vr` is touched, so: the debug runtime was launched on `command1.mis` against **both** installs, stepped 30 frames and queried entities — both load, step and report the same objects (`Double Elevator Door`, `template_id` 142/418/233/277, identical positions; runtime entity ids differ per launch as documented).

### Tests

`shock2vr/src/data_files.rs` gains 2 asset-free unit tests beside #649's three, using the same scratch-directory + generated-`sshock2.kpf` pattern: one asserts a classic layout lists its loose missions **with their on-disk spelling** (`Earth.MIS`, not `earth.mis` — the only name that opens on a case-sensitive filesystem), the other that a 25AE layout enumerates `data/*.mis` out of the archive while excluding `data/res/mschema/motiondb.bin` and anything nested deeper. Negative test first: with the archive-enumeration line disabled (the old loose-only assumption) the 25AE test fails with `left: []`; with it, green.

```
RUSTFLAGS="-D warnings" cargo check --all-targets -p shock2vr -p desktop_runtime -p debug_runtime \
  -p dark_query -p dark_viewer -p bench -p hitbox_analyzer -p debug_command   # clean
cargo test -p shock2vr   # 330 passed
cargo test -p dark       # 63 passed
cargo fmt --all
```

Plus the SDK e2e suite (`SHOCK2_E2E=1`) — result in a comment below.

## Review

`/xreview` (Claude subagent + Codex CLI), three rounds. Both engines independently raised the same two Mediums on the first cut, and both are fixed:

- **Mounting the whole game to read one family.** The first cut's `tool_asset_paths()` eagerly constructed every `.crf`/KPF mount, and `ZipAssetPath` `unwrap()`s its `File::open`, so `dq maps` against a data root missing an unrelated archive panicked where it used to fail with a message (and on 25AE it re-indexed the KPF 13+ times). Replaced by `resource_family_paths(family)`, which mounts one family with each archive existence-guarded — and which let `Game::init` be left completely alone.
- **Lowercasing loose mission names.** `mission_names` reported an on-disk `Earth.mis` under a spelling that cannot be opened on a case-sensitive filesystem. Loose missions now keep their on-disk spelling (dedup stays case-insensitive) and `open_data_file` falls back to the name as given after the lowercased lookup archives need.

Also applied from the reviews: `mission_exists` delegates its archive scan to `mission_names` instead of duplicating it; the font viewer accepts on-disk and data-root-relative paths again; `bench`/`stats` fail when no mission produced output (the same "vacuous success" one step later); `open_data_file`'s doc says plainly that `None` does not carry the underlying io error.

Not changed, deliberately: `mission_names` collapses read/zip errors into "no missions" rather than returning a `Result` (the caller's message covers it, and `mission_exists` is a `bool`); `--json` still omits per-mission load errors when at least one mission succeeds (pre-existing); and `mission_exists` keeps the pre-existing case-mismatch nuance of its filesystem short-circuit — that one belongs in the loader, not here. One nuance worth recording: `with_prefix_opts` keeps the *first* duplicate basename where the classic constructors keep the last, which cannot affect `dq maps` (it requests full relative paths) but is worth watching if another family is exposed this way.

Fixes #648. Cutscene videos split out as #660.

### DIFF
--- a/shock2vr/src/data_files.rs
+++ b/shock2vr/src/data_files.rs
@@ -12,7 +12,7 @@
 //! `File::open`ing the data root and failing on a 25AE install.
 
 use std::cell::RefCell;
-use std::collections::BTreeSet;
+use std::collections::BTreeMap;
 use std::path::Path;
 use std::sync::OnceLock;
 
@@ -51,7 +51,9 @@ pub fn data_file_mounts(data_root: &Path) -> Vec<Box<dyn AbstractAssetPath>> {
     ]
 }
 
-/// Every mission file available in `data_root`, lowercased and sorted.
+/// Every mission file available in `data_root`, sorted, with the spelling that
+/// opens it: a loose file keeps its on-disk case (which is what a case-sensitive
+/// filesystem needs), an archived one its lowercased archive name.
 ///
 /// A classic install has them loose at the data root; a 25AE install has them
 /// inside `sshock2.kpf` under `data/`, so a caller that only reads the directory
@@ -59,19 +61,22 @@ pub fn data_file_mounts(data_root: &Path) -> Vec<Box<dyn AbstractAssetPath>> {
 /// is how a "run against every mission" tool silently ends up running against
 /// nothing.
 pub fn mission_names(data_root: &Path) -> Vec<String> {
-    let mut names: BTreeSet<String> = loose_mission_names(data_root);
-    names.extend(archived_mission_names(data_root));
-    names.into_iter().collect()
+    // Keyed by lowercased name so the same mission present in both layouts is
+    // listed once, with the loose spelling winning.
+    let mut names: BTreeMap<String, String> = archived_mission_names(data_root);
+    names.extend(loose_mission_names(data_root));
+    names.into_values().collect()
 }
 
-fn loose_mission_names(data_root: &Path) -> BTreeSet<String> {
+fn loose_mission_names(data_root: &Path) -> BTreeMap<String, String> {
     let Ok(entries) = std::fs::read_dir(data_root) else {
-        return BTreeSet::new();
+        return BTreeMap::new();
     };
     entries
         .filter_map(|entry| entry.ok())
-        .map(|entry| entry.file_name().to_string_lossy().to_ascii_lowercase())
-        .filter(|name| name.ends_with(".mis"))
+        .map(|entry| entry.file_name().to_string_lossy().into_owned())
+        .filter(|name| name.to_ascii_lowercase().ends_with(".mis"))
+        .map(|name| (name.to_ascii_lowercase(), name))
         .collect()
 }
 
@@ -79,15 +84,15 @@ fn loose_mission_names(data_root: &Path) -> BTreeSet<String> {
 /// classic install, and for an archive that cannot be opened - the caller
 /// reports "no missions" either way, and the mount layer gives the real error
 /// when something then tries to *read* a mission.
-fn archived_mission_names(data_root: &Path) -> BTreeSet<String> {
+fn archived_mission_names(data_root: &Path) -> BTreeMap<String, String> {
     if !is_25th_anniversary_install(data_root) {
-        return BTreeSet::new();
+        return BTreeMap::new();
     }
     let Ok(file) = std::fs::File::open(data_root.join(BASE_ARCHIVE)) else {
-        return BTreeSet::new();
+        return BTreeMap::new();
     };
     let Ok(archive) = zip::ZipArchive::new(std::io::BufReader::new(file)) else {
-        return BTreeSet::new();
+        return BTreeMap::new();
     };
     archive
         .file_names()
@@ -95,7 +100,8 @@ fn archived_mission_names(data_root: &Path) -> BTreeSet<String> {
             let lower = name.to_ascii_lowercase();
             let relative = lower.strip_prefix(ARCHIVE_DATA_PREFIX)?;
             // Only the missions themselves, not anything nested deeper.
-            (relative.ends_with(".mis") && !relative.contains('/')).then(|| relative.to_owned())
+            (relative.ends_with(".mis") && !relative.contains('/'))
+                .then(|| (relative.to_owned(), relative.to_owned()))
         })
         .collect()
 }
@@ -112,7 +118,8 @@ pub fn asset_paths(data_root: &Path) -> Box<dyn AbstractAssetPath> {
 }
 
 /// Open a raw data file by name (`shock2.gam`, `medsci1.mis`, `motiondb.bin`)
-/// from the resolved data root, in whichever layout is present.
+/// from the resolved data root, in whichever layout is present. `None` means
+/// "no mount has that name" - the cause of a filesystem error is not reported.
 ///
 /// The mounts are built once per process because indexing a 25AE archive is not
 /// free and callers like `cargo bn path bench --all` read every mission through
@@ -120,12 +127,14 @@ pub fn asset_paths(data_root: &Path) -> Box<dyn AbstractAssetPath> {
 pub fn open_data_file(name: &str) -> Option<Box<dyn ReadableAndSeekable>> {
     static PATHS: OnceLock<Box<dyn AbstractAssetPath>> = OnceLock::new();
     let data_root = crate::paths::data_root();
-    PATHS
-        .get_or_init(|| asset_paths(data_root))
-        .get_reader(
-            data_root.to_string_lossy().into_owned(),
-            name.to_ascii_lowercase(),
-        )
+    let paths = PATHS.get_or_init(|| asset_paths(data_root));
+    let base_path = data_root.to_string_lossy().into_owned();
+    // Archive keys are lowercased, but the loose-file mount resolves a name
+    // byte-for-byte, so an on-disk `Earth.mis` only opens under its own
+    // spelling on a case-sensitive filesystem.
+    paths
+        .get_reader(base_path.clone(), name.to_ascii_lowercase())
+        .or_else(|| paths.get_reader(base_path, name.to_owned()))
         .map(RefCell::into_inner)
 }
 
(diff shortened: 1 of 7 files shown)


tommy-xr:
**SDK e2e suite** (`SHOCK2_E2E=1`, all 92 test files, run in batches):

```
157 tests, 155 pass, 2 fail
✖ Earth Psionic Training reduces, refills, casts, and cleans up through real objectives
✖ built SDK aims at a live torso after rotated save/load and ordinary fire hits
```

Both are the **known pre-existing** failures (#577 Cryokinesis damage, #650 world-aim save/load — `POST /v1/load ... corrupt or incompatible save`), and both were re-run on the base commit (`daa0558e`, immediately before this PR merged) where they fail identically: `tests 2, pass 0, fail 2`. Nothing else in the suite regressed — notably `missions.e2e.test.ts` is 23/23, and the transition / save-load tests that exercise `mission_exists` (`transition-revisit`, `transitions-trigger`, `save-load`) are 5/5.

Issue #663, opened 2026-07-27
feat(physics): carry the player on moving terrain (platform/elevator passenger transfer)


tommy-xr:
## Classification

- Domain: `[gameplay]`
- Kind: **feature gap** — moving-terrain passenger carry is absent, rather than broken logic in an implemented carry path.
- Mission: `command1.mis`
- Severity: High, hard progression blocker
- Campaign: Command · Marine (standard weapons) · 25th Anniversary assets · seed `1412905396`

## Reproduction

1. Launch `command1.mis` with `DARK_ASSET_PATH=/Users/bryan/ss2-25th`.
2. Traverse from the arrival elevator to the tram using normal frobs and collision-checked movement.
3. Recall the tram, walk fully inside it, and confirm the player is centered near the tram root.
4. Aim at the authored in-car `BaseButton` and activate it with the production squeeze edge.
5. Step exactly 60 fixed-timestep frames, sample player and tram transforms, then step another 60 frames with no input.

Validated sample:

- Before dispatch: player `(-377.79820, -16.39646, 2.80014)`; tram `(-377.53342, -15.80000, 2.72317)`.
- First 60 frames: tram moved `+12.20075` X, but player moved only `+3.63865` X and fell back into the station.
- Next 60 frames: player stayed fixed at X `-374.15955` while the tram moved another `+12.00073` X.

The earlier test where the button was pressed from the platform is explicitly excluded as operator error. This issue is based only on the later fully boarded reproduction.

## Expected

The player standing inside/on authored moving terrain inherits the tram's displacement, remains aboard, and can ride to the later stop before Bulkhead 61, as required by the original Command route.

## Actual

The kinematic tram moves away while the player remains at the departure station, so the normal route to the later stop and Bulkhead 61 cannot be completed.

![Command tram leaves boarded player behind](https://gist.githubusercontent.com/tommy-xr/003dab8b531ae652aed8620529d364f2/raw/command-tram-left-player.png)

[Full playtest data and evidence gist](https://gist.github.com/tommy-xr/003dab8b531ae652aed8620529d364f2)

## Investigation

- The tram inherits `Moving Terrain`, has authored but currently unparsed `P$MovingTer`, and runs `BaseElevator`.
- `BaseElevator::update` advances only the elevator entity through `Effect::SetPosition`.
- `MissionCore` applies that effect to the tram model/kinematic rigid body, but there is no support-motion transfer to the player controller.
[...]

PR #666, opened 2026-07-27
feat(physics): carry a player by the surface they stand on

tommy-xr:
## Summary

- track the moving-terrain body the player is **standing on** and hand its per-frame displacement to the character controller as a first movement pass
- switch off Rapier's own contact-based transfer, which drags the player off anything kinematic they merely brush and leaks ~10% of a platform's travel per second
- negative-first unit + SDK coverage for horizontal carry, vertical carry, an airborne rider, a passing wall, and the full `command1` tram leg

Follow-up to #665, which fixed the blocker in #663 (the tram's authored `PhysAttach` shell now moves with the car). This is the other half: how the *passenger* is carried.

## Root cause

#665 correctly observed that Rapier already carries a grounded player on a moving kinematic support. It does — but through `KinematicCharacterController::move_shape`, which adds `velocity_at_point * dt` for **every kinematic collider the capsule touches** (`detect_grounded_status_and_apply_friction`). That transfer is *contact-based*, not *support-based*, and it has two defects:

1. **It drags the player off anything kinematic they brush.** A wall or a sliding door moving *past* a player standing on static ground tows them along it — measured at **10.6 wu in one second** in `player_is_not_dragged_by_a_wall_they_are_only_brushing`. Nothing about that surface is underfoot.
2. **Where it does apply, it leaks.** About a tenth of the platform's travel per second is lost, so a rider slides backwards along the deck. On `command1`'s tram leg the passenger drifts **1.45 wu** — 15% of the car's length — and finishes well behind where they boarded. A longer or faster ride, or a platform without walls to catch you (every grav lift), strands them.

Measured on `origin/main`, riding the `command1` tram from station 1 (the player's offset from the car origin should not change):

| | boarded | +3 s | +6 s | +9 s | +12 s | arrival |
| --- | --- | --- | --- | --- | --- | --- |
| **before** | +0.039 | −0.349 | −0.535 | −0.729 | −0.926 | −0.715 |
| **after** | +0.039 | −0.167 | −0.167 | −0.167 | −0.167 | +0.039 |

(The constant −0.167 while moving is the one-frame reporting lag — 12 wu/s ÷ 60 — not slide; it resolves to the boarding offset the moment the car stops.)

## The fix

Each frame, probe the near-horizontal surface directly under the player, remember the kinematic body it belongs to, and next frame move the player by however far that body travelled. The displacement is **cast** through the character controller like any other movement, so a platform driving the player into geometry slides them along it rather than pushing them through it, and the player's own input and gravity resolve from where the platform left them.

Rapier's version is switched off by handing the player's movement query pipelines an **empty rigid-body set**. Both reviewers independently enumerated every `queries.bodies` read reachable from `move_shape` and the query methods used, and confirmed that lookup is the only thing affected: the autostep guard that also reads it is unreachable while `CharacterAutostep::include_dynamic_bodies` is true (Rapier's default, which we keep), and the movement filters never set the `EXCLUDE_FIXED/KINEMATIC/DYNAMIC` flags that are the only other consumer. Collision itself is untouched.

The probe casts against kinematic-bodied colliders only. Taking the nearest collider of any kind and rejecting it afterwards would let an immobile surface mask the moving one it sits flush with — which is exactly a tram deck parked at its station floor.

### Deliberately deferred

- **Rotation.** Translation only. No shipped moving platform rotates: `BaseElevator` emits `Effect::SetPosition` and nothing else, and no authored elevator path turns. Orbiting a rotating support would also have to turn the player's *facing*, which lives outside physics, so half-implementing it would be worse than not.
- **Dynamic supports.** Kinematic bodies only — the script-driven platforms, whose motion is authored and exact. The player's weight is not simulated against dynamic bodies, so carrying off a settling crate or a jittering ragdoll limb would inject that noise straight into the camera.

Both are documented on `PlayerSupport`.

## Validation

Negative-first — all of these fail on `ecb99668` and pass here:

| test | on `origin/main` |
| --- | --- |
| `player_is_carried_by_a_horizontally_moving_platform` | rode 11.0 of 12 wu |
| `player_is_carried_by_a_vertically_moving_platform` | rode 0.54 of 12 wu |
| `player_is_not_dragged_by_a_wall_they_are_only_brushing` | towed 10.6 wu |
| `moving-terrain-carry.e2e.test.ts` (tram) | passenger slid 1.449 wu along the deck |

`player_is_not_carried_while_airborne_above_a_platform` and the grav-lift e2e pass on both — they are regression guards for behavior that already worked (a rising deck simply pushes the capsule up).

- `cargo test -p shock2vr` (341 passed) and `-p dark` — includes every ladder, mantle, stair and validated-move test
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime -p dark -p dark_query -p dark_viewer -p bench`
- `cargo fmt --all --check`
- `npm test` in `tools/shock2-sdk` (25 passed)
- the new `moving-terrain-carry.e2e.test.ts` against real `command1.mis` (25th Anniversary assets); the full serial e2e suite was **not** run to completion — see the comment below
- in-game on `command1.mis` with 25th Anniversary assets: the tram carries a boarded passenger from x=−377.5 to station 2 at x=−184.4 with zero deck slide, and grav `Lift 1` (buttons 469/494) carries them up and back down in lockstep, offset from the pad constant at 1.204 at rest

## Review

`/xreview` (Claude subagent + Codex CLI), both asked explicitly whether this uses the real in-world flow or fakes the mechanism. Both answered *real flow* — no debug lever, no per-mission special case; the carry runs through the production character controller on authored moving terrain, and the e2e drives the authored tram with the production squeeze on its in-car call button. No Critical or High findings. The three Medium/Low findings (single-hit probe masking, the 0.5 slope threshold, no re-probe after relocation) are fixed in the second commit and re-validated.

### DIFF
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -93,6 +93,18 @@ const PLAYER_REST_LIFT: f32 = 0.01;
 /// without a jump.
 const PLAYER_STEP_HEIGHT: f32 = 2.0;
 
+/// How far below the player's feet (world units) a surface still counts as the
+/// thing they are *standing on* for support-motion transfer (see
+/// [`PlayerSupport`]). Comfortably clears the controller's contact offset plus
+/// the resting lift the player settles at (~0.044 wu) while staying far short
+/// of a step (0.8 wu), so a platform carries the player only while they are
+/// genuinely on it - never while they are airborne above it.
+const SUPPORT_PROBE_DISTANCE: f32 = 0.1;
+
+/// How horizontal a probed surface must be to count as support. Matches the
+/// slope test the ground pass itself uses: a wall is never something you ride.
+const SUPPORT_MIN_GROUND_NORMAL: f32 = 0.5;
+
 /// Horizontal reach (world units) for ladder detection: the player grips a
 /// climbable surface when their collider, inflated by this much radially,
 /// overlaps it. Standing flush against a ladder leaves a small gap between the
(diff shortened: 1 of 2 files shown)


tommy-xr:
## Visuals deferred — measured deltas instead

Per AGENTS.md this is visible behavior, so a GIF would normally be attached. Capturing one drives the headless debug runtime, and another playtest is holding runtimes on this machine right now; that contention has already produced spurious failures for other runs, so the capture is deferred rather than raced. For this change measured deltas are the better evidence anyway — the whole defect is a slow positional leak that a short loop would barely show.

For the same reason the full serial SDK e2e suite was not run to completion here. What *was* run: `cargo test -p shock2vr` (341 passed) and `-p dark`, `npm test` in `tools/shock2-sdk` (25 passed), and the new `moving-terrain-carry.e2e.test.ts` against real `command1.mis` on 25th Anniversary assets — both cases pass on this branch, and the tram case fails on `ecb99668`. All five CI checks are green.

### The measurement

Riding the `command1` tram from station 1. The number that matters is the **gap** — the player's x minus the car origin's x. It should not change during transit: the passenger is standing still on a deck that is 3.18 wu wide along the travel axis.

| | boarded | +2 s | +3 s | +6 s | +9 s | +12 s | arrival (station 2) |
| --- | --- | --- | --- | --- | --- | --- | --- |
| **#665 alone** (contact-based) | +0.039 | −1.951 | −0.349 | −0.535 | −0.729 | −0.926 | **−0.715 … −1.696** |
| **#666** (support-based) | +0.039 | | −0.167 | −0.167 | −0.167 | −0.167 | **+0.039** |

Two independent runs of the stock-#665 row are shown, since the leak's size depends on where in the acceleration the sample lands: my own run (teleport-boarded, sampled every 3 s) drifts steadily to −0.926 and finishes at −0.715, and a second run taken by a reviewer slips ~2.0 wu on acceleration and holds, finishing at −1.696. Either way the rider gives up most of the deck margin.

On #666 the gap is a flat −0.167 for the entire leg and returns to exactly the boarding value the moment the car stops. That −0.167 is not slip: it is the one-frame reporting lag (12 wu/s ÷ 60 Hz = 0.2, and 0.039 − 0.2 = −0.161), which is why it resolves the instant the tram is stationary.

The same holds vertically. Grav `Lift 1` (buttons 469/494), offset from the pad:

| | at rest | rising | at rest (top) | descending | at rest (bottom) |
| --- | --- | --- | --- | --- | --- |
| **#665 alone** | +1.204 | +1.204 | +1.204 | +1.244 | +1.204 |
| **#666** | +1.204 | +1.164 | +1.204 | +1.244 | +1.204 |

Vertical carry already worked — a rising deck simply pushes the capsule up — so this is a regression guard, not a fix, and both columns are correct. (The ±0.04 while moving is again the one-frame lag at the lift's 2.4 wu/s.)

### The other half of the defect

The leak is the *visible* symptom; the more surprising one is that the built-in transfer fires on contact rather than on support. A kinematic wall sliding along its own length past a player standing on **static ground** tows them 10.6 wu in one second — nothing about that surface is underfoot. That is `player_is_not_dragged_by_a_wall_they_are_only_brushing`, which fails on `ecb99668` at 10.58 wu and passes here at < 0.02.

tommy-xr:
**Correcting a number I supplied for this PR.**

I earlier provided a `#665`-only slip measurement of **−1.951 peak / −1.696 at arrival**, which is cited above as corroboration. A subsequent full playtest measured the same thing far more carefully — **boarding the deck legitimately on foot** (rather than teleporting onto it, as I did) and riding **three** legs on stock `#665`:

| Leg | Travel | Offset at rest | Worst mid-leg | At arrival |
|---|---|---|---|---|
| 1: st.1 → st.2 | −377.53 → −184.43 | +0.005 | −0.734 | −0.710 |
| 2: st.2 → st.3 | −184.43 → −75.83 | +0.004 | −0.229 | −0.029 |
| 3: st.3 → st.2 | −75.83 → −184.43 | +0.056 | +0.495 | +0.296 |

So the real-world slip on `#665` is **~0.73 wu worst case, not ~2.0**. My larger figure is most likely an artifact of teleporting the player onto the deck rather than walking on, which probably left the capsule seated differently. **Please treat the playtest's numbers as authoritative and discount mine.**

Two further observations from that session, both of which still favour merging this PR but are worth recording accurately:

- **The slip does not accumulate across legs.** Each leg resettles to ≈0 when the car stops, and the sign flips with travel direction (consistent with acceleration). So on the tram specifically, stock `#665` was not going to slide a rider off — a rider is never close to the 3.18 wu deck margin.
- **Grav `Lift 1` (431) carries perfectly on `#665` alone** — offset held at exactly +1.200 up and down, zero slip.

None of this weakens the **wall-tow** finding (a kinematic wall sliding past a player on static ground towing them 10.6 wu/s), which remains the strongest argument here: it is a correctness problem independent of the tram, and doors in this engine are kinematic sliders.

PR #675, opened 2026-07-28
fix(vr): crouch from tracked head height

### LINKED ISSUE #512: VR: crouch does not shrink the collider (flat-only in #502)
#502 added a real crouch (collider shrink + headroom-gated stand-up) for the flat runtimes: desktop `LeftControl` and the debug runtime's `crouch` input channel both feed `InputContext.crouch`.

The Oculus runtime never sets `InputContext.crouch`, so in VR the capsule stays 4.8 ft tall regardless of physical crouching — the HMD pose only moves the camera. Authored crouch-only routes (e.g. the MedSci air shaft behind keypad 45100, ~3.1 ft choke) therefore remain impassable in VR.

Suggested approach: derive the crouch request from tracked head height with hysteresis (e.g. crouch below ~70% of calibrated standing eye height, stand above ~85%), feeding the same `InputContext.crouch` so the physics path is shared. Needs headset-on validation, and interacts with the VR scale/anchor correction (slice 7 of the character-controller roadmap).

Found during cross-engine review of #502.

tommy-xr:
Fixes #512

## What this adds

VR crouch, from two sources, on top of the feet-anchored tracked-pose pipeline from #848:

- **Physical crouch** (`shock2vr/src/vr_crouch.rs`): a stateful detector turns floor-relative tracked head height into the shared `InputContext.crouch` request. It calibrates standing height as the running max of valid samples, enters crouch below 70% and exits above 85% (hysteresis dead band against tracking noise), retains the last request through tracking loss, and recalibrates on session restart and reference-space changes (deferred to the exact `changeTime`). The existing feet-planted collider resize and headroom-gated stand-up path are reused unchanged.
- **Button crouch**: left-thumbstick click toggles a latched crouch request (mirroring jump on the right stick), ORed with the detector. The latch resets with the session and on reference-space changes, and its edge detection is gated on action liveness so a click spanning a focus loss cannot mint a spurious toggle.
- **Eye containment**: the rendered eye is capped just below the current capsule's crown (`player_eye_cap_above_center`). A tracked head is a real body's, not the capsule's — a physically crouched adult's eye sits well above the 2.8 ft crouched capsule, and uncapped the player sees over and through the very geometry the capsule clears. Crouched, the cap is derived directly from the flat runtime's `PLAYER_CROUCH_EYE_HEIGHT`, so VR and flat can never disagree about the crouched eye line; standing, it binds only for improbably tall players, so tracking stays 1:1.

## Review

Cross-engine adversarial review (two independent engines + a dedicated duplication pass) ran on the branch; all agreed findings are addressed in the final commit. Notably, an earlier revision rigidly dropped the tracked poses during a latched-but-standing crouch — both engines flagged that it could place the eye and hands below the floor and popped the view a frame-size step at the hysteresis thresholds; it was removed in favor of the cap, which achieves the same containment continuously. One accepted transient remains: when the detector fires mid-descent, the cap engages with a sub-second catch-up while the head continues down — on-device this reads as normal crouch feel.

## Verification

- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` — clean
- `cargo test -p shock2vr` — 552 passed (includes the detector unit suite)
- `cargo apk build --release` — zero warnings (validates against openxr 0.21.1, post-#718)
- **Quest 3 on-device, iterated across three sessions**: physical crouch traverses the previously-impassable MedSci crouch-only duct; the crouched eye stays inside the duct (no seeing over/out of the world); standing eye height verified at the retail line against an NPC; steady 90 Hz, zero skipped frames throughout.

## Notes

- Debugging the eye-height reports along the way surfaced two adjacent issues, handled separately: launch-time ANRs from the unserviced Android event queue (#911, follow-up #910) and STAGE floor-origin instability when Guardian is disturbed by unattended launches (device automation now restores Guardian immediately after launch; a future hardening option is self-calibrating the floor from the detector's standing height).
- The detector's running-max calibration can be inflated by lifting the headset overhead while donning; thresholds recover on the next session or reference-space change. A decay/percentile calibration is a possible follow-up if it bothers in practice.

### DIFF
--- a/runtimes/oculus_runtime/src/lib.rs
+++ b/runtimes/oculus_runtime/src/lib.rs
@@ -1116,7 +1116,7 @@ fn render_swapchain(
     let width = swapchain.width;
     let height = swapchain.height;
 
-    let head_offset = stage_to_pawn(
+    let mut head_offset = stage_to_pawn(
         cgmath::Vector3::new(
             view.pose.position.x,
             view.pose.position.y,
@@ -1124,6 +1124,15 @@ fn render_swapchain(
         ),
         game.player_center_above_floor() + tracked_pose_drop,
     );
+    // The tracked eye belongs to a real body, not to the game capsule, so it
+    // must be held inside the collider crown: a physically crouched adult's
+    // eye sits well above the short crouched capsule, and uncapped the player
+    // sees over and through the very geometry the capsule clears (looking out
+    // of the world from inside a duct). Only the view is capped - hand poses
+    // have no such clipping concern and clamping them would break reaching up.
+    // The cap binds essentially only while crouched (or button-latched);
+    // standing it sits above any realistic head, so tracking stays 1:1.
+    head_offset.y = head_offset.y.min(game.player_eye_cap_above_center());
     let head_rotation = cgmath::Quaternion::new(
         view.pose.orientation.w,
         view.pose.orientation.x,
--- a/runtimes/oculus_runtime/src/lib.rs
+++ b/runtimes/oculus_runtime/src/lib.rs
@@ -11,6 +11,7 @@ use shock2vr::Game;
 use shock2vr::GameOptions;
 use shock2vr::input_context::InputContext;
 use shock2vr::paths;
+use shock2vr::vr_crouch::VrCrouchDetector;
 use std::collections::HashSet;
 use std::time::{Duration, Instant};
 
@@ -460,6 +461,8 @@ fn main() {
     // No discrete actions are mapped for VR controllers yet; this stays empty
     // until an OculusInputMapper is added.
     let mut action_state = shock2vr::input::InputActionState::new();
+    let mut vr_crouch = VrCrouchDetector::default();
+    let mut pending_stage_change_time = None;
 
     let _camera_pos = vec3(0.0, 5.0, 10.0);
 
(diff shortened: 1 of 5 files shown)


tommy-xr:
❌ **Triage: not solid as filed — VR camera is lowered twice** (cross-PR review sweep, 2026-08-02)

The detector itself is clean (4 negative-first unit tests: hysteresis, late calibration, invalid-tracking latch, reset), but the integration has a concrete unshipped bug:

- Crouching shifts the player capsule center down by `(6.0−2.8)/2/2.5 = 0.64` world units (`physics/mod.rs:3147-3155`). `mission_core.rs:5214` returns that capsule center as `camera_pos`, and `oculus_runtime/src/lib.rs:1029` adds the *real tracked* head pose on top. So a physical crouch lowers the VR view **twice** — once because the player's head actually went down, and again by 0.64 wu of capsule shift. Flat runtimes compensate via `player_eye_height_for`; VR doesn't. The camera anchor needs decoupling from the crouch capsule shift (or explicit compensation in the VR render path) before this ships.
- Zero integration coverage — the unit tests stop at the detector, and the author flags the missing device validation. This is exactly what the vr-device-loop exists for.
- Calibration is "tallest sample ever seen this session": one spurious tall sample permanently raises the standing reference until a STAGE change/restart. The 0.70/0.85 thresholds and 3.0 m clamp are unmeasured on hardware.

PR #695, opened 2026-07-29
fix(physics): remove invented marker colliders

### LINKED ISSUE #564: Ectoplasm apparition is solid and partially blocks a doorway (eng1)
The `Ectoplasm` apparition (template 181) in eng1 at `(21.19, -15.22, -44.78)` has a solid collider and partially blocks a doorway.

Confirmed during review: physics body 214 is kinematic with `is_sensor: false` in collision group `selectable`. It also never animates.

### Expected

A scripted ghost/apparition is set dressing — it should not be solid to the player, and should not obstruct a doorway.

Found by the automated play-through loop (eng1 iteration 0).

tommy-xr:
## Reproduction

On current main, `eng1.mis` mission object 181 (`Ectoplasm`) at `(21.1938, -15.2194, -44.7814)` instantiated as one kinematic, non-sensor body in the `selectable` group. Static data shows that it is a `NoRender` script marker with no model and no authored physics; it only relays its `AIWatchObj` action to `Radapp`.

A negative unit test for that data shape failed before the fix because `create_physics_representation` returned a body.

## Fix

Treat inherited `FrobInfo` as interaction behavior, not authored collision geometry. Frobbable entities with neither a model reference nor `PhysType`/`PhysDimensions` now get no invented default collider. Rendered objects, bitmap-backed targets, and explicitly physical invisible objects retain the existing fallback behavior.

Added a durable SDK scenario that discovers Ectoplasm by mission object/template id 181 and verifies it owns no physics bodies.

## Verification

- `cargo test -p shock2vr model_less_frobbable_without_authored_physics_gets_no_collider -- --nocapture` — pass (negative-first: failed before fix)
- `SHOCK2_E2E=1 node --test dist/test/eng1-ectoplasm.e2e.test.js` — pass
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` — pass
- `cargo fmt --all -- --check` — pass
- `npm test` — pass (25 passed, 148 opt-in skipped)
- Full `npm run test:e2e` — 171/173 passed, including the new Ectoplasm scenario and all 23 mission-load tests. The two failures reproduce identically on clean unmodified `origin/main`: Earth Cryokinesis does not damage the Training Droid, and `world-aim-e2e` cannot reload its newly-created save as compatible.

## Visual evidence

No rendered appearance changes: Ectoplasm is explicitly `NoRender`; this removes only its invisible collision body, so screenshot/GIF evidence would be identical.

Fixes #564

### DIFF
--- a/shock2vr/src/mission/entity_creator.rs
+++ b/shock2vr/src/mission/entity_creator.rs
@@ -750,8 +750,9 @@ pub fn create_physics_representation(
     let (
         v_pos,
         v_phys_attr,
-        _v_phys_type,
-        _v_phys_dimensions,
+        v_phys_type,
+        v_phys_dimensions,
+        v_model_name,
         v_frob_info,
         v_hud_select,
         v_creature,
@@ -763,6 +764,7 @@ pub fn create_physics_representation(
             View<PropPhysAttr>,
             View<PropPhysType>,
             View<PropPhysDimensions>,
+            View<PropModelName>,
             View<PropFrobInfo>,
             View<PropHUDSelect>,
             View<PropCreature>,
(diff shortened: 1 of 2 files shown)


tommy-xr:

The guard is right for the eng1 marker, and the negative-first unit test + template-id e2e are proportionate. Two items:
- **Rebase** — the conflict is test-module-text only (adjacent to #705's test), so it's mechanical.
- **Bound the blast radius**: any entity with `PropFrobInfo` but no model and no `PhysType`/`PhysDimensions` now gets no collider — which also means no hover label and no frob raycast. The only coverage is the mission-load smoke suite (loads ≠ still frobbable). One `dark_query` sweep counting how many entities across the 23 missions match that shape would turn "probably fine" into evidence.

Sequencing: open PR #767 touches the same collider fallback in the opposite direction — land the two deliberately in sequence, not in parallel.

tommy-xr:
#767 displaced this one

PR #742, opened 2026-07-30
fix(hud): localize object hover labels

### LINKED ISSUE #739: feat(hud): resolve hover labels through OBJNAME.STR
## Classification

- Domain: **[visual]**
- Kind: **feature gap** (the HUD reads the real property but never implements the Dark Engine string-resource resolution step)
- Severity: Low / non-blocking
- Campaign: **engineering · none · legacy · seed 1378752978**

## Summary

Player-facing world-hover labels expose the serialized `PropObjName` resource-reference syntax and append a fabricated unknown condition. Ordinary authored objects render strings such as:

- `Elevator_button: "A two-state button." | ?`
- `AudioLog: "An audio log: %s." | ?`
- `HumanCorpses: "A corpse." | ?`

The property is a `key: "fallback"` reference into the shipped `OBJNAME.STR` table, not finished UI text. The `%s`/`%d` tokens are likewise formatting placeholders. The current HUD prints the property verbatim and uses `hitpoints.unwrap_or("?")`, so every object without `PropHitPoints` gains a meaningless `| ?` suffix.

## Fresh/live reproduction (legacy assets)

1. Launch a fresh production mission on current `main` with the legacy data root: `DARK_ASSET_PATH=<legacy-root> cargo dbgr --mission eng1.mis --port 8080`.
2. Approach or teleport near the authored Sanger audio log (mission object 451 at approximately `[0.782, -2.369, -166.920]`) and aim at the physical disc so the normal selection brackets/hover label appear.
3. Step and capture the rendered frame through `/v1/step` then `/v1/screenshot`.
4. Observe `AudioLog: "An audio log: %s." | ?` rather than resolved player-facing text.
5. The same issue reproduces without campaign state on any authored `Button #1`/elevator control (template `-201`) and on the ENG Female Corpse in `eng2.mis` (mission object 233 at `[-33.313, -12.318, -282.729]`).

`cargo dq` confirms that these entities inherit `PropObjName("AudioLog: \"An audio log: %s.\"")`, `PropObjName("Elevator_button: \"A two-state button.\"")`, and `PropObjName("HumanCorpses: \"A corpse.\"")`. The legacy `OBJNAME.STR` contains the corresponding case-insensitive keys and localized English values.

## Expected

- Resolve `PropObjName` through the game's mounted/localized `OBJNAME.STR` resource flow (including its embedded fallback semantics).
- Render friendly text rather than the `key: "fallback"` token.
- Apply supported object-specific placeholder values (`%d` stack counts and `%s` instance text) without exposing unresolved formatting syntax.
- Show hit points/condition only when the entity actually has meaningful condition data; do not synthesize `| ?`.
- Preserve debug IDs when `debug_show_ids` is enabled.

## Actual

`shock2vr/src/hud/item_outline.rs::draw_item_name` formats `PropObjName.0` directly, only replaces `%d` for stack counts, and always appends `hitpoints.unwrap_or("?")`. `Game::new` loads `objname.str` into `AssetCache` but the result is discarded and the HUD never consults it.

This should use the real asset/localization path. A hardcoded object-name table or debug-only display shim would not close the feature gap.

## Live campaign evidence

The defect was repeated during ordinary in-world play on unrelated objects, not a one-off debug artifact.

Core Access button

Sanger audio log

Cargo lift control

ENG corpse

Evidence gist: https://gist.github.com/tommy-xr/4d6045dd0d8f4696b2b8f42e14c5cf1f

tommy-xr:
Fixes #739

campaign: engineering · none · legacy · seed 1378752978

## Root cause

The HUD loaded `OBJNAME.STR` but discarded it, then rendered serialized `PropObjName` values such as `AudioLog: "An audio log: %s"` directly. `P$NameType` was also unparsed, so the original stack/log-title/weapon substitution flow could not run. The suffix unconditionally emitted `| ?` when hit points were absent.

## Design

- Parse the authored `P$NameType` property, preserving unknown raw values.
- Resolve `PropObjName` through the real `OBJNAME.STR` resource with its embedded fallback.
- Apply the original NameType substitutions through authored `levelNN.str` log names, stack counts, and `weapon.str` gun-condition strings.
- Show `| HP` only when an object actually has a hit-point property; preserve debug entity IDs without inventing an unknown condition.
- Add a real `eng1.mis` SDK scenario that discovers the Sanger log by stable template, selects it through the production reticle, captures the rendered label, and proves hovering does not collect it.

This follows the original Looking Glass resource flow rather than introducing a debug-only name table or presentation shim:
- [GameStrings parsing](https://github.com/dima424658/darkengine/blob/4aa92d74e727a503954eb01914f69b3997324fc3/src/framewrk/gamestr.cpp)
- [object-name substitutions](https://github.com/dima424658/darkengine/blob/4aa92d74e727a503954eb01914f69b3997324fc3/src/shock/shkprop.cpp)
- [weapon condition buckets](https://github.com/dima424658/darkengine/blob/4aa92d74e727a503954eb01914f69b3997324fc3/src/shock/shkgun.cpp)

## Verification

- Negative-first HUD tests failed on missing localization/formatting helpers before implementation.
- `cargo test -p dark object_name_type_reads_retail_variants_and_preserves_unknown_values`
- `cargo test -p shock2vr hud::item_outline::tests` (12 passed)
- `cargo test -p shock2vr` (390 passed)
- `RUSTFLAGS="-D warnings -C debuginfo=0" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`
- `npm test` (26 passed, 158 skipped)
- focused authored `eng1.mis` hover-label e2e passed
- full SDK e2e on the final commit: 184 tests — 181 passed, 0 failed, 3 expected skipped

## Visual evidence

Localized Sanger hover label loop

<sub>The authored Sanger audio-log disc falls under mission physics while the production reticle tracks it; the localized title stays resolved and the unknown `| ?` suffix is gone. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

### Final still

Localized Sanger hover label

### Before → after

Serialized hover property before and localized log title after

### DIFF
--- a/dark/src/properties/mod.rs
+++ b/dark/src/properties/mod.rs
@@ -369,6 +369,52 @@ pub struct PropObjIcon(pub String);
 #[derive(Debug, Component, Clone, Serialize, Deserialize)]
 pub struct PropObjName(pub String);
 
+/// How the original Shock UI substitutes placeholders in an object's localized
+/// long/short name (`P$NameType` / `ObjNameType`).
+#[derive(Clone, Copy, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub enum ObjectNameType {
+    #[default]
+    Normal,
+    StackCount,
+    LogTitle,
+    Weapon,
+    /// Preserve unfamiliar retail/mod values rather than silently treating
+    /// them as one of the known formatting strategies.
+    Unknown(i32),
+}
+
+impl ObjectNameType {
+    pub fn from_raw(value: i32) -> Self {
+        match value {
+            0 => Self::Normal,
+            1 => Self::StackCount,
+            2 => Self::LogTitle,
+            3 => Self::Weapon,
+            value => Self::Unknown(value),
+        }
+    }
+
+    pub fn raw(self) -> i32 {
+        match self {
+            Self::Normal => 0,
+            Self::StackCount => 1,
+            Self::LogTitle => 2,
+            Self::Weapon => 3,
+            Self::Unknown(value) => value,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Default, Deserialize, Eq, PartialEq, Serialize, Component)]
+pub struct PropObjectNameType(pub ObjectNameType);
+
+impl PropObjectNameType {
+    pub fn read<T: io::Read + io::Seek>(reader: &mut T, len: u32) -> Self {
+        assert_eq!(len, 4, "P$NameType must contain one signed 32-bit value");
+        Self(ObjectNameType::from_raw(read_i32(reader)))
+    }
+}
+
 #[derive(Debug, Component, Clone, Serialize, Deserialize)]
 pub struct PropObjShortName(pub String);
 
@@ -1626,6 +1672,12 @@ pub fn get<R: io::Read + io::Seek + 'static>() -> (
             accumulator::latest,
         ),
         define_prop(
+            "P$NameType",
+            PropObjectNameType::read,
+            identity,
+            accumulator::latest,
+        ),
+        define_prop(
             "P$ObjState",
             PropObjState::read,
             identity,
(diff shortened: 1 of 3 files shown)


tommy-xr:

Good direction and 12 real unit tests, but one concrete correctness bug and two gaps:

- 🔴 `item_outline.rs:109-113` computes weapon condition as `(condition as i32)/10`, assuming 0..100 — but this repo stores `PropGunState.condition` as **0..1** (`dark/src/properties/prop_gun_state.rs:24`, `mission/reload.rs:203`, `scripts/energy_weapon.rs:80`). A pristine gun labels **"Terrible: 1"**. The unit test tests the helper's own wrong assumption.
- `item_outline.rs:35`/`165` can return `""` → no label at all where the raw string used to appear; and `%d` now only substitutes when `NameType == StackCount`, so a stackable lacking the prop renders a literal `%d`.
- The e2e is a render smoke only (asserts resolution/byte-size, `hover-label-localization.e2e.test.ts:485-489`) — it passes pre-fix. Assert the actual label text via the debug API if possible.

Note: `objname.str` load unwraps in a per-frame path (`:163`). Also FYI this adds text via a third UI stack (`SceneObject::screen_space_text`) — doesn't deepen the gui/UI2DRenderer fork, but worth knowing it exists.

PR #743, opened 2026-07-30
fix(gameplay): loot a keycard through its scripted frob path

### LINKED ISSUE #583: Access-card pickup paths do not preserve both key access and the inventory item
Frobbing an access card removes it from the world without ever placing it in the player's inventory. The player's *access rights* are still granted, so doors continue to open — but the card itself silently ceases to exist.

### Cause

`shock2vr/src/scripts/internal_keycard_script.rs:38-41` — on `Frob`, `KeyCardScript` unconditionally emits `AcquireKeyCard` **plus** `DestroyEntity`, regardless of the object's `PropFrobInfo`. Template −157 (`ID Cards`) carries `world_action: MOVE`, i.e. "move it into the player's inventory".

This is the same class of bug as #560 (fixed for `FrobQB` in PR #578) but in a different script, so PR #578 does not address it.

### Repro (eng2)

1. `cargo dbgr --mission eng2.mis --port 8161`
2. The Cargo Bay 2A/2B access card is template **520** at `(-34.03, 2.80, -279.91)`, top tier of Cargo Bay 1A.
3. Frob it: `POST /v1/entities/<id>/message {"type":"Frob"}`
4. Observed: `/v1/player/inventory` stays empty, `GET /v1/entities/<id>` returns null, and no template-520 entity remains in the world.

### Severity: NOT a progress blocker

Verified explicitly, because an automated playtest first reported this as a hard blocker. With the card destroyed and inventory empty, frobbing the Cargo Bay 2 card slot (tmpl 1585 at `(18.66, -10.97, -251.56)`) **still opens the doors** — leaf 636 moved z −253.10 → −251.32 and leaf 705 −255.70 → −257.48. `AcquireKeyCard` puts the region into the player's keyring and `TweqLockedButton`'s lock check honors that, so the chain proceeds without the physical card.

What is actually lost is fidelity and legibility: in System Shock 2 an access card is a visible inventory item. Here the player gets no feedback that a pickup happened, cannot see which cards they hold, and the object is unrecoverable. It also makes automated play misjudge its own state — the playtest concluded it was blocked and reached for `/v1/player/give` as a workaround it did not actually need.

### Expected

Honor `PropFrobInfo.world_action`: a `MOVE` keycard should be acquired **and** moved into the player's inventory (the fix in PR #578 for `FrobQB` is the shape to follow — reuse the shared `can_grab_item` eligibility rule and the `DropEntityInfo` transfer). Cards whose frob info does not request `MOVE` keep the consume behavior.

Found by the automated play-through loop (Engineering iteration 1); severity corrected during manager verification.

tommy-xr:
Fixes #583.

## The bug

Taking a key source out of a corpse or container went through the Container MFD's generic **Take** path, which just re-parents the object into the backpack. The object's `Frob` never ran, so `internal_keycard` never registered the card on the keyring — the Rec2 crew card (obj 996, inside corpse 419) gave no Recreation access — and the card's authored `FrobQB` quest bit (`crewcahd`, worth 20 cyber modules) never fired either.

## The fix

1. **Container Take dispatches Frob for scripted objects.** `ContainerGui`'s `Take` now sends `MessagePayload::Frob` when `uses_scripted_world_frob` holds — an authored `SCRIPT` world action *or* a derived `internal_keycard` (any `PropKeySrc`). The object's own script then decides what happens to it, the same as frobbing it in the world.

2. **A key source keeps its engine fate: register and vanish.** `internal_keycard` grants `AcquireKeyCard` and destroys the object. Access cards are logical access, not backpack inventory — this is unchanged from `main` (an earlier revision of this PR made cards backpack-retained; that is reverted).

3. **One Frob, one owner of the card's fate.** Registering-and-consuming collides with the generic pickup transfer on objects that are both: the MedSci2 R&D card (obj 772) is `MOVE | SCRIPT` with `FrobQB` *and* a derived `internal_keycard`, so a single Frob emitted a backpack `DropEntityInfo` **and** a `DestroyEntity` for the same card. `FrobQB` and `internal_frob_move` now skip the transfer for key sources — `FrobQB` still awards its quest bit — leaving the physical fate to `internal_keycard`. The check is one shared helper, `script_util::is_key_source`, which `uses_scripted_world_frob` also now reuses.

## Verification

- `cargo fmt`; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` — clean.
- `cargo test -p shock2vr` — 721 passed. New/updated unit tests cover: container Take dispatching Frob for a grabbable keycard; `internal_keycard` registering *and* consuming both `MOVE` and use-only key sources; `FrobQB` awarding `rdgrab` while emitting neither a transfer nor a destroy for a key source; `internal_frob_move` returning `NoEffect` for a key source.
- SDK `npm test` (unit + typecheck) — pass.
- e2e (`SHOCK2_E2E=1`): `container-loot`, `hydro2-keycard-gates`, `command-locked-card-relay` — pass.
- `tools/shock2-sdk/test/rec2-crew-keycard.e2e.test.ts` (new): resumes the Rec2 campaign frontier, aims and squeezes at corpse 419, clicks the card button in the real loot MFD, then asserts the card object is gone (registered, not stashed), `crewcahd` is set, and exactly 20 cyber modules are paid once; a second process reloads the save, transitions to `rec1.mis`, frobs the crew card slot and asserts the linked crew doors open with no `hackfail`. It self-skips without the copyrighted campaign frontier save, which is not present in this environment.

No visual change (script/effect logic only).

### DIFF
--- a/shock2vr/src/scripts/frob_qb.rs
+++ b/shock2vr/src/scripts/frob_qb.rs
@@ -2,7 +2,7 @@ use shipyard::{EntityId, UniqueView, World};
 
 use crate::{mission::PlayerInfo, physics::PhysicsWorld};
 
-use super::{Effect, MessagePayload, Script, script_util::set_quest_bit_effect};
+use super::{Effect, MessagePayload, Script, script_util, script_util::set_quest_bit_effect};
 
 pub struct FrobQB {}
 impl FrobQB {
@@ -40,7 +40,16 @@ impl Script for FrobQB {
                     // and leaving the object frobbable would let a second frob
                     // re-fire the other scripts attached to it - `BaseButton`
                     // resends every SwitchLink - after the bit is already set.
-                    if crate::virtual_hand::can_grab_item(world, entity_id) {
+                    //
+                    // A key source (the MedSci2 R&D card, obj 772: `MOVE |
+                    // SCRIPT` + `FrobQB` + a derived `internal_keycard`) is the
+                    // exception: its sibling keycard script registers it on the
+                    // keyring and consumes it, so awarding the bit here is all
+                    // this script does. Transferring it too would emit both a
+                    // backpack transfer and a destroy for one Frob.
+                    if crate::virtual_hand::can_grab_item(world, entity_id)
+                        && !script_util::is_key_source(world, entity_id)
+                    {
                         match world.borrow::<UniqueView<PlayerInfo>>() {
                             Ok(player) => Effect::combine(vec![
                                 quest_bit_effect,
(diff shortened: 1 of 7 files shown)


tommy-xr:
⚠️ **Triage: needs a real rebase + a scope decision** (cross-PR review sweep, 2026-08-02)

The core fix (keep MOVE-flagged cards in the backpack while granting access) is right and the eng2 corpse-233 → card-520 → slot-1585 e2e is genuine. But:

- 🔴 **A third of this is already merged**: #688 (2026-07-29) landed `uses_scripted_world_frob` at `virtual_hand.rs:518` and the identical dispatch in `flat_player_controller.rs:122-129` plus an equivalent unit test. The whole `flat_player_controller.rs` hunk (+46/-2) and the duplicate test should be dropped on rebase.
- 🔴 **Unrequested widening of a shared path**: the helper adds `FrobFlag::SCRIPT` on top of the `PropKeySrc` check, so *every* `MOVE|SCRIPT` grabbable now dispatches `Frob` — newly including container `Take` (`gui/container.rs:327`). Those items deliberately lack `internal_frob_move` (`mission/entity_creator.rs:332`), so any archetype whose authored Frob handler doesn't transfer the item becomes **unlootable**. `frob_qb` is safe; the rest are unaudited — either narrow back to `PropKeySrc` or audit the archetypes.

Minor: `internal_keycard_script.rs:38-52` lacks the carried-item guard its sibling `internal_frob_move.rs:52` has. Sequencing: rebase after #688 (done) and coordinate with #760/#772 (same keycard/`drop_entity_into_container` subsystem).

PR #745, opened 2026-07-30
fix(physics): bound ordinary jump mantle height

tommy-xr:
## Summary

- cap ordinary-jump elevated mantle landings at the jump's ballistic apex
- require a clear head-to-raised path before any elevated mantle may use the parentless-terrain lip exception
- cover both authentic Engineering ceiling escapes: Cargo 2B in `eng2.mis` and OG-Pipe in the `eng1.mis` maintenance maze

Fixes #744.

## Stack

This PR targets `codex/fix-708-jump` / PR #714 at final commit `cd19e396`; ordinary jumping does not exist on `main` yet. The dependency chain is:

```text
main
  └─ PR #712 — original player collision footprint
       └─ PR #714 — ordinary jump
            └─ this PR — elevated-mantle ceiling containment
```

It does not include unrelated play-through fixes.

## Gameplay invariant

The jump mantle's temporary compressed body exists only to model Dark's discontinuous sphere stack crossing a local terrain lip. It must not create extra vertical reach or move the player upward through world geometry.

Two general checks now govern elevated landings:

1. The landing rise cannot exceed the ordinary jump's ballistic apex:

   ```text
   max_rise = jump_speed² / (2 × jump_gravity)
            = 28² / (2 × 40)
            = 9.8 SS2 ft
   ```

2. The initial head-to-raised route must be clear against all world geometry. The parentless-terrain exception remains available only for the subsequent horizontal compressed-body lip crossing.

This preserves intentional stacked routes and the SHODAN same-height barrier descent while rejecting ceilings selected as elevated floors. No mission-specific coordinate, object ID, or threshold is added.

## Campaign reproductions

Both cases use campaign `engineering · none · legacy · seed 1378752978` and one production `Jump` rising edge while holding forward.

### Cargo 2B exterior roof

- Mission: `eng2.mis`
- Stable objects: lift `669`; path nodes `477 → 478 → 479`; buttons `480`, `481`, `620`
- Exact launch: `(43.9006, 3.1440, -168.2003)`, aimed west toward `(39, ~3.2, -170)`
- World-only bidirectional rays verify the traversed slab at `y=6.8`

Negative on `cd19e396`:

```text
(43.900566, 3.144000, -168.200300)
  -> (34.538307, 8.043818, -171.638290)
  -> unchanged for 120 more fixed-timestep frames
```

Green with this PR:

```text
normal north exit -> (43.642235, 3.243999, -164.120060)
jump start        -> (43.900566, 3.144000, -168.200300)
[...]

PR #747, opened 2026-07-30
fix(gameplay): preserve elevator station after load

### LINKED ISSUE #746: fix(gameplay): preserve elevator station across save/load
## Summary

A save/load made at Cargo Bay 2B East lift 669's authored top station restores the platform and player visually at the top, but resets `BaseElevator.current_index` to path index 0. The next ordinary top-button press therefore sends the lift to the middle station instead of cycling to the bottom. Repeated presses alternate middle/top, making the restored elevator cycle observably different from uninterrupted play.

This surfaced during a new detailed play-through after collecting Sanger's Engineering Control code. The generic elevator save/load defect is confirmed, but it is **not established as the sole cause** of the campaign's still-unresolved return-route finding. Retail walkthroughs say to head below and backtrack; independent product-input testing confirms that an uninterrupted top press reaches the bottom and ordinary movement can leave the bottom car north to about `z=-158`, but it has not established a path from that bottom room to the Engineering bulkhead. The campaign-confirmed middle-stop north/west route does reach the raised central bridge.

Campaign: `engineering · none · legacy · seed 1378752978`
Dependency stack: PR #712 → PR #714 → PR #745. Ordinary jump and Cargo 2B containment are required to reach this state on the campaign branch.

## Stable reproduction

Mission: `eng2.mis`

Stable mission objects:

- Cargo Lift 669
- path nodes 477 (bottom y=-12.6) → 478 (middle y=-5.1) → 479 (top y=2.1) → 477
- buttons 480 (bottom), 481 (middle), 620 (top)

1. Drive the real lift 669 from bottom through middle to top.
2. Stand on the top lift at player `(43.900566, 3.144, -168.2003)`.
3. Save, then load that save.
4. Rediscover button 620 by stable mission object id, aim through the production flat camera, and use an ordinary squeeze edge.
5. Step until the lift settles.

Expected: the authored loop advances 479 → 477; lift `(43.9,-12.6,-168.2)`, carried player approximately `(43.90065,-11.556,-168.20041)`.

Actual: after load it advances as if still at index 0, 477 → 478; lift `(43.9,-5.1,-168.2)`, carried player `(43.90083,-4.056,-168.20055)`. A second press returns to top rather than reaching bottom.

Control without save/load: the same production top-button squeeze correctly reaches node 477 and the player can ordinarily walk north off the bottom lift to `(43.90004,-11.156,-163.00003)` and about `(43.9,-11.156,-158)`.

## Cause

`BaseElevator::initialize` reads the loaded platform position, but always leaves `current_index` at 0. It should generically reconcile the initialized platform position with the nearest authored path station (with explicit tolerance/tie behavior), so the next `TurnOn` advances from the station actually restored. No Cargo-specific IDs or coordinates should be needed.

## Evidence

A local private reproduction save exists (not attached because it contains retail-derived state): SHA-256 `88646bd22dd2496d36fe86ede10757e35f0ba9473872070a03126f93ffa077ae`. Fresh mission setup plus the steps above reproduces without distributing it.

Walkthrough cross-checks:

- GameBanshee Engineering B §15 says after collecting Sanger's log it is time to return to Engineering A.
- Port Forward Cargo Bay 2 says to head below, backtrack to the four-way corridor, and take the Engineering Control grav lift.

## Acceptance

- infer the initialized elevator station generically from the loaded platform position
- preserve ordinary uninterrupted cycling
- save/load at each Cargo lift station routes to the authored next station
- production top-button use reaches the bottom, bottom-room movement reaches about `z=-158`, and the middle-stop route reaches the raised bridge
- do not claim this issue alone resolves the separate campaign route until replay proves the four-way-corridor/bulkhead path
- include negative-first unit/E2E coverage and matched headless before/after visual proof

tommy-xr:
## Summary

- reconstruct `BaseElevator`'s current path index from the save-restored moving-terrain position
- use the native 0.1-world-unit station tolerance, nearest-station selection, and a deterministic lower-index tie break
- retain the original index-zero fallback when a platform is between stations instead of inventing a station
- add a durable Cargo 2B scenario covering save/load at all three authored stops, real button interaction, rider carry, and the ordinary middle-stop bridge exit

This PR is stacked on `fix/cargo2-ceiling-escape` / #745.

## Campaign evidence and scope

`Campaign: engineering · none · legacy · seed 1378752978`

The clean campaign exposed the defect after reaching Sanger: moving terrain for Cargo lift mission object `669` restored at its top node, while a fresh `BaseElevator` script silently reset `current_index` to zero. The same production press of top button mission object `620` therefore dispatched:

- uninterrupted control: top `y=2.1` → bottom `y=-12.6`
- save/load regression: top `y=2.1` → middle `y=-5.1`

The confirmed defect is generic elevator save/load state loss. It is not claimed as the sole cause of the campaign's still-unresolved return-route finding. Independent physical testing confirms ordinary movement on the bottom floor north to about `z=-158`, but does not establish a route from that room to the Engineering bulkhead. The campaign-confirmed middle-stop north/west route does reach the raised central bridge.

The retail walkthrough says to head below after Sanger and then backtrack to the four-way corridor:

- [Port Forward Cargo Bay 2 walkthrough](https://portforward.com/games/walkthroughs/System-Shock-2/Cargo-Bay-2.htm)
- [GameBanshee Engineering B walkthrough](https://www.gamebanshee.com/systemshock2/walkthrough/engineeringb.php)

## Negative-first regression

On exact stacked base `125c9ac3`, the durable scenario failed after the top-station reload:

```text
expected lift y=-12.6, got -5.1
```

On this branch it passes through the full authored cycle:

- loaded top → bottom: lift `[43.9,-12.6,-168.2]`, player `y=-11.556`
- ordinary bottom-room movement: `(43.881,-11.156,-158.322)`
- loaded bottom → middle: player carried to `y=-4.056`
- ordinary middle-stop exit: `(34.000,-3.956,-174.600)`
- loaded middle → top: lift `[43.9,2.1,-168.2]`

Concrete runtime entity IDs are rediscovered after every load; only stable mission object IDs identify the lift and buttons.

## Verification

- `git diff --check`
- `cargo test -p shock2vr` — 396 passed / 0 failed
- `cd tools/shock2-sdk && npm test` — 188 total / 26 passed / 0 failed / 162 expected E2E skips
- focused final Cargo lift scenario — 1 passed / 0 failed (`101.1 s`)
- full serial SDK E2E suite on final head — 188 total / 185 passed / 0 failed / 3 expected SHODAN finale skips (`1,976.2 s`)

## Visual verification

Cargo 2B elevator save/load before and after

<sub>Same fresh `eng2.mis` state, top-stop save/load, production aim/button squeeze, and 600 fixed frames. Before, the restored top stop forgets its index and stops at the middle floor; after, it completes the authored top-to-bottom leg.</sub>

### Before → after

Cargo 2B elevator final stations before and after

Fixes #746

### DIFF
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -1224,8 +1224,17 @@ fn plan_jump_mantle(
                     .filter(|(normal_y, _)| *normal_y > CLIMB_TOP_OUT_MIN_GROUND_NORMAL)
                     .map(|(_, floor_y)| floor_y)
                     .filter(|floor_y| {
-                        *floor_y - current_feet_y
-                            > (PLAYER_STEP_HEIGHT + PLAYER_CONTACT_OFFSET) / SCALE_FACTOR
+                        let rise = *floor_y - current_feet_y;
+                        rise > (PLAYER_STEP_HEIGHT + PLAYER_CONTACT_OFFSET) / SCALE_FACTOR
+                            // The compressed body models Dark's discontinuous
+                            // sphere stack crossing a terrain lip; it does not
+                            // add vertical reach. Require the player's feet to
+                            // be able to reach the landing within the ordinary
+                            // ballistic apex. Otherwise an open lift shaft can
+                            // turn the room ceiling ahead into an "elevated
+                            // floor" and script the player onto its exterior
+                            // roof (#744).
+                            && rise <= max_rise
                     });
                 // A same-height floor visible above the lower candidate is the
                 // stacked-terrain signature: the continuous capsule cannot
--- a/shock2vr/src/scripts/base_elevator.rs
+++ b/shock2vr/src/scripts/base_elevator.rs
@@ -13,6 +13,11 @@ use super::{
     script_util::{get_first_link_of_type, get_first_link_with_data},
 };
 
+/// Native elevator reroutes treat a platform within this distance as already
+/// occupying a station. Use the same tolerance when reconstructing script
+/// state from a save-restored moving-terrain position.
+const ELEVATOR_STATION_TOLERANCE: f32 = 0.1;
+
 pub struct BaseElevator {
     path_offset: Vector3<f32>,
     current_index: u32,
@@ -94,6 +99,29 @@ impl BaseElevator {
         self.move_to_target(target_index);
         true
     }
+
+    fn restored_station_index(&self) -> Option<u32> {
+        let tolerance_squared = ELEVATOR_STATION_TOLERANCE * ELEVATOR_STATION_TOLERANCE;
+        let mut nearest: Option<(u32, f32)> = None;
+
+        for (index, (_, station, _)) in self.path.iter().enumerate() {
+            let distance_squared = (self.current_position - station.position).magnitude2();
+            if distance_squared > tolerance_squared {
+                continue;
+            }
+
+            // Strictly closer replaces the candidate. Equal-distance ties
+            // retain the earlier (lower) path index for deterministic routing.
+            if nearest
+                .map(|(_, nearest_distance)| distance_squared < nearest_distance)
+                .unwrap_or(true)
+            {
+                nearest = Some((index as u32, distance_squared));
+            }
+        }
+
+        nearest.map(|(index, _)| index)
+    }
 }
 impl Script for BaseElevator {
     fn initialize(&mut self, entity_id: EntityId, world: &World) -> Effect {
@@ -123,6 +151,9 @@ impl Script for BaseElevator {
 
         // Save the path
         self.path = path;
+        if let Some(restored_index) = self.restored_station_index() {
+            self.current_index = restored_index;
+        }
 
         Effect::NoEffect
     }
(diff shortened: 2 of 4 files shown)


tommy-xr:
❌ **Triage: blocked — retarget to main and use the script_states envelope** (cross-PR review sweep, 2026-08-02)

Real progression bug and the best test quality of its batch (negative-first with quoted failure, tolerance-boundary + tie-break unit cases, a thorough 3-station e2e). Two structural blockers:

- **Stranded base**: `fix/cargo2-ceiling-escape` is 5 commits of unreviewed physics work ahead of main; only one of those has a PR (#745, itself open). Verified escape hatch: `git diff origin/main...origin/fix/cargo2-ceiling-escape -- base_elevator.rs elevator-save.e2e.test.ts` is **empty** — cherry-picking the single commit onto main applies clean, and the e2e uses only SDK calls present on main.
- **Ad-hoc state reconstruction instead of the `script_states` envelope**: `current_index` is textbook script-private state, and main has the mechanism for it (#757/#758, `script_state_key`/`save_state`/`restore_state`, worked example in `picture_swap.rs:105-119`). Position-inference in `initialize()` only fixes the parked case — `is_moving`/`desired_position`/direction are still lost, and a save mid-travel falls back to index 0: the same defect, narrower. The envelope fixes it completely.

Also note: `restored_station_index()` runs unconditionally in `initialize()`, so a *fresh* load of an elevator authored parked at a non-zero station now starts there too — probably more correct, but it's an untested behavior change beyond the title. Nit: the e2e helper named for templates actually passes mission object ids (works — `template_id` field — but confusing).

PR #752, opened 2026-07-30
fix(gameplay): honor authored interaction overlays

### LINKED ISSUE #751: Fluidics console cannot be frobbed through its authored control overlay
## Reproduction

1. Load `eng1.mis` before installing the coolant regulator and stand at approximately `(0.956, -1.556, -175.278)` facing the Fluidics Control console.
2. Discover `Fluidic Button` by stable template ID 1114 and `Fluidic Comp` by stable template ID 1152.
3. Aim the flat production crosshair at the visible Fluidics control panel and press the normal use/squeeze input.

## Actual

The production interaction ray hits the non-frobbable `Fluidic Comp` (1152) first and stops. `FlatPlayerController::update` then filters that first hit through `is_frobbable`, leaving no target and dispatching no `Frob`. The Fluidics panel never reacts.

Measured from a fresh runtime on `origin/main` using the campaign save:

- player: `(0.956, -1.556, -175.278)`
- button: runtime 1281, template 1114, distance 2.347
- comp: runtime 743, template 1152, distance 2.306
- production-equivalent ray first hit: runtime 743 / template 1152 at distance 1.860
- UI before and after squeeze: shooter mode, no active panel

A broad interaction-ray scan also finds no view ray where template 1114 is the first combined ENTITY/SELECTABLE/WORLD hit.

## Expected

The authored selectable/frobbable Fluidics control overlay should receive the normal `Frob`, both for the pre-install attempt and the later coolant purge. Non-frobbable decorative entity geometry may yield to a frobbable selectable at effectively the same surface, but walls and other world occluders must still block interaction.

## Suspected area

`shock2vr/src/flat_player_controller.rs` currently performs one combined first-hit raycast and only afterwards applies `is_frobbable`. `P$PickBias = -2000` on the computer and the distinct `HUDSelect`/SELECTABLE overlay semantics are relevant authored evidence.

tommy-xr:
## Summary

- honor authored negative `P$PickBias` when the flat crosshair first hits the Fluidics computer shell
- let a `HUDSelect(false)` visual shell yield only to its direct, co-located `HUDSelect(true)` interaction overlay, with tight body-origin and surface-gap bounds
- keep the combined recast's ENTITY/WORLD/UI/RAYCAST blockers, so walls, closed doors, and unrelated controls still occlude interaction
- preserve Dark's compatibility rule that any `P$FrobInfo` marks an interactable object: scripted objects such as `StdDoor` can consume `Frob` even with an empty world-action mask
- cover Fluidics and Master Power through the production camera + squeeze path, plus the existing Earth door path

Fixes #751

### Fluidics Control

eng1 Fluidics Control interaction demo

At the same production crosshair pose, the fix exposes the authored Fluidics hover target and the following production squeeze consumes the one-shot offline EmailTrap. [Full-resolution source stills and binary media](https://gist.github.com/tommy-xr/6c164c7de355a45ac43853cc923008b7).

eng1 Fluidics Control before and after

### Master Power

eng1 Master Power interaction demo

From the same Engine Core fixture and camera/squeeze sequence, the base stays `ENG. CORE OFFLINE`; the fix reaches the authored overlay and powers it to `ENG. CORE ONLINE`. [Full-resolution source stills and binary media](https://gist.github.com/tommy-xr/113cd5337096d5e57f88fcc688a8a6a1).

eng1 Master Power before and after

Both captures use matched pre-fix/fixed debug-runtime binaries, fresh or cold campaign fixtures, rediscovered runtime entity IDs, and deterministic `/v1/step` + `/v1/screenshot` sequences.

## Verification

- `RUSTFLAGS="-D warnings" cargo test -p shock2vr` — 386 passed
- `env -u SHOCK2_E2E npm test` in `tools/shock2-sdk` — 26 passed, 159 E2E scenarios skipped as designed
- focused runtime regression set: Earth `StdDoor`, eng1 Fluidics, eng1 Master Power — 3 passed
- full `npm run test:e2e` — 182 passed, 0 failed, 3 intentionally skipped

The Master Power E2E starts from fresh eng1, satisfies only the prior nacelle condition, then uses production `aimAt` + squeeze and verifies the display, `CorePower`, linked notes, one-shot router consumption, and elevator unlock.

### DIFF
--- a/shock2vr/src/flat_player_controller.rs
+++ b/shock2vr/src/flat_player_controller.rs
@@ -14,7 +14,7 @@ use shipyard::{EntityId, Get, View, World};
 
 use dark::{
     SCALE_FACTOR,
-    properties::{PropFrobInfo, PropLimbModel, PropPlayerGun},
+    properties::{PropFrobInfo, PropLimbModel, PropPickBias, PropPlayerGun},
 };
 
 use crate::{
@@ -146,19 +146,13 @@ impl FlatPlayerController {
         // hitbox proxies to their parent, and ignoring the weapon we hold).
         let forward = look.rotate_vector(vec3(0.0, 0.0, -1.0));
         self.last_aim = Some((point3(camera_pos.x, camera_pos.y, camera_pos.z), forward));
-        let highlighted = physics
-            .ray_cast(
-                point3(camera_pos.x, camera_pos.y, camera_pos.z),
-                forward,
-                InternalCollisionGroups::ENTITY
-                    | InternalCollisionGroups::SELECTABLE
-                    | InternalCollisionGroups::WORLD
-                    | InternalCollisionGroups::UI
-                    | InternalCollisionGroups::RAYCAST,
-            )
-            .and_then(|r| r.maybe_entity_id)
-            .map(|e| resolve_proxy_entity(world, e))
-            .filter(|e| Some(*e) != self.wielded_entity && is_frobbable(world, *e));
+        let highlighted = interaction_target(
+            world,
+            physics,
+            point3(camera_pos.x, camera_pos.y, camera_pos.z),
+            forward,
+            self.wielded_entity,
+        );
 
         // Place the viewmodel + fire on the trigger edge.
         if let Some(entity_id) = self.wielded_entity {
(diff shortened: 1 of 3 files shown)


tommy-xr:
Campaign blocker found after both Engineering nacelles: Master Power repeats #751 without negative PickBias, so this PR's current skip condition is too narrow.

Independent evidence: https://github.com/tommy-xr/shock2quest/issues/751#issuecomment-5151293096

Two exact resets and 126 normal-camera ray+squeeze attempts produced zero selectable hits on mission object 1870; co-located non-frobbable display 1103 wins whenever the console is in view. Unlike Fluidic Comp 1152, Master Power Comp 1103 has no `P$PickBias`, so `interaction_target` returns `None` before its second ray.

Please extend the focused fix to recognize an effectively co-surface frobbable selectable behind non-frobbable decorative entity geometry while keeping an unrelated second entity or world/door hit blocking. The live campaign is frozen at this frontier.

PR #753, opened 2026-08-01
fix(gameplay): handle player death and QBR respawn

tommy-xr:
## Summary

- clamp lethal player damage at zero and enter an explicit alive/dead/respawning lifecycle
- use each activated Resurrection Station scanner model and its authored SwitchLink teleport target to reconstruct after five seconds
- atomically charge the retail 10-nanite cost, restore half maximum health, suppress dead-player input, and refuse saves while dead
- expose `player.life_state` through the debug runtime/SDK and cover terminal death plus save-persistent QBR activation end to end
- update the player-death design note to describe the implemented lifecycle and remaining presentation work

## Reproduction

Before this change, the new negative-first runtime scenario sent lethal Damage to the real MedSci player and observed `-100` HP with no lifecycle/loss state. Activating the authored QBR scanner did not change that: life state was absent and reconstruction never occurred.

## Verification

- `cargo fmt --all -- --check`
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`
- `cargo test -p shock2vr` (383 passed)
- `cargo test -p shock2vr player_death_tests --lib` (2 passed, post-rebase)
- SDK `npm test` (26 passed, 159 opt-in skipped)
- focused player-death e2e (2 passed, post-rebase)
- full SDK `npm run test:e2e` (182 passed, 0 failed, 3 explicitly gated finale tests skipped; all 23 missions loaded)

Strict Clippy was also attempted, but the unchanged `engine` crate currently has baseline `-D warnings` failures such as `module_inception` and `new_ret_no_self`.

## Remaining risk

This implements the authoritative gameplay lifecycle and automation-visible loss signal. A dedicated death overlay or VR camera-collapse presentation remains a follow-up; this change adds no rendering path. Difficulty selection is not implemented in the current game, so the retail Normal-or-higher cost of 10 nanites is used.

Fixes #670

PR #755, opened 2026-08-01
fix(gameplay): make healing items restore health

### LINKED ISSUE #668: Medical Kit cannot be used - MedkitScript is an UnimplementedScript
**Domain:** [gameplay] · **Kind:** feature gap

Using a Medical Kit from the inventory does nothing — no healing, and the item is never consumed. The player has no way to heal from a carried medkit.

### Root cause

`shock2vr/src/scripts/mod.rs:655`:

```rust
"medkitscript" => Box::new(UnimplementedScript::new(&script_name)), // cyber modules
```

The `Medical Kit` template carries `PropScripts { scripts: ["MedkitScript"] }`, and its `-49 Goodies` ancestor sets `PropFrobInfo { inventory_action: SCRIPT }` — so the inventory-use path dispatches to a script that is explicitly unimplemented. The use path provably does nothing **regardless of how the click is delivered**.

(The trailing `// cyber modules` comment looks like a copy-paste leftover from the neighbouring arm and is unrelated to medkits.)

### Repro

1. Launch any mission with a Medical Kit in inventory (`POST /v1/player/spawn-item {"template":"Medical Kit"}` is the quickest route).
2. Take damage so HP is below max.
3. `POST /v1/input/action {"action":"ToggleUseMode"}`, then click the Medical Kit in the `/v1/ui` strip.

**Expected:** HP increases and the medkit is consumed.
**Actual:** HP unchanged; the item remains in inventory.

Observed in a `command1` playtest where the player finished the session at **16/30 HP** with an unusable medkit carried the whole way.

### Second symptom to fix alongside

A **single** click on the medkit makes it **vanish from the inventory strip** until use mode is toggled off and on again. The item is still in `/v1/player/inventory` — it's a UI/strip refresh defect, but it makes the item look consumed when it wasn't.

### Not a duplicate

- **#437** is direct weapon *selection* (1-9 keys) and **#612** is the Psi Amp *wield* path — both are wield/selection, not item **use**. Weapons in the same strip wield correctly in the same session, so this is not the known 20-frame double-click discoverability problem.
- **#591** covers medpatch *pickup*, not use.

`campaign: command · marine-standard · 25th · seed 1412905396`

tommy-xr:
Fixes #668

> **Stack dependency:** this PR is based on [#975](https://github.com/tommy-xr/shock2quest/pull/975) (`fix/974-vr-released-item-physics`). The review order is #967 → #968 → #975 → **#755** → #980 → #986. Review only commit `d2f6eae6`; its diff against #975 is the nine-file healing change.

## Campaign evidence

`earth → medsci* → eng* → hydro → ops* · VR · --auto · --restart · seed 976894603`

The exact MedSci2 replay at campaign head `9698555b` physically opened Desk mission object 471, looted its contained Med Patch mission object 1054 through the real world panel, opened the X backpack, retrieved the patch into the left hand, and pulled the held trigger. At `pt028` / `pt031` / `pt032`, HP remained **29/35** and the patch survived. The production trigger delivered the patch's authored `Frob`; the current gamesys mapping still resolved `medpatchscript` to `UnimplementedScript`.

Run evidence: `.claude/skills/play-through/runs/earth-ops-vr-seed-976894603-iter-3-medsci2-replay-02/`.

The finding is a valid high-value gameplay gap but not a campaign blocker: recovery is materially constrained, while evasion and alternate routes remain possible.

## What changed

- map the authored `medpatchscript` and `medkitscript` names to one shared, data-driven healing implementation
- reproduce retail behavior recovered from shipped `allobjs.osm`:
  - Med Patch: 10 HP budget, 2-HP pulses
  - Medical Kit: 200 HP budget, 5-HP pulses
  - first pulse after 0.1 seconds, then every 1.5 seconds
- consume exactly one stack unit only when use starts below maximum HP; preserve the item at full health
- apply Pharmo-Friendly's retail 20% integer bonus
- route every timed pulse through the central `Effect::AdjustHitPoints` handler, preserving death guards and HP tracing
- serialize the active timer course for same-mission save/load and reject corrupt/non-finite saved doses without looping

The active course is intentionally global player state: the source item is destroyed as soon as use succeeds, so no surviving object script can own the timer. Ordinary mission transitions start a fresh course; persistent same-mission saves restore it.

## Negative-first and regression coverage

Before implementation, the durable MedSci2 VR scenario failed at the intended assertion: the exact Desk 471 patch remained alive and HP never changed after the production held-trigger `Frob`.

The final stacked head now verifies:

- real MedSci2 VR Desk panel → physical loot → world drop → trigger storage → X backpack → left squeeze → held trigger
- no debug-only shortcut and no `TriggerPull` substitution for the patch's authored `Frob`
- 2-HP timed pulses, exact 10-HP budget, entity/stack consumption, and mid-course save/load cadence
- flat Medical Kit 5-HP pulses to the authored maximum
- full-health Med Patch preservation across save/load
- Pharmo-Friendly purchase, cross-mission persistence, and exact 12-HP patch budget
- #975's real MedSci2 two-drop/card/save-load path still passes on the healing head

## Validation

- `git diff --check`
- full core unit suite on the exact campaign integration: **694 passed, 0 failed**
- final stacked SDK unit run: **261 total, 26 passed, 235 E2E-skipped, 0 failed**
- final stacked `healing-items.e2e.test.ts`: **3 passed, 0 failed**
- final stacked `medsci2-vr-released-item.e2e.test.ts`: **1 passed, 0 failed**
- exact-campaign healing commit `18524291`; campaign continuation cherry-pick `48a75b97`
- review commit `72a2dd53` on exact #975 head `82f40001`

No Oculus/OpenXR runtime code changed, so no headset install cycle was required. The production VR hand path is covered by the real debug-runtime VR scenarios; Android CI is required to pass before merge.

## Player-observable evidence

VR Med Patch timed healing

<sub>The real MedSci2 Desk 471 Med Patch is looted through the VR backpack and used from the left hand. Exact base `9698555b` leaves HP at 20/30 and keeps the item held; fixed `6c1799bd` consumes it and visibly advances 20 → 22 → 24 → 26 → 28 → 30 on the retail cadence.</sub>

### Before → after at the first pulse

Base vs fixed at first retail pulse

### Full 10-HP course

Full 10 HP course

[Full-resolution media and capture details](https://gist.github.com/tommy-xr/09ae745590c0e1f3fa3169894c380714). The matched runs use the same deterministic `/v1/step` and `/v1/screenshot` sequence; hosted bytes were hash-verified against the inspected local GIF/PNGs.

### DIFF
--- a/shock2vr/src/lib.rs
+++ b/shock2vr/src/lib.rs
@@ -1388,6 +1388,12 @@ impl Game {
             rotation,
             quest_info,
             player_vitals: save_load::capture_player_vitals(self.active_game_scene.world()),
+            active_healing: self
+                .active_game_scene
+                .world()
+                .borrow::<UniqueView<crate::scripts::healing_item::ActiveHealing>>()
+                .map(|active| active.clone())
+                .unwrap_or_default(),
             active_mission: self.active_game_scene.scene_name().to_string(),
             is_crouched,
         };
--- a/shock2vr/src/mission/mission_core.rs
+++ b/shock2vr/src/mission/mission_core.rs
@@ -328,6 +328,13 @@ enum ComestibleUseOutcome {
     Consumed,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum HealingItemUseOutcome {
+    NotUsed,
+    DecrementedStack,
+    DestroyEntity,
+}
+
 fn update_player_psi_points(world: &World, update: impl FnOnce(i32) -> i32) -> bool {
     let player_entity = world.borrow::<UniqueView<PlayerInfo>>().unwrap().entity_id;
     let mut psi_states = world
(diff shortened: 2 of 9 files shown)


tommy-xr:

High-value fix with strong tests (full-health refusal, stack-vs-destroy, legacy-save serde). Two things before merge:

- **Route `tick_player_healing` through `Effect::AdjustHitPoints`** instead of writing `PropHitPoints` via `ViewMut` (`healing_item.rs:592-598`). As written it skips the documented single HP funnel and its `hp:` trace — and once #753 lands, it skips the death-lifecycle guard, so a running healing course would revive a `Dead` player (the unit test at `healing_item.rs:717` literally heals a player at −2 HP). This is a concrete cross-PR interaction, not a style point.
- **Justify the new `GlobalData.active_healing` save field in the body.** It's a defensible exception to the `script_states` envelope (the source item is destroyed on use, so there's no owning entity), but it should be an explicit, stated exception given the repo just standardized script-private persistence.

Minor: `ActiveHealing::advance` can loop forever on a corrupt save with `pulse == 0 && pulse_interval_secs <= 0` — `queue()` validates, deserialization doesn't.

PR #756, opened 2026-08-01
fix(gameplay): run periodic radius stim sources

### LINKED ISSUE #390: Act/react receptron fidelity: stim-type-aware responses and periodic emitters
#388 implements one-shot radius stim blasts with generic damage. Two fidelity gaps remain, both needing the receptron side of the act/react database (`L$Receptron`/`LD$Receptro` — parseable, 88-byte records, same versioned framing as `LD$arSrcDes`):

1. **Stim-type-aware responses**: EMP / Stasis / Psi / Cryo blasts currently deal generic damage to everything with hit points. The original engine routes stims through per-archetype receptrons (EMP damages robots only, stasis immobilizes, radiation applies damage-over-time...).

2. **Periodic emitters**: radius stim sources without the `explosiontype` class tag (electrical sparks: Electricity 5 @ r2.4; Swarm; Rad Burst — the rad barrel's lingering radiation) fire on a period in the original engine, not once. #388 deliberately excludes them via the class-tag gate. The period/max-firings fields live in the unparsed remainder of the 108-byte `sStimSourceDesc` record.

Also worth noting from the same review (pre-existing, systemic): two same-frame non-lethal `Damage` messages that are jointly lethal both read the same pre-decrement HP in `InternalSimpleHealth`, leaving the entity alive at negative HP — damage application isn't atomic within a frame.

tommy-xr:
## Summary

- parse the periodic lifecycle fields from Dark Engine stimulus source descriptors
- attach a scheduler to inherited non-explosion radius sources and emit their authored stim at birth/cadence, including finite firing counts, intensity slopes, and destroy-on-completion
- route radius stimuli through receptrons without explosion impulse, including the transform-less player entity
- cover the scheduler with unit tests and command2 electrical-sparks behavior with a real-runtime SDK scenario

## Reproduction

Before this change, command2 electrical sparks never changed player HP: entity creation explicitly excluded non-explosion radius sources, and the source lifecycle bytes were discarded. The parser regression test was first run against the old behavior and failed because descriptors with different authored periods parsed identically.

## Validation

- `cargo test -p dark` (70 unit + 9 mission-loading tests)
- `cargo test -p shock2vr` (383 tests)
- `cd tools/shock2-sdk && npm test` (26 passed, 158 e2e-gated)
- targeted `periodic-stim.e2e.test.ts` (passed)
- `cd tools/shock2-sdk && npm run test:e2e` (181 passed, 0 failed, 3 existing gated skips; all 23 missions loaded)

Fixes #390

### DIFF
--- a/dark/src/properties/mod.rs
+++ b/dark/src/properties/mod.rs
@@ -670,13 +670,30 @@ pub enum StimPropagator {
 pub struct StimSourceOptions {
     pub intensity: f32,
     pub propagator: StimPropagator,
+    /// The temporal portion of the source descriptor. Radius sources use this
+    /// to fire once at birth and then at the authored cadence.
+    #[serde(default)]
+    pub lifecycle: StimSourceLifecycle,
+}
+
+/// Dark Engine's standard periodic source lifecycle (`sPeriodicLifeCycle`).
+/// The values occupy the first 16 bytes of the 32-byte lifecycle block in an
+/// `sStimSourceDesc` record.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct StimSourceLifecycle {
+    pub period: Duration,
+    pub max_firings: i32,
+    pub no_max_firings: bool,
+    pub destroy_on_completion: bool,
+    pub intensity_slope: f32,
 }
 
 impl StimSourceOptions {
-    pub fn read(reader: &mut Box<dyn ReadAndSeek>, _len: u32) -> StimSourceOptions {
+    pub fn read(reader: &mut Box<dyn ReadAndSeek>, len: u32) -> StimSourceOptions {
         // sStimSourceDesc (108 bytes): propagator id, intensity, then
-        // propagator-specific params (a redundant propagator name string sits
-        // at +76). Only the fields the game consumes are parsed.
+        // a 32-byte shape block and 32-byte lifecycle block. A redundant
+        // propagator name string sits at +76. Radius shape data starts at +12;
+        // the standard periodic lifecycle is flags/period/max/slope at +44.
         let propagator_id = read_u32(reader);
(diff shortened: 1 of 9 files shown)


tommy-xr:
❌ **Triage: not solid as filed** (cross-PR review sweep, 2026-08-02)

The parser work is careful (per-record `Cursor`, so the `SeekFrom::Start(44)` is safe; negative-first payload test) — but two blockers:

- **Firing state is not persisted.** `InternalPeriodicStim` holds `sources: Vec<PeriodicSourceState>` with no `script_state_key`/`save_state`, and `initialize` resets `next_firing = 0` — so **every source re-fires at birth on every load**. Standing in sparks and quick-loading takes a free damage pulse each time, and a `destroy_on_completion` source resurrects its full firing budget. This is exactly the case the `script_states` envelope (#757) exists for.
- **A fleet-wide gate flip validated against one archetype.** `entity_creator.rs:329-355` gives *every* radius `StimSource` without an `explosiontype` tag an emitter, across all 23 missions, with parse offsets validated only against command2's `-2216` sparks. The failure mode is severe: a period misparsed as 0 with `no_max_firings` set means `MINIMUM_PERIOD = 1ms` → ~16 firings per 60 Hz frame → instant death near the source. At minimum validate the other command2 archetypes (`-1527`, `-2409`, `-3751`) and make the degenerate-period case fail safe (skip + warn, not machine-gun).

Also unstated in the body: `radius_blast` now delegates to `radius_stim`, which adds the player to the in-range set — the player becomes newly damageable by **explosions**, not just periodic sources. Probably a fix, but it's a player-facing change that needs stating (and it compounds with #753: misparses become fatal, not noisy). e2e nit: pinning the player with thumbstick `0.5` to cancel gravity is a magic constant tied to current tuning.

PR #760, opened 2026-08-01
fix(gameplay): grant Bridge Card access

### LINKED ISSUE #687: Bridge Card pickup does not grant key access
## Summary

The Command B Escape Pod 4 Bridge Card can be picked up and persists physically in inventory, but its key region is never acquired. The required bridge card slot then plays `hackfail` and leaves the bridge door closed, blocking campaign progression.

## Reproduction

1. Load a state before collecting Command B mission object 135 (Bridge Card, template -2594, `P$KeySrc` region 8192).
2. Reach Escape Pod 4 through normal play and pick up the visible card with the flat use control.
3. Return to card slot mission object 138 (`P$KeyDst` region 8192) and frob it.
4. Observe `hackfail`; security door 144 stays closed.

## Evidence

The saved held item is the Bridge Card and retains `P$KeySrc { region_id: 8192 }`, but serialized `QuestInfo.key_cards` contains only regions 32, 32768, and 512. The card is authored `MOVE`-only (`P$FrobInfo.world_action.bits = 1`). Flat pickup routes MOVE-only items directly through `StoreItem`, so the injected `internal_keycard` script never receives `Frob` and never emits `AcquireKeyCard`.

The slot is working as authored: it is locked, requires region 8192, emits `hackfail` when the region is absent, and SwitchLinks to door 144.

Validated during play-through iteration 19 on commit `072480caa1c2b2625d5e6630debf69250507062b` with debug-runtime binary SHA-256 `2957df2ac1a8c54bd230d2913abd695bf414acf86eb5c92f3c34408e1782b950`.

## Expected

Any world pickup carrying `P$KeySrc` must execute the keycard acquisition path even when its authored frob flags contain MOVE without SCRIPT, preserving normal physical transfer while granting the key region.

tommy-xr:
## Summary

- route any world pickup carrying `PropKeySrc` through the runtime-injected `internal_keycard` Frob path, including flat world use and VR grip
- grant the key region and retain physical MOVE keycards in the backpack as one scripted operation
- keep the injected keycard script as sole transfer owner while preserving ordinary MOVE, weapon, and MOVE | SCRIPT pickup behavior
- exercise the real Command2 Bridge Card, authored slot, and linked security door through production flat input

## Root cause

Command2 mission object 135 inherits `PropFrobInfo.world_action = MOVE` and carries `PropKeySrc` region 8192. `FlatPlayerController::pick_up` stored every non-weapon directly, so the injected keycard script never received Frob and `AcquireKeyCard` never updated `QuestInfo`. The physical Bridge Card survived in inventory, but slot 138 correctly rejected region 8192 and left door 144 closed.

The same bypass existed for a VR grip: it emitted `HoldItem` without running the injected script. Both pickup surfaces now use `PropKeySrc` as the general data-driven discriminator. There are no mission, object, region, or runtime-ID special cases in the product fix.

## Relationship to #688

PR #688 implemented the routing predicate on the non-default `playthrough/water-exit-680` stack, whose base already contained broader keycard-transfer and scripted-loot changes. It never reached `main`, and its two-file diff alone would destroy physical cards on current main. This PR reimplements the focused fix against current main: physical keycard transfer, sole ownership, and flat/VR routing are included, while the unrelated stacked MOVE | SCRIPT composition remains unchanged.

## Negative-first proof

Before the production change:

```text
cargo test -p shock2vr world_use_of_move_only_keycard_dispatches_frob_for_injected_script -- --nocapture
FAILED: a MOVE-only PropKeySrc pickup must run its injected keycard Frob path
```

After the fix, focused tests cover ordinary MOVE, weapon, ordinary MOVE | SCRIPT, MOVE-only keycard, VR grip routing, physical/use-only keycard fate, and sole transfer ownership.

## Verification

- `git diff --check`
- `cargo test -p shock2vr` — 388 passed
- `cd tools/shock2-sdk && npm test` — 26 passed, 158 opt-in tests skipped
- focused `command2-bridge-card.e2e.test.ts` — passed; real card pickup retained the card, removed its world body, and the real slot opened linked door 144
- full serialized SDK e2e — 180 passed, 3 expected SHODAN fixture skips, all 23 missions and the new Command2 scenario passed; the sole failure was unrelated `DebugForceChase` async pathfinding divergence, which passed its immediate exclusive rerun on the unchanged commit

No visual capture is required; this changes interaction/access state without altering rendered output.

Fixes #687

### DIFF
--- a/shock2vr/src/flat_player_controller.rs
+++ b/shock2vr/src/flat_player_controller.rs
@@ -23,7 +23,9 @@ use crate::{
     runtime_props::RuntimePropReloading,
     scripts::{Message, MessagePayload},
     util::resolve_proxy_entity,
-    virtual_hand::{VirtualHandEffect, can_grab_item, is_wieldable_weapon},
+    virtual_hand::{
+        VirtualHandEffect, can_grab_item, is_wieldable_weapon, world_pickup_requires_frob,
+    },
 };
 
 /// Fallback viewmodel framing offset (look space: +x right, +y up, -z forward),
@@ -118,7 +120,9 @@ impl FlatPlayerController {
     /// Apply the original flat pickup split: weapons become the first-person
     /// viewmodel, while ordinary loot goes straight into the backpack.
     fn pick_up(&mut self, world: &World, entity_id: EntityId) -> Vec<VirtualHandEffect> {
-        if is_wieldable_weapon(world, entity_id) {
+        if world_pickup_requires_frob(world, entity_id) {
+            vec![out_message(entity_id, MessagePayload::Frob)]
+        } else if is_wieldable_weapon(world, entity_id) {
             self.wield(entity_id)
         } else {
             vec![VirtualHandEffect::StoreItem { entity_id }]
@@ -287,7 +291,15 @@ fn is_frobbable(world: &World, entity_id: EntityId) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use dark::properties::PropPlayerGun;
+    use dark::properties::{FrobFlag, KeyCard, PropFrobInfo, PropKeySrc, PropPlayerGun};
+
+    fn frob_info(world_action: FrobFlag) -> PropFrobInfo {
+        PropFrobInfo {
+            world_action,
+            inventory_action: FrobFlag::empty(),
+            tool_action: FrobFlag::empty(),
+        }
+    }
 
     fn pistol() -> PropPlayerGun {
         PropPlayerGun {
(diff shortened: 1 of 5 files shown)


tommy-xr:

Data-driven fix for a hard progression blocker, with the right test shape (negative-first proof quoted, plus non-regression pins for ordinary MOVE / auto-wield / direct transfer). Authored data verified: command2 obj 135 → template -2594 `PropKeySrc(region 8192)` inheriting `MOVE` from -157.

Before merge:
- **Rebase** (currently conflicting with main).
- **Surface the VR change explicitly**: `virtual_hand.rs:462` means gripping a keycard in VR no longer holds it — it frobs straight to the backpack. Intentional per the body, but it removes physically holding/inserting a card in VR and deserves a call-out under its own heading, not a bullet.
- `internal_keycard_script.rs:188`: re-frobbing a stored card can reach the `DestroyEntity` branch via `can_grab_item` — untested; please add a case.

Nit: `world_pickup_requires_frob` as `pub(crate)` in `virtual_hand` called from a script is an odd dependency direction — `script_util` is the natural home. Sequencing: coordinate with #743 and #772, which touch the same keycard/`drop_entity_into_container` subsystem.

tommy-xr:
The behavior for VR is incorrect - keycards should never go to inventory. They should set the questbit and then be removed (should happen in VR instead of holding the item)

tommy-xr:
superseded by #688

PR #761, opened 2026-08-01
fix(physics): traverse diagonal low treads

### LINKED ISSUE #517: earth intro: on-foot navigation residuals (capsule maze low-walls, post-return enlistment path, over-conservative /v1/player/move)
During the play-through campaign, earth.mis's montage-loop bug (#515) was fixed and the level now reaches the station.mis transition. But a fully-pure on-foot run of the earth intro is still blocked by **character-controller / debug-harness navigation limitations** (NOT mission logic). Tracking them here; non-blocking for campaign progress (station is reachable).

## 1. Capsule stalls on the montage maze's ~1u low walls (Medium)
In the "years of service" montage maze (~z 235–248), the player capsule cannot reliably step over ~1-unit-high low wall bands (notably z≈240). Real thumbstick locomotion with step-probing sometimes crosses; `/v1/player/move` never does. Same class as the known rapier resting-contact-freeze-on-angled-cuboids limitation (see character-controller work). A player CAN reach the return tripwire on foot with effort, but it's unreliable.

## 2. Post-montage-return → enlistment on-foot path uncertain (Medium)
After the scripted return teleport, the player lands in the BASIC TRAINING booth at ~(11.6, 24, 65.8). The ChooseService branch markers (which fire the station transition) are at z≈68–84, just past the montage-ENTRY tripwire box (z≈66.7, spans ~63.9–69.5). Observed: the player CAN walk forward within the box to z≈67.6 and stays (entry tripwire correctly ignores them — they were never tracked). But reaching the markers beyond the box on foot was not achieved without a corrective teleport to the concourse. Needs verification: is the geometry/capsule blocking the walk-out to the markers, or is enlistment intended to be approached from the main concourse rather than from the BASIC TRAINING return booth? Distinguish geometry vs capsule vs intended-flow.

## 3. `/v1/player/move` shapecast is over-conservative (Low, but harms playtest reliability)
`/v1/player/move`'s shapecast stalls ~2u short of where live thumbstick walking reaches (e.g. gravshaft mouth: move stalls at z=12.7 vs thumbstick reaches z=14.8). This over-conservatism produced **two false "walled-off / unreachable" conclusions** during this campaign (the vestibule booth and the montage exit), each of which required a follow-up triage to overturn. Making the debug-runtime move endpoint match real locomotion reach (or documenting that playtests must use thumbstick + step-probe, not move, for reachability claims) would materially improve automated-playtest reliability.

Found during the play-through campaign (earth iterations 1–2). Reference: #515 (fixed montage loop), #514 (droid-escort guidance).

tommy-xr:
## Summary

- detect diagonal low-tread obstruction from the full resolved movement vector, so preserved sideways sliding no longer hides a lost component
- keep validated movement walking while the real controller is still advancing around a finite wall, while retaining the strict requested-radius safety bound
- accept a substep-scale target residual as arrival instead of reporting a crossed walkable tread as blocked
- document that `/v1/player/move` is a local fixed-heading primitive, plus the authored Earth field sequence and backward post-Basic return route

## Reproduction and diagnosis

At the Earth Basic Training tread near z 240.4, the old step test saw roughly 80 percent projected progress because the capsule slid sideways, even though it lost the entire component needed to climb. The negative-first synthetic fixture ended near y 1.004 / z -0.357 without climbing.

The validated-move loop separately treated low straight-line target progress as a stall even while the controller was moving laterally around a finite wall. The negative-first fixture stopped at x 0.855 / z -0.410; the corrected loop rounds the end to x 2.488 / z 3.398.

The Earth course initially has intentional BlueForceField mission objects 548 and 566 over the tread. The runtime regression discovers and activates Destroy Trap mission object 300 first, confirms those fields are removed, and only then checks the exposed one-foot tread. It also uses stable mission template IDs rather than runtime entity IDs.

The post-course return is authored correctly: walk back through tripwire 359 and door leaves 358/368 to the hub, then use the x approximately 5 service passage. That route is now explicit in the playtest walkthrough.

## Validation

- `cargo test -p shock2vr` — 384 passed before rebase
- post-rebase `cargo test -p shock2vr physics::tests` — 53 passed
- SDK `npm test` — 26 passed, 158 gated e2e skipped
- focused post-rebase player-move e2e — 3 passed
- full SDK e2e — 181 passed, 0 failed, 3 skipped

No visual capture is included because this changes invisible collision endpoint behavior; deterministic runtime positions are the review evidence.

Fixes #517

### DIFF
--- a/.claude/skills/play-through/walkthroughs/earth.md
+++ b/.claude/skills/play-through/walkthroughs/earth.md
@@ -91,6 +91,12 @@ Follow the instructor prompts and demonstrate each available lesson:
    trap 379 must return the player once to approximately
    **(11.61, 24.26, 65.79)**, without an automatic re-entry loop.
 
+After that return, leave the Basic booth by walking **back/decreasing z**
+through tripwire 359 and its door leaves 358/368, then cross the hub to the
+x≈5 service passage described in Phase 4. Walking forward through the old
+Basic-entry volume is not the route to enlistment; its correctly ignored
+tripwire is not evidence that the concourse is blocked.
+
 The course begins behind timed fields. Wait for the authored delays, operate
 the highlighted Simple Button through player Frob, cross the lesson tripwires,
 climb the real ladder, and traverse the upper platform. Do not damage or
@@ -140,8 +146,9 @@ inventory; prove each exercise before leaving its room.
 Each lobby aperture has an authored **0.8-world-unit raised threshold**. Back
 off, center on the room's tripwire z coordinate, face west, and use sustained
 normal locomotion so the character controller can autostep onto it. The debug
-`/v1/player/move` helper performs a direct capsule shape cast without autostep,
-so a blocked helper move is not evidence that these entrances are impassable.
+`/v1/player/move` helper uses the same controller and step probe, but remains a
+single fixed-heading hop; a blocked direct hop is not evidence that these
+entrances are impassable without a centered or thumbstick retry.
 
 ### Weapons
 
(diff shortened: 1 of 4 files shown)


tommy-xr:
❌ **Triage: not solid as filed — global locomotion change, under-verified** (cross-PR review sweep, 2026-08-02)

The Earth tread fix is real, but the mechanism is a material loosening of the autostep gate that applies to *every grounded frame in every mission*, not just the tread:
- `physics/mod.rs:390`: the early-return flips from "applied ≥ 50% of desired along heading → skip probe" to "vector residual ≤ 25% → skip probe". A frame keeping 70% of its speed used to skip; now a 30% residual runs the step probe — so it fires on essentially every wall-slide and many slope frames. That's both per-frame cost on Quest and a wider surface for spurious autostep hops ("player ratchets up things they shouldn't").
- The defense is two synthetic fixtures plus an e2e that opens with `let Some(level) = try_load_level("earth.mis") else { return; }` — it **silently passes wherever Data/ is absent**, so it guards nothing in CI.
- `PLAYER_MOVE_TARGET_EPSILON = 0.1` (5× the retained arrival epsilon) changes the `blocked` predicate with only "substep-scale" as justification; stall detection now runs until the radius break instead of 3 substeps when sliding laterally.

What would make this mergeable: evidence that ordinary walking/slopes/wall-sliding didn't change — e.g. a before/after sweep of the existing movement e2es plus a frame-cost measurement, and the e2e failing loudly (not skipping) without data.

PR #762, opened 2026-08-01
feat(ui): open reader for received emails

### LINKED ISSUE #494: Flat UI: email reader overlay (emails are currently audio-only)
Follow-up to #468, which shipped the **audio-log** reader MFD. Emails still play audio only (`EmailTrap` -> `Effect::PlayEmail`) with no readable overlay.

## Original behavior

In the original game, receiving an email opens the same reader overlay as an audio log, but with the **EMAIL backdrop** instead of the LOG one, and the header/transcript resolved from the per-deck string files:

- Backdrop art: `EMAIL.PCX` (188x296, in `iface.crf` - request it via the archive-qualified `iface/email.pcx` key introduced in #468, since basenames can collide across archive mounts)
- Strings: `level<deck:02>.str` -> `EmailName<n>` / `EmailText<n>` / `EmailPortrait<n>` / `EmailIcon<n>` (same shape as the `Log*` keys; verified present in the shipped string files, e.g. `EmailName15` "POLITO 12.JUL.14 / re: Ship damage" on deck 2)
- Portraits/deck icons come from the same `book.crf` art the log reader already uses (`PEMAIL0/1.PCX` in iface.crf are the HUD pickup icons, not the reader backdrop)

## Suggested shape

`MediaGui` already renders backdrop + portrait + icon + header + word-wrapped scrollable transcript off `RuntimePropLogData`; emails need:

1. A media-kind (log vs email) on `RuntimePropLogData` (or a parallel prop) selecting the backdrop and the `Email*` string keys - the `PropLog.email` field (33 = unset sentinel, same decode as `log`) identifies email entries.
2. An open path: emails arrive via `EmailTrap` `TurnOn` rather than a frobbable disc, so the reader must be openable without a frobbed world object (e.g. bind the panel to the trap entity, or surface received emails for replay like the log collection in `QuestInfo`).
3. E2e: extend `tools/shock2-sdk/test/log-reader.e2e.test.ts` - medsci1's EmailTrap 1131 (EM0201) already exercised there for audio.

tommy-xr:
Fixes #494

## Reproduction

On current `origin/main`, triggering medsci1's authored EmailTrap 1131 plays `EM0201`, but `PlayEmail` does no UI work. The active MFD remains the previously opened Amanpour log (template 1608), so no `EmailName1` / `EmailText1`, portrait, icon, or EMAIL backdrop is shown.

Negative-first, `log-reader.e2e.test.ts` was extended before the product change. Its real-runtime run failed after the audio assertion passed:

```text
expected synthetic panel template_id -1
actual 1608
```

## Fix

- Add a media kind to the reader's derived runtime data so the existing `MediaGui` selects archive-qualified `iface/log.pcx` or `iface/email.pcx` while reusing the faithful portrait/icon/header/word-wrapped transcript layout.
- Create one flat-only, nonserialized synthetic email-reader host per mission. Email traps remain one-shot and can destroy themselves without closing the displayed message.
- On a newly played email, resolve `EmailName<n>`, `EmailText<n>`, `EmailPortrait<n>`, and `EmailIcon<n>` from `level<deck>.str`, attach them to the host, reset the reader to page one, and open it as an unbound MFD.
- Treat decoded `PropLog.email == 33` as the unset bitmask sentinel rather than attempting `EMxx33`.
- Extend the existing log-reader scenario and update map/elevator scenarios for authentic incoming-email takeover of the game's single MFD slot (EM0215 and EM0203 respectively).

## Verification

- `cargo fmt --all -- --check` ✅
- `RUSTFLAGS='-D warnings' cargo check -p shock2vr -p desktop_runtime -p debug_runtime` ✅
- `cargo test -p shock2vr` ✅ (386 passed)
- focused `MediaGui` tests ✅ (9 passed)
- focused `TrapEmail` tests ✅ (2 passed)
- `npm test` in `tools/shock2-sdk` ✅ (26 passed; 157 e2e-gated skips)
- affected UI e2es together ✅: email/log reader, automap interruption/reopen, elevator email/replacement (3 passed)
- full serial `npm run test:e2e` ✅ (183 total; 180 passed; 0 failed; 3 expected SHODAN-finale skips; all 23 missions loaded)

## Visual evidence

Incoming email reader demo

<sub>Receiving the authored EM0201 opens the localized EMAIL reader while the one-shot trap is consumed. Captured headlessly via deterministic fixed-timestep stepping.</sub>

## Before → after

Incoming email reader before/after

### DIFF
--- a/projects/flat-ui-panels.md
+++ b/projects/flat-ui-panels.md
@@ -563,6 +563,13 @@ QuestInfo; EmailTrap 1131 TurnOn → `EM0201` plays once, replay-deduped.
 (If audio observability proves awkward, the PR may defer audio *assertion* —
 not audio playback — and must say so.)
 
+Implemented follow-up: audio-log pickup/reader and collection shipped in #468;
+received-email presentation is completed by #494. `PlayEmail` now resolves the
+per-deck `Email*` strings onto a synthetic nonserialized `MediaGui` host and
+opens it as an unbound flat MFD with the archive-qualified `iface/email.pcx`
+backdrop. The one-shot authored `EmailTrap` can still be destroyed immediately
+without closing the reader.
+
 **PR B2 — Map panel** (~600–900 LOC, stacks on B1's QuestInfo edits)
 Files: new `scripts/gui/map.rs` (`MapGui` composing `MapChunkData` +
 `PAGE001`/decals/markers into `GuiComponent`s — port `map_renderer.rs` logic to
--- a/shock2vr/src/gui/gui_script.rs
+++ b/shock2vr/src/gui/gui_script.rs
@@ -111,6 +111,10 @@ where
                 self.gui.prepare_state_on_frob(&mut self.state);
                 Effect::combine(vec![Effect::OpenPanel { entity: entity_id }, frob_effect])
             }
+            MessagePayload::OpenUnboundPanel => {
+                self.gui.prepare_state_on_frob(&mut self.state);
+                Effect::OpenUnboundPanel { entity: entity_id }
+            }
             MessagePayload::GUIHover {
                 held_entity_id,
                 screen_coordinates,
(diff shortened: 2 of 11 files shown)


tommy-xr:
Actually, the original game didn't do this - scrapping

PR #767, opened 2026-08-01
fix(physics): restore model-bounds colliders

### LINKED ISSUE #597: Decide whether the model-bounds collider fallback should apply beyond climbables (~187 objects)
#595 made collider creation fall back to the model's bounding box when `PropPhysDimensions` is absent, but **gated the fallback on `PropPhysAttr.climbable != 0`** so it only affects ladders. The general case is deferred here.

### Measured scope

With the fallback ungated, **~187 non-ladder objects across all 21 missions** gain colliders that they do not have today — hydro2 pipe runs, station windows and crates, bar stools, and similar. That was measured on the branch, not estimated.

### The question

That is plausibly *faithful* — DromEd appears to default physics dimensions from the model, which is exactly why the ladders that do carry authored dims have `PhysDims.size` equal to their model bbox after the Z-up→Y-up swizzle. So these props arguably *should* be solid.

But it is also **traversal-affecting**: 187 newly-solid objects could seal corridors, block doorways, or wedge the player, and this port already has issues where an unexpected collider blocks a route (#592 grav-lift walls, #562 corpses sealing a corridor). Turning them all solid at once, in a change whose purpose was fixing ladders, was too risky to bundle.

### What is needed

Decide whether the fallback should be general, and if so land it as its own change with traversal testing — at minimum the SDK mission smoke test plus a play-through pass on the decks with the most affected objects (hydro2, station).

Split out of #589 / #595 as deferred work.

tommy-xr:
Fixes #597

## Summary

- restore the model-bounds collider fallback for every supported dimensionless OBB or sphere, instead of limiting it to climbables
- preserve authored `PhysType::None`, model-less marker suppression, authored dimensions, and the existing safe kinematic policy for dimensionless spheres
- add live SDK regression coverage in Station and Hydro2, including a `PhysType::None` negative control

## Why this is faithful

The original Dark Engine source marks both PhysType and PhysDims as instantiated, non-inherited properties. Its PhysType listener creates an instance PhysDims when a concrete physical object loads, and `InitPhysProperty(PHYS_DIMS)` initializes dimensions from `ObjGetUnscaledDims` plus object scale ([source](https://github.com/dima424658/darkengine/blob/4aa92d74e727a503954eb01914f69b3997324fc3/src/physics/phprop.cpp#L117-L137), [listener](https://github.com/dima424658/darkengine/blob/4aa92d74e727a503954eb01914f69b3997324fc3/src/physics/phprop.cpp#L557-L631), [dimension initialization](https://github.com/dima424658/darkengine/blob/4aa92d74e727a503954eb01914f69b3997324fc3/src/physics/phprop.cpp#L786-L870)). Raw mission objects that inherit a physical type but omit PhysDims are therefore not intentionally nonphysical; Dark materializes their model-derived instance dimensions at load.

The fallback is gated to exact supported OBB and sphere types. Raw type 3 (`None`, currently displayed as the combined bitflags value) remains bodyless even when a model exists, avoiding the warning noise and false railing/faucet/window bodies an ungated model lookup would introduce.

## Measured scope

An instrumented all-mission audit found 261 non-climbable, non-frobbable, positioned objects with inherited PhysType, no PhysDims, and usable model bounds:

- 207 exact OBB + 6 exact sphere = 213 supported objects restored
- 48 exact `PhysType::None` objects remain bodyless
- largest deltas: Hydro2 +95, Station +46, SHODAN +23, Ops3 +12

An exact base-versus-branch runtime comparison over all 23 missions added 213 physics bodies, matching the supported audit one-for-one, with no losses or unexplained gains. All new bodies are kinematic, so dimensionless props do not unexpectedly begin falling.

## Traversal verification

- Station: opened both real Security Doors through normal aim + squeeze, walked directly beneath and around the restored pipe bank, confirmed mission object 583 owns an enabled non-sensor body, and reached the Main Hall lift boundary without a new obstruction.
- Hydro2: opened the real elevator through its button, traversed out and south to the Sector A upper route beside the 95 restored props, confirmed mission object 706 owns its body, and raycast the only stop to existing `WorldRep` geometry rather than a restored pipe. Hand Rail object 1007 remained bodyless.

## Tests

- negative-first unit: the ordinary dimensionless OBB assertion fails on base and passes with the fallback
- `cargo test -p shock2vr` — 398 passed
- `npm test` — 26 passed, 163 opt-in skipped
- focused live model-bounds SDK scenario — 2 passed
- full `npm run test:e2e` — 186 passed, 0 failed, 3 expected skips; all 23 missions and both new regressions passed

Post-rebase focused Rust, warning-free checks, SDK units, and both live model-bounds scenarios are also green.

## Visual verification

Model-bounds collider restoration demo

<sub>Matched deterministic capture in `earth.mis` against Bar Stool mission object 353. From the same approach, base `2e5e394` clips through the visible stool (1.206u advanced) while fix `3524cea` stops against its restored kinematic, non-sensor collider (0.095u advanced). Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`).</sub>

### Restored behavior

Restored solid Bar Stool

### Before → after

Before and after model-bounds collision

### DIFF
--- a/shock2vr/src/mission/entity_creator.rs
+++ b/shock2vr/src/mission/entity_creator.rs
@@ -977,24 +977,22 @@ pub fn create_physics_representation(
             .map(|pa| pa.climbable != 0)
             .unwrap_or(false);
 
-        // `P$PhysDims` is optional: an author who never opened the physics
-        // dimensions dialog leaves the object with only a physics *type*, and
-        // the object then got no collider at all - which is why most ladders
-        // were neither solid nor climbable (issue #589). The original engine
-        // falls back to the model's bounds, and the shipped data proves it:
-        // every ladder that *does* carry authored dimensions carries exactly
-        // its model's bbox (e.g. eng1's `Ladder 16'` #945: size
-        // (1.65, 6.40, 0.14) == ladder.bin's bounds, offset zero == its
-        // bbox center).
-        //
-        // Deliberately limited to *climbable* objects. The same fallback
-        // applied to every dimension-less physics object turns ~187 further
-        // props solid across the shipped missions (hydro2 pipe runs, station
-        // windows and crates, bar stools) - plausibly also faithful, but a
-        // separate change with its own traversal testing, not a silent
-        // side effect of a ladder fix.
+        // `P$PhysDims` is an instantiated, non-inherited property in Dark. A
+        // concrete object can therefore inherit a physics type without storing
+        // dimensions in the mission: the original PhysType listener loads its
+        // model and initializes an instance PhysDims from the scaled model
+        // bounds. Do the same here for every supported physical model, not only
+        // climbable ones (#597). Objects with PhysType `None` and model-less
+        // markers still receive no fallback.
         let maybe_dimensions = v_dimensions.get(entity_id).ok();
-        let model_bounds = if maybe_dimensions.is_none() && is_climbable {
+        let has_supported_physics_type = v_phys_type
+            .get(entity_id)
+            .map(|phys_type| {
+                phys_type.phys_type == PhysicsModelType::ORIENTED_BOUNDING_BOX
+                    || phys_type.phys_type == PhysicsModelType::SPHERE
+            })
+            .unwrap_or(false);
+        let model_bounds = if maybe_dimensions.is_none() && has_supported_physics_type {
             // Raw model bounds - deliberately NOT `abs_dimensions`, whose
             // minimum-size clamp would make a fallback ladder 41% thicker
             // than the authored ladder standing next to it. Degenerate sizes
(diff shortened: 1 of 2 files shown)


tommy-xr:
✅ **Triage: merge-ready** (cross-PR review sweep, 2026-08-02)

Smallest diff of the open physics PRs with the strongest evidence: one predicate change, negative-first unit pair plus `PhysType::None` control, and the 213-collider audit reconciled across all 23 missions. Verified the `bitflags` ordinal comparison is exact (`==` on the whole pattern), so NONE(3) is correctly excluded — though the bitflags-for-an-enum modelling is a latent trap for a future `.contains()`.

Restacked on current main (clean rebase over 8 commits); post-rebase: `-D warnings` check clean, all 12 entity_creator unit tests pass, both e2e scenarios (station pipes, hydro2 supported/None) pass.

Residual risk: a restored collider sealing a corridor in one of the 21 unwalked missions — playtest-discoverable. Sequence deliberately with #695 (touches the same fallback in the opposite direction) and re-run #768's hydro2 e2e after this lands (+95 colliders in that mission).

Issue #744, opened 2026-07-30
fix(gameplay): prevent Cargo 2B jump from escaping through ceiling


tommy-xr:
## Summary

An ordinary forward jump west from the top stop of Engineering Cargo 2B East lift 669 passes through the room's parentless world ceiling and leaves the player stably supported on its exterior roof.

This is a bypassable gameplay collision bug. The valid route walks north from the lift toward top call button 620, then follows the upper aisle west.

## Campaign

- Scenario: `engineering`
- Tweak: `none`
- Assets: `legacy`
- Seed: `1378752978`
- Mission: `eng2.mis`
- Jump dependency: PR #714 at `cd19e396`, stacked on PR #712

## Stable authored identities

- Cargo Lift mission object `669`
- Lift path: object `477` at `y=-12.6` → `478` at `y=-5.1` → `479` at `y=2.1`
- Call buttons: bottom `480`, middle `481`, top `620`

These are mission-file object/template identities; no runtime entity ID is hardcoded.

## Reproduction

1. Launch `eng2.mis` with the stacked ordinary-jump implementation from PR #714.
2. Drive lift 669 to authored top node 479.
3. Stand at approximately `(43.9006, 3.1440, -168.2003)`.
4. Aim ordinary locomotion west toward approximately `(39, 3.2, -170)`.
5. Hold forward and pulse the production `Jump` input once; release it after the rising edge.
6. Let the simulation settle.

The durable negative-first SDK scenario additionally raycasts with only the `world` collision group from both sides and verifies that the traversed slab is the world ceiling/exterior roof at `y=6.8`.

## Expected

The upward character cast collides with the ceiling and cancels the jump rise, leaving the player inside Cargo 2B.

## Actual

The player crosses the `y=6.8` world slab and remains stably supported outside. The independent negative run ended:

```text
(43.900566, 3.144000, -168.200300)
  -> (34.538307, 8.043818, -171.638290)
  -> unchanged for another 120 fixed-timestep frames
```

The held-edge controller launches only once, the lift is settled at node 479 before the jump, and the result persists with zero input, ruling out repeated-jump input, lift carry, save-state residue, and a transient expected arc.

![Player supported outside on Cargo 2B's roof](https://gist.githubusercontent.com/tommy-xr/5ecdcd28b6a4c19ba80634ef371f464d/raw/campaign-ceiling-escape.png)

## Suspected cause

[...]

PR #768, opened 2026-08-01
fix(physics): preserve Dark mantle compression

### LINKED ISSUE #765: Hydro2 Sector C ladder approach is narrower than the player collider
## Summary

The normal Hydro2 route to the Sector C ladder is blocked by the authored partition/rail opening: the production player collider cannot fit through the final approach even though the shipped AIPATH graph routes through it.

## Reproduction

1. Load `hydro2.mis` and follow the ordinary world route toward the Sector C ladder.
2. From head position approximately `(56.125, -0.756, 24.101)`, walk the connected navigation route through anchors near `(56.64, 24.10)`, `(57.58, 24.04)`, and `(58.20, 22.80)` toward the ladder base `(59.60, 22.77)`.
3. Try normal walking, turning, and ordinary jump/mantle input through the partition/rail opening.

## Actual

The player becomes wedged against the opening and cannot reach the ladder. Releasing and re-approaching at multiple angles and jumping/mantling do not clear it. This blocks authentic campaign progress through Hydroponics.

## Expected

The production player collision shape should fit through the authored route and reach the Sector C ladder without teleporting or another debug bypass.

## Evidence

Fresh campaign play-through iteration 15 reproduced the blocker, then cold-loaded the same local state and reproduced it again. The local reproduction state is private retail-derived save data and will not be published. A deterministic SDK regression and before/after debug-runtime capture will accompany the fix PR.

## Build

Reproduced on campaign integration commit `a8726f243e1271a089a60f073a2acf44dc0f066a`.

tommy-xr:
Fixes #765

## Summary

- model Dark's ordinary mantle as an atomic collapse onto the existing head sphere instead of sweeping the capsule center through a nearby lip
- validate the exact compressed endpoint against newly introduced collider surface features, while continuing to sweep every raised/forward transition step
- retry higher bounded mantle routes when an early point-probe candidate cannot complete, and safely reverse the atomic collapse if live entity geometry blocks recovery
- add a Hydro2 Sector C mission regression that identifies both authored railing OBBs by stable template ID, proves a supported in-span vault, and continues to the ladder base

## Verification

- `cargo test -p shock2vr` (411 passed)
- `npm test` (26 passed; E2E scenarios skipped as intended without `SHOCK2_E2E=1`)
- `npm run test:e2e` from a clean rebuilt `dist` with an isolated target and ports (186 passed, 0 failed, 3 retail-save-gated skips)
- focused `jump.e2e.test.js` (4 passed: Hydro2 railing, SHODAN descent, Engineering crate, SHODAN log)
- focused `climbing.e2e.test.js` (2 passed)
- hosted CI: format and all build jobs pass on macOS, Ubuntu, Windows, and the Android build job
- private reproduction save remained unchanged at SHA-256 `341b6586037f62a5da8126afd171d7815b2a3b431754314f4d285317afcf5bbc` (save not published)

## Visual evidence

Hydro2 Sector C railing vault before and after

<sub>Ordinary jump now vaults Hydro2 Sector C railing and reaches the ladder base. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

## Before → after

Hydro2 Sector C railing vault before and after still

### DIFF
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -586,6 +586,9 @@ struct ClimbTopOut {
     save_pose: Vector<Real>,
     reversing: bool,
     is_crouched: bool,
+    /// Dark atomically collapses an ordinary mantle's sphere stack at the
+    /// existing head sphere; ladder recovery continues to sweep normally.
+    instant_compression: bool,
 }
 
 struct PlayerMovement {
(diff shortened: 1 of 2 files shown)


tommy-xr:
⚠️ **Triage: fix-first + sequencing** (cross-PR review sweep, 2026-08-02)

Behavior and tests are solid (strongest suite of the physics set), but:

- **The new subset guard is a tautology**: `physics/mod.rs:1191` places `head` at exactly the capsule's top-sphere center with the same radius, so `head_obstructions ⊆ current_obstructions` always holds and the `!is_subset → return None` branch can never fire — ~50 lines of `shape_obstructions` machinery that never rejects anything. Delete it or make it a real assertion.
- **Sequencing with #745 is semantic, not just textual**: both rewrite `plan_jump_mantle` from the same pre-image. This PR's `simulate_waypoint` sweeps `scripted_queries`, which excludes parentless world terrain, so it cannot catch the ceilings #745 blocks — and the retry-higher-rises search widens exactly the space that produced the #744 escape. Land #745 first, rebase this onto it, re-run the eng1/eng2 ceiling e2es.
- Per-jump cost moved inside the search loop (`simulate_waypoint` per rise iteration, `preflight_transition` per surviving candidate ≈ 5×512 casts worst case) — unmeasured on Quest.
- VR comfort: the first mantle frame is an unswept ~1.8 ft instantaneous camera pop (and reversal snaps back) — not mentioned in the body, no VR verification.

Also re-run the hydro2 e2e after #767 lands (+95 colliders in the same Sector C area).

PR #769, opened 2026-08-01
fix(debug): report carried entity locations

### LINKED ISSUE #565: /v1/entities lists inventoried items at stale world positions
`GET /v1/entities` keeps listing an item at its old world position after the item has been picked up into the player's inventory.

Reproduced during review: `POST /v1/player/give` on RadKey Card 81 → the card appears in `/v1/player/inventory` **and** still appears in the `/v1/entities` listing at its previous world position, with a distance-from-player value computed from that stale position.

### Expected

An item held in the player's inventory should either be excluded from the world-entity listing or clearly marked as carried, so automation can tell "still on the floor" from "already picked up".

### Impact

Automation-facing. The `playtest` / `play-through` skills navigate by querying entity positions; a phantom world position sends an agent walking back to an item it is already carrying, and makes "did the pickup work?" unanswerable from the listing alone.

Found by the automated play-through loop (eng1 iteration 0).

tommy-xr:
## Summary

- add `location` to `/v1/entities` summaries using the existing `world` / `inventory` / `left_hand` / `right_hand` vocabulary
- report backpack and nested carried items at the current player position with distance `0` instead of their retained floor coordinates
- preserve direct hand/viewmodel live transforms and keep every entity discoverable by runtime id and `byTemplate`
- document the SDK contract and cover both debug give and real flat aim+squeeze pickup paths

## Reproduction

On current main, a runtime-discovered MedSci `20 Nanites` item (stable mission/template id `1045`) initially listed at `[-15.801491, -6.14005, 20.971342]`, distance `19.46652`, while the player was at `[-34.96759, -4.7559557, 17.85841]`.

After `POST /v1/player/give`, `/v1/player/inventory` reported the same runtime entity with `location: inventory`, but `/v1/entities` still returned the original floor position and distance unchanged.

With this change, the same transition lists the entity as `location: inventory`, position equal to the current player position, and distance `0`. The runtime id and template lookup remain available.

## Verification

- `cargo test -p shock2vr -p debug_runtime` — 410 core tests and 1 debug-runtime test passed
- `npm test` — 26 passed, 0 failed; 162 opt-in e2e skipped
- focused post-fetch e2e: flat real aim+squeeze pickup, `/v1/player/give`, and muzzle/viewmodel tracking — 3 passed, 0 failed
- `npm run test:e2e` — 188 total: 185 passed, 0 failed, 3 authored SHODAN finale skips; all 23 mission loads passed

Fixes #565

## Semantics

This is a semantic change to the existing `/v1/entities` API, not just an added field: for carried items (backpack and nested containers) `position` and `distance` no longer report the retained floor coordinates - they now report the current player position and `0`. Callers that used `position`/`distance` to locate a carried item on the floor will see different values. Directly held (hand/viewmodel) items keep their live transform.

### DIFF
--- a/runtimes/debug_runtime/src/commands.rs
+++ b/runtimes/debug_runtime/src/commands.rs
@@ -723,6 +723,8 @@ pub struct EntitySummary {
     pub id: i32,
     pub name: String,
     pub template_id: i32,
+    /// "world", "inventory", "left_hand", or "right_hand".
+    pub location: String,
     pub position: [f32; 3],
     pub distance: f32,
     pub script_count: usize,
--- a/runtimes/debug_runtime/src/main.rs
+++ b/runtimes/debug_runtime/src/main.rs
@@ -1603,6 +1603,7 @@ fn process_command(
                             id: e.id,
                             name: e.name,
                             template_id: e.template_id,
+                            location: e.location,
                             position: e.position,
                             distance: e.distance,
                             script_count: e.script_count,
--- a/shock2vr/src/game_scene.rs
+++ b/shock2vr/src/game_scene.rs
@@ -276,6 +276,8 @@ pub struct DebugEntitySummary {
     pub id: i32,
     pub name: String,
     pub template_id: i32,
+    /// "world", "inventory", "left_hand", or "right_hand".
+    pub location: String,
     pub position: [f32; 3],
     pub distance: f32,
     pub script_count: usize,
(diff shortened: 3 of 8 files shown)


tommy-xr:

Correct and well-tested, but two items before merge:
- **Rebase onto #772** — both PRs rewrite the same 8 lines at the top of `give-item.e2e.test.ts` (verified conflict); #772 should land first and own that hunk.
- This is a **semantic change to an existing API**, not purely additive: `position`/`distance` for carried items now report the player's position / 0. Nothing in-tree breaks, but any external agent script reading `position` for a carried entity changes behavior — worth a line in the body.

Follow-up (file an issue, not a blocker): the `location` vocabulary has no `container` kind — an item inside a corpse/crate still reports `world` at a stale position, which is exactly the case #772 is about. Also note this hides the "entered inventory but failed to unphysicalize" bug class.

PR #772, opened 2026-08-01
fix(debug): preserve container loot on give

### LINKED ISSUE #572: [tooling] /v1/player/give permanently severs Contains links — silently invalidates container-loot testing
## Summary

`POST /v1/player/give` on an entity that lives **inside a container** (has an incoming `Contains` link) permanently and irreversibly severs that link. The container is then empty forever, in that save lineage and every save descended from it.

## Why this matters

This is not a game bug — it's a **testing-harness hazard**, and a nasty one, because it silently destroys the evidence that a later test depends on.

Confirmed in `shock2vr/src/mission/mission_core.rs:2079`: `drop_entity_into_container` drops all incoming `Contains` links for the item. That's correct for a *real* loot (give reuses the legitimate pickup path behind the `can_grab_item` gate) — the problem is that `give` lets automation do it to an item the player never actually looted.

## How it bit us

In the `play-through` campaign (Ops deck):

1. Iteration 3 used `/v1/player/give` on ops3's `Chip C` purely as a *wrong-chip probe* against `SimComp_3`'s type matching.
2. Iteration 4 then killed the chip's carrier (`Docile`, object 125) **legitimately** — and its corpse container opened **empty**, with no `Contains` link at all.
3. The mission's only Chip C could no longer be looted the way a player must. The acceptance criterion "Chip C obtained from its carrier's container" became **unprovable** from that entire save lineage.

Links are serialized into saves, so there is no recovery short of restarting from a pre-`give` save.

## Requested change

Make `/v1/player/give` refuse — or at minimum loudly warn — when the target entity has an **incoming `Contains` link**.

Rationale: an item inside a container is exactly the item whose acquisition a test most wants to verify through the real loot path. Handing it over via a debug lever both skips that path *and* destroys it for every future run.

A refusal with a clear message ("entity N is container-held; loot it via the container MFD") would have prevented this outright. An opt-in override flag would be acceptable if some workflow genuinely needs it, but the safe behavior should be the default.

## Note for automation authors

Container loot must go through the real path: get close → `Frob` the container → click the item by `label` in the `/v1/ui` MFD panel. `/v1/player/give` is for world-placed items only — and per #569, the intended world pickup is `right_hand.squeeze` against the crosshair target, not `Frob`.

Found by the `play-through` campaign (Ops deck, iterations 3–4); root cause confirmed in source by adversarial review.

tommy-xr:
## Summary

- reject `POST /v1/player/give` before mutation when the target has an incoming `Contains` link
- return an actionable 400 naming both runtime entities and directing automation through the container MFD
- keep ordinary world-item give behavior covered and move unrelated UI setup away from arbitrary mission loot

## Reproduction

On base `736b8fc`, the new negative-first SDK scenario discovered the Psi Amp through Male Corpse 1's live `Contains` link and `/v1/player/give` accepted it instead of rejecting it:

```text
AssertionError: Missing expected rejection: container-held items must be looted through the real container UI
```

`MissionCore::give_item` passed every grabbable entity directly to `drop_entity_into_container`, whose legitimate reparenting behavior removes all prior incoming `Contains` links. That made the debug shortcut permanently empty the authored container.

## Fix

`give_item` now scans the live concrete links for an incoming `Contains` relationship after pickup eligibility is established and before any inventory or physics mutation. It returns:

```text
entity <item> is container-held by entity <container>; loot it via the container MFD
```

The regression asserts the exact original link remains unchanged and the item never enters the backpack. A separate control gives the known world-placed MedSci Cryo Card successfully.

## Verification

- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p debug_runtime`
- `RUSTFLAGS="-D warnings" cargo test -p shock2vr -p debug_runtime` — shock2vr 407/407, debug_runtime 1/1 (one display-only test ignored)
- `npm test` in `tools/shock2-sdk` — 26 passed, 0 failed
- focused give-item e2e — 2 passed, 0 failed
- full SDK e2e — 181 passed initially, 5 exact failures rerun green, 3 existing skips; final accounting 186 passed, 0 unresolved failures, all 23 missions loaded
  - two failures were invalid arbitrary-Nanites `/give` setup and pass after provisioning fresh Nanites instead
  - two launches hit `ENOSPC` and pass after generated incremental artifacts were cleared
  - the unrelated #481 alertness-churn route observation passed its isolated rerun

## Visual proof

Non-renderable: this changes only the debug HTTP endpoint's error contract and whether an internal `Contains` link is mutated. A rendered frame cannot expose either the 400 response or link preservation. The player-facing loot MFD and ordinary world pickup presentation are intentionally unchanged; their existing e2e scenarios pass. The deterministic SDK before/after assertion is the direct proof for this tooling-only behavior.

Fixes #572

### DIFF
--- a/shock2vr/src/game_scene.rs
+++ b/shock2vr/src/game_scene.rs
@@ -746,7 +746,8 @@ pub trait DebuggableScene {
 
     /// Put an existing world entity into the player's inventory (a headless
     /// "pick up" for testing). Models a pickup - the item lands in the backpack;
-    /// wielding is a separate, presentation-specific concern. Default:
+    /// wielding is a separate, presentation-specific concern. Container-held
+    /// items are rejected so tests must exercise the real loot UI. Default:
     /// unsupported.
     fn give_item(&mut self, _entity_id: EntityId) -> Result<(), String> {
         Err("scene does not support giving items".to_string())
--- a/shock2vr/src/mission/mission_core.rs
+++ b/shock2vr/src/mission/mission_core.rs
@@ -6619,6 +6619,32 @@ impl crate::game_scene::DebuggableScene for MissionCore {
             return Err(format!("entity {:?} is not a pickup item", entity_id));
         }
 
+        // This debug lever is only for world-placed items. Moving an authored
+        // container's loot straight into the backpack would make
+        // `drop_entity_into_container` remove the original Contains link,
+        // permanently bypassing (and invalidating tests of) the real loot UI.
+        // Reject before mutating anything so the container remains intact.
+        let containing_entity = {
+            let links = self.world.borrow::<View<Links>>().unwrap();
+            links.iter().with_id().find_map(|(container_id, links)| {
+                links
+                    .to_links
+                    .iter()
+                    .any(|link| {
+                        matches!(link.link, Link::Contains(_))
+                            && link.to_entity_id.map(|wrapped| wrapped.0) == Some(entity_id)
+                    })
+                    .then_some(container_id)
+            })
+        };
+        if let Some(container_id) = containing_entity {
+            return Err(format!(
+                "entity {} is container-held by entity {}; loot it via the container MFD",
+                entity_id.inner(),
+                container_id.inner()
+            ));
+        }
+
         let inventory_entity = {
             let player = self.world.borrow::<UniqueView<PlayerInfo>>().unwrap();
             player.inventory_entity_id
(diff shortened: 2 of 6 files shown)


tommy-xr:
✅ **Triage: merge-ready** (cross-PR review sweep, 2026-08-02)

Debug-surface only (`DebuggableScene::give_item`), rejects before any mutation, and the test asserts the exact rejection plus byte-identical `Contains` links afterwards. The migrated tests correctly use `player.spawnItem`.

Sequencing: merge **before #769** — both rewrite the same 8 lines at the top of `give-item.e2e.test.ts` (verified conflict); this PR should own that hunk and #769 rebase onto it.

Nit: the guard matches *any* incoming `Contains` including the player's own inventory, so re-giving a carried item reports a misleading "container-held" message (rejection still correct).

PR #774, opened 2026-08-02
test(sdk): survive the Cargo 2B jump setup

tommy-xr:
## Summary

- park the player on the Cargo 2B top landing for the lift ride so the eng2 ceiling case's 1200-frame setup cannot kill the player it is about to measure
- assert the staged player is alive, so a future setup hazard reports itself instead of masquerading as a locomotion or mantle result

Stacked on `fix/cargo2-ceiling-escape` (PR #745), which owns the test.

## The drift

`jump-ceiling.e2e.test.ts` drives the real Cargo 2B lift (frob 480, step 600, frob 481, step 600) before teleporting the player onto its top stop. Throughout those 1200 fixed-timestep frames the player stands unattended at eng2's mission start.

Two changes have landed on `main` since the test was written:

- **#750** (`fix(ai): let hostile ranged attacks damage player`) - the wandering `OG-Shotgun` hybrid near the eng2 start now actually damages the player. Measured on `origin/main` @ `cd00795f`: it goes `Wander/Low -> RangedAttack/Moderate` around frame 305 and lands 6 damage roughly every 90 frames, taking the player from 30 hp to 0 by ~frame 865 - before the lift finishes its ride.
- **#753** (`fix(gameplay): handle player death and QBR respawn`) - a dead player's input is suppressed.

Together they make the setup fatal, and the *control* assertion is what fails:

```text
AssertionError: ordinary walking should leave the lift north toward top button 620,
got {"x":43.900566,"y":3.144,"z":-168.2003}
```

The player did not move a millimetre because it was a corpse (`life_state: "dead"`, 0 hp) - the jump under test was never exercised. This reproduces byte-identically on `origin/main` @ `cd00795f` with the test file dropped in.

## Validation

All runs `SHOCK2_E2E=1`, `DARK_ASSET_PATH` pointed at unpacked retail data.

| build | test | result |
| --- | --- | --- |
| `origin/main` @ `cd00795f` (no mantle bound) | original | FAIL on setup control - player dead, zero movement |
| `origin/main` @ `cd00795f` (no mantle bound) | **this PR** | setup + control pass, then FAILs on the real assertion: `(43.900566, 3.144, -168.2003) -> (34.538307, 8.043818, -171.638290)`, above the `y=6.8` ceiling - the guard is live |
| `fix/cargo2-ceiling-escape` @ `c42bfd82` | **this PR** | PASS, both eng2 and eng1 cases |

The staged pose settles at `(43.805, 3.244, -165.207)` on the top landing and holds 30/30 hp across the full 1200-frame ride, so the lift assertions (`y=-5.1` then `y=2.1`) and the campaign-observed jump pose are unchanged.

`npm test` (SDK unit suite): 26 pass / 0 fail. `tsc` clean on both bases. No Rust changed.

### DIFF
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -1154,10 +1154,11 @@ fn plan_climb_top_out(
 /// standing pose must fit against *all* colliders, and every scripted substep
 /// still collides with parented entities. Same-height and lower landings
 /// require a genuinely clear all-world point probe above the lip, so
-/// full-height walls remain solid. An elevated landing may use Dark's
-/// parentless-terrain jump-through probe: that is what permits authored
-/// stacked corridors such as shodan's log platforms, whose upper floor is a
-/// ceiling to the lower cell.
+/// full-height walls remain solid. An elevated landing may cross the local
+/// platform lip with Dark's parentless-terrain jump-through probe, but its
+/// initial vertical rise must remain clear against all world geometry. That
+/// is what permits authored stacked corridors such as shodan's log platforms
+/// without treating a room ceiling directly overhead as an exterior landing.
 fn plan_jump_mantle(
     controller: &KinematicCharacterController,
     validation_queries: &QueryPipeline,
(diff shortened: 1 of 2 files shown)


tommy-xr:
superseded by #800

Issue #783, opened 2026-08-02
[functionality] debug_runtime resident memory grows to multiple GiB over long sessions


tommy-xr:
During a ~35-minute stepped hydro2 session, host free disk fell to 0 bytes and only recovered when the runtime exited — final shutdown freed ~5.5 GiB (208 MiB → 5.7 GiB free), with stdout already redirected to /dev/null. That points at resident runtime memory being swapped out (macOS swap lives on the root volume), i.e. unbounded memory growth over a long session (screenshots, step loop, or asset/texture accumulation are candidates).

**Repro sketch:** launch `cargo dbgr --mission hydro2.mis`, load a mid-game save, alternate /v1/step + /v1/screenshot for ~30+ minutes, watch RSS/swap grow (e.g. `footprint`/`vmmap` snapshots every 5 min).

Med [functionality]. Needs profiling investigation; a concurrent build on the host contributed disk pressure but not the runtime-exit reclaim.

campaign: engineering-through-shodan · none · legacy · seed 1378752978

tommy-xr:
## Investigation: not reproducible as a runtime leak — the exit reclaim is host swap accounting

Profiled `debug_runtime` on `hydro2.mis` across five workloads (fresh `origin/main` @ `cd00795f`, release build, macOS 15.5 / 8 GiB / arm64), sampling `footprint(1)` — *physical footprint*, which counts compressed pages and GPU/IOAccelerator memory, unlike `ps rss` — plus `vmmap -summary` at start and end.

**Every workload holds flat in a 220–280 MiB band. There is no unbounded growth to find.**

### Measurements

| Variant | Duration / work | Footprint start → end | Peak |
|---|---|---|---|
| **Idle** (paused, zero HTTP requests) | 18 min | 225 → 232 MiB | 281 MiB |
| **step + screenshot, hammered** | 3 min, **540 cycles**, 32 400 sim frames | 233 → 231 MiB | — |
| **Campaign-like** (walk / turn / fire / screenshot / `/v1/entities`) | 25 min, **1 088 cycles** ≈ 3 264 steps + 1 088 screenshots + 1 088 entity queries | 218 → 249 MiB | 279 MiB |
| **Level reload ×12** (`DebugReloadLevel`, full mission teardown + rebuild) | 12 reloads | 222 → 249 MiB | 402 MiB |
[...]

PR #793, opened 2026-08-02
feat(sdk): reap stale/orphaned debug runtimes on launch

### LINKED ISSUE #786: [sdk] GameServer should reap stale/orphaned debug runtimes
Long automation campaigns accumulate orphaned `debug_runtime` processes when an agent dies without reaching `/v1/shutdown` (a 10-hour orphan at ~100% CPU was found during the #783 investigation; five runtimes were alive at once). The SDK's `GameServer.launch` already passes `--instance-id`; it should (a) support/perform reaping of prior instances by port and instance-id on launch (opt-in flag or default for same port), and (b) the playtest/play-through harness docs should instruct a reap step before launching.

Low [sdk] enhancement. Evidence: #783 diagnosis comment.

tommy-xr:
## Summary

- **Fixes #786**: `GameServer.launch` now reaps (SIGKILL) a stale `debug_runtime` already bound to the requested launch port before starting a new one, by default (`reapStale?: boolean`, default `true`). Long automation campaigns accumulate orphaned `debug_runtime` processes when an agent dies without reaching `/v1/shutdown` - a 10-hour orphan at ~100% CPU was found during the #783 investigation, with five runtimes alive at once.
- Only ever targets the **exact requested port** (`port ?? 8080`) - never a fallback port from the existing bind-race retry loop (`hint + attempt`), and never a port a live `GameServer` in the *same* process already owns (that's an in-flight sibling launch, not an orphan - the pre-existing `reservedPorts` in-process bookkeeping already makes the free-port walk-up skip it, so reaping it would only destroy a live instance for no benefit).
- Discovery: `lsof` for candidate pids bound to the port, then `ps -o command=` for each one's full command line (`lsof`'s own `COMMAND` column truncates to 9 chars and drops args, so it can't check for `--port <port>` in argv). A candidate is only killed if its executable basename is exactly `debug_runtime` **and** its args contain the literal token pair `--port <port>` - tokenized matching, not a raw substring check, so e.g. `debug_runtime_proxy` or a `--port 8080` vs. `--port 808` collision can't be mistaken for a match.
- Passing `reapStale: false` restores the old behavior (walk up to the next free port, never touch what's already there).

## Test plan

- [x] `npx tsc --noEmit` - clean
- [x] `npm test` (unit, 35 pass / 0 fail) - includes:
  - `reapStaleRuntimeOnPort` core logic: kills a matching candidate, never kills a non-`debug_runtime`, only kills the candidate with the exact matching `--port`, rejects substring look-alikes (executable name and port), no-op when nothing's listening, tolerates a kill failing, swallows discovery errors
  - Option plumbing (mocked): `GameServer.launch` calls the reap hook exactly once with the literal requested port; `reapStale: false` skips it entirely
- [x] `npm run test:e2e`'s new `reap-stale.e2e.test.ts` (`SHOCK2_E2E=1`, real processes, no mocking): spawns a genuine orphan `debug_runtime` directly via `child_process` (bypassing `GameServer.launch`, since going through it would register the port in the SDK's own same-process bookkeeping instead of exercising the real cross-process reap path), verifies the default path reaps it and the successor reclaims the *exact* port; a second test verifies `reapStale: false` leaves the orphan alone and walks up instead. Both verify ownership via `/v1/health`'s `instance_id` to stay safe on a shared/busy dev host. **Both pass.**
- [x] `cargo fmt --all` - no Rust files touched, no-op
- [x] README's lifecycle section updated to document reaping and its scope/risk

## Review

Cross-engine adversarial review (`/xreview`: Claude subagent + Codex CLI) both independently flagged that the initial version would kill *any* `debug_runtime` on the port including a live sibling launched by the same process - fixed via the `reservedPorts` same-process guard above. Both also converged on tightening the process-match from substring to exact-token matching - done. A follow-up Codex pass flagged the e2e test could mistake an unrelated pre-existing service on the port for its own fixture - fixed via the `instance_id` ownership check. Remaining lower-severity notes (a theoretical PID-reuse race between `lsof` and `kill`, and best-effort discovery failures degrading silently to the pre-existing walk-up behavior) were judged acceptable residual risk for a local dev-automation tool rather than requiring further complexity.

### DIFF
--- a/tools/shock2-sdk/README.md
+++ b/tools/shock2-sdk/README.md
@@ -126,6 +126,21 @@ const game = await GameServer.connect("http://127.0.0.1:8080");
 - `GameServer.launch` finds the cargo workspace by walking up from `cwd`;
   pass `repoRoot` to override. First launch may take minutes while cargo
   compiles; the default readiness timeout is 5 minutes.
+- **Lifecycle / reaping stale runtimes**: a long automation session that dies
+  without reaching `/v1/shutdown` (a crashed agent, a killed test runner)
+  leaves its `debug_runtime` orphaned - still bound to its port and burning a
+  CPU core. To guard against that, `launch()` reaps (SIGKILL) any
+  `debug_runtime` already bound to the requested port *before* starting a new
+  one, by default. This only ever targets the exact port you asked for
+  (`port ?? 8080`) - never a fallback port picked by the bind-race retry loop,
+  and never a process that isn't a `debug_runtime` by command line, and never
+  a port a live `GameServer` in *this* process already owns (that one is
+  skipped as "not stale," not killed). It does **not** distinguish an actual
+  orphan from another agent's currently-running session that happens to share
+  the same port - if you rely on the existing walk-up-to-a-free-port behavior
+  to coexist with concurrent runs on a shared default port, pass an explicit
+  unique `port` per run, or set `reapStale: false` to fall back to walk-up
+  instead of killing what's there.
 - Injected actions apply on the next game update, which runs even while
   paused (with zero delta time).
 - `game.logs()` returns recent runtime output (also included in launch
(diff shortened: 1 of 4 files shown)


tommy-xr:
Often have multiple agents running e2e tests and debug_runtime, this proposed fix would cause agents to close and terminate each other's running processes. Need to consider an alternative that doesn't have this downside.

PR #800, opened 2026-08-03
fix(physics): put the player camera at the head sphere

### LINKED ISSUE #795: [gameplay] Authored corpse containers cannot be frobbed by the flat interaction ray
## Summary

`HYD Male Corpse` / `HYD Female Corpse` entities (ContainerScript, model `corm_hyd`) are never selected by the flat-mode interaction ray: no highlight, no hover label, no container panel — verified across a 35-point aim sweep at 1.2u range and multiple positions. A debug Frob message opens the container instantly and lists its contents (Hydro Card B), so the container logic itself works; only ray selection fails. Crates, desks, and creature corpses select fine with the identical production frob, and the corpse has a physics body in the same kinematic/selectable group as a working desk.

**Impact: campaign blocker.** All three Hydro deck keycards (A/B/D) are stored in such corpses, gating the card chain → hydro1/hydro3 bulkheads → ACR1/ACR4.

## Likely regression

Earlier sessions (iterations 17 and 20 of this campaign) looted corpse containers fine on older builds; the tree has since advanced past #767 (`fix(physics): restore model-bounds colliders`, cd00795f). Check whether the restored model-bounds collider for `corm_hyd` changed the raycast target (e.g. ray now hits a collider not mapped back to the entity, or the corpse's selectable body was replaced).

## Repro

hydro2.mis, walk to a HYD corpse (e.g. near (73.08, −0.76, −14.92)) and aim at it in use mode — no selection. A hashed local repro save exists (`iter21_corpse_frob_repro`, SHA-256 c4e0fd2e714c7246d7f723c03667130984a76ac5df4b731bccaaa99c70a2db17, player at (73.08, −0.76, −14.92)).

High [gameplay] bug — blocker.

tommy-xr:
## Summary

hydro2's `HYD Male Corpse` (mission object 754, holds **Hydro Card B**) could not be highlighted or frobbed by the flat interaction ray. The container logic was never at fault - a debug `Frob` opened it fine.

Root cause: **the player's camera sat above the player's own head.** `PLAYER_EYE_HEIGHT` was a magic `4.0` SS2 ft above the body origin, but the standing collision capsule is `6.0` ft tall, so its crown is at `+3.0`. The eye was therefore **1.0 ft above the collider crown — 7.1 ft above the feet for a 6-foot body.** That corpse lies in an alcove with a **7-foot ceiling** (floor `-1.96`, ceiling `0.80`): the body clears it, the camera does not. The eye lands at `0.844`, *outside* the room, so the crosshair ray starts above the ceiling and hits it `0.05` units in front of the eye. Nothing in that alcove can be selected, and the view renders from above the slab.

The fix derives the eye from the original game's camera placement instead of a magic constant. The original anchors the first-person viewpoint to the player's **head sphere** — `PLAYER_HEAD_POS = (PLAYER_HEIGHT / 2) - PLAYER_RADIUS` = **1.8 ft** above the body origin, from the same `PLAYER_RADIUS 1.2` / `PLAYER_HEIGHT 6.0` collision profile this repo already matches — and then raises it by a separate **default eye offset of 0.8 ft** (configurable as `"eyeloc"`; our data ships no override). Standing eye = 1.8 + 0.8 = **2.6 ft above the body center, i.e. 5.6 ft above the floor**.

Both halves are named constants (`physics::PLAYER_HEAD_POS`, `physics::PLAYER_EYE_OFFSET`) and the head half is derived from the collision profile, so the camera stays *inside* the capsule by construction: 2.6 ft is still 0.4 ft below the 3.0 ft crown, and a future footprint change cannot push the eye back through ceilings.

Two consequences of moving the eye *into* the capsule are handled in the second commit:

- The flat projectile ray starts at the camera and includes the `PLAYER` group (so AI shots can hit the player). With the eye inside the capsule, Rapier's solid raycast reported the shooter's own collider as a distance-0 hit and **every flat shot resolved on the player** (terrain spangs, metal impact sounds). A camera-origin ray is by definition the player's own shot, so it now drops `PLAYER`; AI/turret projectiles keep it.
- `THROW_SPAWN_DISTANCE` (0.5) no longer cleared the capsule radius (0.48) for a downward throw; it is now derived from that radius.

## Regression window

Not #767 (`fix(physics): restore model-bounds colliders`) as the issue suspected - **it is #712** (`fix(physics): restore original player footprint`, `962a2a2b`). The eye offset is measured from the capsule **center**, so restoring the standing capsule from 4.8 ft to the authentic 6.0 ft raised the camera 0.6 ft and pushed it over this 7-foot ceiling.

Same deterministic probe (17 standing aim points ringing the corpse, production camera height, crosshair ray with the flat controller's mask):

| build | camera y | corpse selected |
| --- | --- | --- |
| `962a2a2b~1` (before #712) | 0.604 | **yes** (8/17, incl. ang 0 @ 1.5u) |
| `cd00795f~1` (after #712, before #767) | 0.844 | **no** at every level-floor point; ray stops on the ceiling at 0.051u |
| `cd00795f` (main, after #767) | 0.844 | **no** - byte-identical to the line above, so #767 is exonerated |
| this branch | 0.284 | **yes** |

Camera y for this branch is measured from the runtime (`/v1/info`): body center `-0.756` + the live `camera_offset` `1.04` world units (= 2.6 SS2 ft / 2.5), i.e. `2.24` world units (5.6 ft) above the `-1.96` floor and `0.52` units below the `0.80` ceiling. (An earlier revision of this table listed `0.684` for this row; that probe added the eye offset to a position that already included it.)

## Tests

- **Negative-first unit** (`shock2vr/src/physics/mod.rs`): a seven-foot room built from real `WorldRep`-style level geometry, a `selectable` container on the floor, and the flat controller's exact crosshair mask. With the old 4.0 ft eye it resolves to the ceiling (`hit ... at Point3 [0.033, 2.8, 0.0]`); with the fix it resolves to the container. A fixture precondition asserts the pre-fix eye really would have escaped this room, so the repro can't silently stop biting. A second test pins the eye to `PLAYER_HEAD_POS + PLAYER_EYE_OFFSET` and asserts it sits strictly below the capsule crown.
- **Negative-first unit** (`internal_fast_projectile.rs`): a shot fired from the player's own eye reaches the target, not the shooter; the existing "an AI/turret projectile must be able to hit the player" test still passes.
- **Negative-first e2e** (`tools/shock2-sdk/test/hydro2-corpse-frob.e2e.test.ts`): stands at the campaign repro spot, aims with the production `aimAt`, squeezes, and asserts the loot MFD opens bound to the corpse and lists Hydro Card B. On `cd00795f~1`: `active_panel: null`. On this branch: passes.
- `cargo fmt --all`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`, `cargo test -p shock2vr` (426 passed), SDK `npm test` (26 passed), full `npm run test:e2e`.
- Several SDK aim fixtures encoded the old camera height (hardcoded `body + 1.6` look targets, a fixed keypad pitch, the Earth world-use staging offset). They now use the exported `PLAYER_EYE_HEIGHT_WORLD` (0.72 -> **1.04** world units with the eye offset) or the production `aimAt`, and `/v1/info` now reports the **live** eye height so automation aims correctly while crouched too.

Fixes #795

## Visual verification

camera head-sphere fix demo

<sub>hydro2 alcove, this branch: the camera pans down onto the <code>HYD Male Corpse</code>, which now highlights (hover label + selection brackets), and a frob opens the SEARCH loot MFD listing <b>Hydro Card B</b>. Captured headlessly via the debug runtime (<code>/v1/step</code> + <code>/v1/screenshot</code>, deterministic fixed-timestep). <b>Note:</b> these captures predate the eye-offset commit, so they show the camera 0.8 ft (0.32 world units) lower than the current build; the alcove is reachable at both heights and the e2e frob test passes on the current build.</sub>

### Before → after

before/after

<sub>Same position and the same aim quaternion on both builds. Before (<code>origin/main~</code>): the eye is above the alcove's 7-ft ceiling, so the view is the featureless top of the ceiling slab and the crosshair ray never reaches the room — <code>/v1/ui</code> stays <code>active_panel: null</code> after a frob. After: the eye is at the original's standing height (head sphere + eye offset), inside the capsule and inside the alcove.</sub>

### DIFF
--- a/projects/flat-weapon-handling.md
+++ b/projects/flat-weapon-handling.md
@@ -106,7 +106,8 @@ Done:
   rotation — applying it over-rotates each gun by exactly its heading — so it is
   deliberately unused.
 - **Eye-height alignment** — the shot origin + viewmodel use the shared
-  `shock2vr::PLAYER_EYE_HEIGHT` (4.0 SS2 units), matching every runtime's render
+  `shock2vr::PLAYER_EYE_HEIGHT` (the original engine's `PLAYER_HEAD` sphere,
+  1.8 SS2 units above the body origin), matching every runtime's render
   camera, so shots land on the crosshair and the debug-runtime camera matches
   desktop (was rendering ~1 world unit low).
 - **Tooling**: debug-runtime controllable input over HTTP (`head.look` + hand
--- a/projects/flat-weapon-handling.md
+++ b/projects/flat-weapon-handling.md
@@ -106,8 +106,9 @@ Done:
   rotation — applying it over-rotates each gun by exactly its heading — so it is
   deliberately unused.
 - **Eye-height alignment** — the shot origin + viewmodel use the shared
-  `shock2vr::PLAYER_EYE_HEIGHT` (the original engine's `PLAYER_HEAD` sphere,
-  1.8 SS2 units above the body origin), matching every runtime's render
+  `shock2vr::PLAYER_EYE_HEIGHT` (the original game's camera placement: the
+  head sphere 1.8 SS2 units above the body origin plus the 0.8 eye offset =
+  2.6, i.e. 5.6 ft above the floor), matching every runtime's render
   camera, so shots land on the crosshair and the debug-runtime camera matches
   desktop (was rendering ~1 world unit low).
 - **Tooling**: debug-runtime controllable input over HTTP (`head.look` + hand
--- a/runtimes/debug_runtime/src/main.rs
+++ b/runtimes/debug_runtime/src/main.rs
@@ -1970,7 +1970,9 @@ fn capture_frame_snapshot(game: &Game, time: &Time, frame_counter: u64) -> Frame
                     .as_ref()
                     .map(|s| s.life_state.clone())
                     .unwrap_or_else(|| "alive".to_owned()),
-                camera_offset: [0.0, shock2vr::PLAYER_EYE_HEIGHT / SCALE_FACTOR, 0.0],
+                // The LIVE eye height, not the standing constant: automation
+                // aims from this, and a crouched player's camera is lower.
+                camera_offset: [0.0, game.player_eye_height() / SCALE_FACTOR, 0.0],
                 camera_rotation: [1.0, 0.0, 0.0, 0.0], // TODO: Get camera rotation
                 wielded_entity_id: state.as_ref().and_then(|s| s.wielded_entity_id),
                 right_hand_entity_id: state.as_ref().and_then(|s| s.right_hand_entity_id),
--- a/shock2vr/src/lib.rs
+++ b/shock2vr/src/lib.rs
@@ -52,15 +52,17 @@ pub use mission::visibility_engine::CullingInfo;
 /// this for [`PLAYER_CROUCH_EYE_HEIGHT`] - flat runtimes should use
 /// [`Game::player_eye_height`] rather than reading the constants directly.
 ///
-/// This is the original engine's `PLAYER_HEAD` sphere: Dark's first-person
-/// camera IS that submodel's location (`CameraUpdate` calls
-/// `PhysGetSubModLocation(objid, 0 /* PLAYER_HEAD */)`), and the sphere sits at
-/// `PLAYER_HEAD_POS = (PLAYER_HEIGHT / 2) - PLAYER_RADIUS` above the body
-/// origin. Deriving it from the collision profile keeps the eye *inside* the
-/// standing collider: an eye above the crown starts the crosshair ray outside
-/// any room whose ceiling the body itself clears, so the ray hits that ceiling
-/// from above and nothing in the room can be highlighted or frobbed (#795).
-pub const PLAYER_EYE_HEIGHT: f32 = physics::PLAYER_HEAD_POS;
+/// This follows the original game's camera placement: the viewpoint is
+/// anchored to the player's head sphere - `(PLAYER_HEIGHT / 2) -
+/// PLAYER_RADIUS` = 1.8 ft above the body origin - and then raised by the
+/// default eye offset ("eyeloc") of 0.8 ft, for a standing eye 2.6 ft above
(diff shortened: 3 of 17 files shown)


tommy-xr:
👀 The eye height ends up feeling incorrect with this fix in place for flat - it's too low.

tommy-xr:
**Follow-up: the camera was still 0.8 ft below the original's standing eye.**

The first pass put the camera at the head sphere center (1.8 ft above the body origin). That is where the original *locates* the head, but not where it puts the viewpoint: it then raises the camera by a separate default eye offset of **0.8 ft** (configurable as `"eyeloc"`; our data ships no override). So the branch as first pushed sat 0.8 ft low, which is noticeable in play.

`f234f32` adds a named `physics::PLAYER_EYE_OFFSET = 0.8` beside `PLAYER_HEAD_POS` and derives the standing eye from their sum:

- **2.6 SS2 ft above the capsule center = 5.6 ft above the floor**, matching the original's standing eye.
- Still 0.4 ft below the 3.0 ft capsule crown, so the #795 property that the camera is inside the collider by construction still holds.
- Crouch (`PLAYER_CROUCH_EYE_HEIGHT` = 1.2) is deliberately untouched — crouch parity is its own issue.
- SDK `PLAYER_EYE_HEIGHT_WORLD` 0.72 → **1.04** world units to match.

Verified at the taller eye: `cargo test -p shock2vr` 426 passed / 0 failed (including the seven-foot-room crosshair fixture — eye 5.6 ft under a 7 ft ceiling), SDK `npm test` 26 passed / 0 failed, and the `hydro2-corpse-frob` e2e still opens the corpse's loot MFD with Hydro Card B. Measured in hydro2 at the repro spot: body center `-0.756`, `camera_offset` `1.04`, camera y `0.284`, ceiling `0.80`.

Issue #801, opened 2026-08-03
[physics] Hydro2 QBR alcove collision pocket permanently strands the player (respawn can land in it)


tommy-xr:
## Summary

A collision pocket at x≈43.72 in the hydro2 Resurrection Station alcove immobilizes the player completely: bounded move reports blocked/0.0 in 8 directions AND raw thumbstick input (12 headings × 5 stick combos × 60 frames) moves the player 0.0. One metre away on the QBR pad, movement is normal.

**Critical aggravator: QBR reconstruction can drop the player INTO the pocket** — observed live: a death respawned the player at (43.78, 3.64, 31.46) and the run ended there. Any death on Deck 3 can permanently end a session.

## Evidence

- Place, not save: a freshly launched runtime teleported to the spot is equally stuck; save/load round-trips elsewhere are clean.
- Hashed local repro saves: `iter22_qbr_respawn_wedge_repro` (SHA-256 5ab601ae8b213c0448c6250211e525bbb2b7f78591518fad0cd224a954eff54d, player (43.72, 3.69, 31.32)) and `iter22_frontier_wedge_repro` (SHA-256 726918df798aaeed0872cb3edbb51bb8ad1efaabccf98986455bb1d0d3b50b21).

## Expected

The alcove floor around the QBR pad is walkable (it is in retail SS2), and QBR reconstruction places the player at a spot from which they can move.

High [physics] bug — campaign blocker (converts every Deck 3 death into a potential run-ender).

campaign: engineering-through-shodan · none · legacy · seed 1378752978

PR #685, opened 2026-07-28
fix(physics): stop dead creatures blocking players

tommy-xr:
Fixes #562
Fixes #716

### Exact Operations blocker

Campaign: `custom Operations→SHODAN · detailed · System Shock 2: 25th Anniversary Remaster assets · legacy pre-randomization ledger (no seed)`

On unmodified baseline `c07099af`, cold-load the locally retained save `campaign_ops_shodan_25th_iter03_corpse_block_20260808` (3,090,510 bytes; SHA-256 `6576c87093e72a865ce27b85e0077e4c96ad422eebef0dce4e77fdda261969e8`). The save is intentionally not published because it contains player/campaign state and retail-derived data.

The normally killed Protocol Droid (stable ops2 mission object `354`) lies transversely across the only ~2.5-unit Crew-lower catwalk near `(63.25, -9.50, 107.33)`. It remains visibly present and its empty SEARCH panel opens normally, but its one sleeping dynamic body reports `blocks_player: true` and `blocks_actor: true`.

From the saved player position `(64.25889, -9.95597, 108.02995)`:

- Collision-checked movement toward five center/edge lanes beyond the corpse advances only ~0.000–0.006 units and never crosses south of `z=106.4`.
- 180 fixed frames of normal production thumbstick input likewise stop at `z=108.02429`.
- Railings and walls close both edges, so there is no player-sized route around it.

The new exact-save E2E failed negative-first on main at `blocks_player true !== false`.

### Restarted Engineering campaign revalidation

Campaign: `engineering · verify-camera-alarms · 25th · seed 839293941`

A fresh `eng1.mis` run killed the authored OG-Pipes (stable mission objects `924` and `1670`) through ordinary `Damage`. On current `origin/main` (`ff09cde1`), the focused negative-first check retained the living `actor` collision group. The rebased fix leaves both corpses present and dead while reporting `blocks_player: false`, `blocks_actor: false`, and collision group `selectable`. The same result holds after loading the locally retained reproduction save `pt-eng-iter0-corpse-choke-repro` (SHA-256 `723672f4811b84bec72e1c3e935b8b28b39071e07f2c678fa7ee176146974884`). The save is intentionally not published because it contains player/campaign state and retail-derived data.

Adversarial review downgraded this occurrence from blocker to playable/non-blocking: an ordinary jump plus forward movement clears the narrow coolant-pipe choke on unmodified main. The retained collider is still a real defect covered by #562, but it did not gate campaign progress and this PR was not stacked onto the campaign branch.

[...]

PR #810, opened 2026-08-03
fix(physics): stop unsimulated debris wedging the player

### LINKED ISSUE #803: [physics] Kinematic debris bodies (Eggbit gibs, corpse rigid bodies) wedge the player
## Summary

Two related immobilization sources from campaign iteration 23 (hydro2 cold storage):

1. **Eggbit gibs**: annelid Floor Pods block cold-storage aisles and must be smashed; smashing one leaves permanent kinematic gib bodies that can wedge the player completely — immobilized at (36.15, −7.41, −32.28) with 5.2u of raycast-open floor to the west. Repro save `iter23_eggbit_wedge_repro` (SHA-256 cdd3e5aca4865d4d24d7f2f6a653fad24b06834c96ba96dd69c9fae37dc8a50c).
2. **Corpse rigid bodies pinch the player** in narrow aisles (Med) — same symptom family.

## Expected

Small debris (gibs) should not be solid to the player capsule (retail SS2 gibs are decorative), and depenetration should recover a capsule overlapped by kinematic debris rather than yielding 0-movement in all directions. Possibly the same depenetration robustness gap as #801.

Med-High [physics]. Both wedges were escaped only by save reload.

tommy-xr:
Fixes #803

> **Stacked on #805** (`fix/801-qbr-alcove-wedge`). It reuses that branch's
> `CollisionGroup::non_solid_to_player` helper; review this PR's own diff with
> `git diff origin/fix/801-qbr-alcove-wedge...HEAD`. Rebase onto `main` once #805 lands.

## Summary

- an object Dark simulates as a **movable** sphere, for which this engine can only stand in an **immovable kinematic box**, no longer stops the player capsule — smashed Floor Pod gibs stop sealing hydro2's cold-storage aisles
- the debris keeps its collider, its `entity` membership and everything built on it: raycasts, frobbing, projectiles and AI are untouched
- `/v1/physics/bodies` now reports `blocks_player`, because collision-group *membership* cannot show that a collider stopped filtering on the player

## Root cause

`Eggbit` (template `-4647`), the gib a smashed annelid Floor Pod flinderizes into, inherits `PropPhysType { SPHERE }` from `Monster Parts` (`-2182`) and carries **no `P$PhysDims` anywhere in its chain**, and no `P$FrobInfo`. It therefore lands in `entity_creator`'s non-frobbable branch on the #597/#767 model-bounds path, where a dimension-less object is given a **kinematic** collider — deliberately, because that fallback exists to make static geometry solid.

Measured live in `hydro2.mis` after smashing the cold-storage pod (mission object `1354`):

```
Eggbit  kinematic  pos [39.11, -8.34, -29.50]  groups ["entity"]  blocks_player true
Eggbit  kinematic  pos [40.75, -7.56, -29.03]  groups ["entity"]  blocks_player true
Eggbit  kinematic  pos [40.82, -7.70, -29.76]  groups ["entity"]  blocks_player true
```

collider half-extents ≈ 0.33 x 0.37 x 0.16 wu, against a 0.48-radius player capsule.

A kinematic body never yields, and rapier 0.31's `KinematicCharacterController` has **no depenetration pass** (`check_and_fix_penetrations` is a no-op), so a capsule overlapping one resolves to a zero-length move — in every direction, forever (the same substrate as #801). Because the pods block aisles the player is *required* to smash through, a gib that lands against them is a softlock. The campaign escaped it only by reloading, and gibs are not persisted, which is why the repro save loads free.

## Why this is faithful

`PhysType` is what makes an object physical in Dark, and **which** model it gets says what kind of object it is: `OBB` is the static/terrain/door model, `SPHERE` is the *simulated* one — a sphere hull under rigid-body dynamics that the player pushes out of the way. The shipped data separates cleanly along that line:

| class | example templates | `PhysType` | `Immobile` |
| --- | --- | --- | --- |
| terrain, structural, windows, doors | `Structural` −491, `Windows` −248, `Ladders` −783 | OBB | yes |
| gibs / debris / gore | `Monster Parts` −2182, `SecDroid Parts` −2888, `Chair Parts` −4820, `Misc Parts` −4832, `Gore` −197 | SPHERE | no |

Flinderize itself says the same thing: the link carries an **impulse**, and `slay_entity` calls `set_velocity` on the spawned flinder — it is meant to fly, bounce and be shoved. Our stand-in cannot move at all, which is not "more solid than retail", it is a different object: an invisible pillar where retail has a thing you kick aside.

Since this engine cannot push debris (the character controller applies no collision impulses), the closest faithful residue is that the debris does not stop the player. That is the same call #685 made for corpses.

## Per-class decisions

- **Gibs / loose debris (this PR)** — non-frobbable, `PhysType SPHERE`, no authored `P$PhysDims`, no `P$Immobile`, not climbable → keep the collider, drop `PLAYER` from its filter.
- **Corpses** — #803's second symptom is freshly-killed AI corpses (dynamic capsules, mass 0.33) drifting into the player and pinching them against geometry. That is exactly what **#685** ("stop dead creatures blocking players") fixes, with a `corpse()` group that keeps the body on `WORLD` and in selection rays but out of the player's way. Not duplicated here — #803 closes when both land.
- **Pre-placed corpse props** are unaffected by either: they carry `P$FrobInfo` and go down the creature branch.

## Scope

Measured, not argued: every mission in `Data/` was loaded and every body reporting `blocks_player: false` was resolved back to its template.

- **732 such bodies across all 23 missions. 729 of them are #805's typeless frobbables** (buttons, card slots, computers, keypads, audio logs, trainers, `RepBase`, pictures, `Debris #N` piles, ectoplasm).
- **This change adds exactly three**: `Cut Head` (template `-916`, `Gore` → `Decorative`, `PhysType SPHERE`, no dims, not immobile) — one in `many.mis`, two in `rec1.mis`. A severed-head prop; in retail you kick it around.
- Everything else it affects is **spawned at runtime**: gibs and droid/chair/misc parts from deaths and destruction.
- Nothing climbable, structural, door-like or `Immobile` is in the set; `!is_climbable` is asserted locally rather than relying on every ladder inheriting `PropImmobile`.
- Frob/selection/melee/projectile raycasts build `InteractionGroups::new(ALL, groups)`, so dropping `PLAYER` from a *collider's* filter leaves gibs shootable and selectable, and `ENTITY ↔ ENTITY` is untouched, so AI still collides with them.
- Objects authored explicitly mobile (`PropImmobile(false)` — `Chairs`, `Crystals`, `BrainBoss Pieces`) are deliberately **not** swept in: the branch uses the same property-presence test the dynamic path beside it already uses, and none of them is a reported wedge.

## Evidence

Both runs are the same probe on the same commit with the rule enabled / disabled; gib scatter is randomized per run, so each build carries its own pre-smash control.

**Grid probe** — 49 poses on a 0.4 lattice around the cold-storage pod, `/v1/player/move` best of 8 compass directions, before and after smashing the pod:

| | gibs' `blocks_player` | cells open before smash | cells open after smash | cells the gibs newly froze to 0 |
| before | true, true, true | 49 / 49 | 48 / 49 | **1** (1.000 → 0.024) |
| after | false, false, false | 49 / 49 | 49 / 49 | **0** |

**Standing on the debris** — smash every Floor Pod in hydro2's cold storage and south half, drop the player onto each resulting gib, best of 8 bounded moves:

| | poses sampled | fully frozen (0.000 in all 8 directions) |
| --- | --- | --- |
| before | 26 | **2** |
| after | 27 | **0** |

Every "after" pose reached the full 1.0 wu the probe asks for.

## Verification

- `RUSTFLAGS="-D warnings" cargo check --all-targets -p shock2vr -p desktop_runtime -p debug_runtime -p dark_query -p dark_viewer -p debug_command -p engine` — clean
- `cargo test -p shock2vr` — 427 passed
- `cargo fmt --all` — clean
- `entity_creator::tests::dimensionless_movable_debris_does_not_block_the_player` covers the shipped conditional (CI does not run e2e, so the behaviour change needs a unit test). **Negative-first verified**: with the new branch disabled it fails on `SPHERE (immobile: false) should not block the player`
- `entity_creator::tests::dimensionless_sphere_never_becomes_dynamic` now also asserts a climbable stays solid
- new `tools/shock2-sdk/test/debris-wedge.e2e.test.ts`; **negative-first verified** — with the rule disabled it fails with `gib 249 at [39.02, -7.49, -30.03] must not be solid to the player`
- full `SHOCK2_E2E=1 npm run test:e2e` — 197 tests, 194 passed, 0 failed, 3 skipped
- re-ran the movement- and physics-adjacent e2e on the final commit — 42/42: `debris-wedge`, `qbr-alcove-clearance` (#801), `model-bounds-colliders` (#597/#767), `climbing`, `jump`, `player-move`, `crouch-shaft`, `flinderize`, `monster-death`, `ragdoll-death`, `world-frob-pickup`, and all 23 mission loads

## Review

Both `/xreview` engines ran over the diff. Their agreed findings are fixed in the second commit: a climbable that reached the rule would have become unclimbable as well as walk-through (the climb probe queries as `PLAYER` too); `blocks_player` claimed more than it checked (player casts exclude sensors and require `ALL_COLLIDABLE` membership); and the new physics wedge test duplicated the frob-box wedge test beside it, so it was dropped — the mechanism is already covered there, and what this change actually decides is covered by the entity_creator test.

## Not fixed here

Gibs hang in mid-air where they spawn: `slay_entity` gives every flinder the impulse its Flinderize link authors, but a `kinematic_position_based` body ignores both that and gravity, so a smashed pod leaves its gibs floating inside the volume it occupied. Filed as #815 — it is a simulation/visual gap, not the softlock.


tommy-xr:
⚠️ Triage: should AI also not be blocked by these?

PR #852, opened 2026-08-08
fix(sdk): reap prior owned debug runtimes

### LINKED ISSUE #786: [sdk] GameServer should reap stale/orphaned debug runtimes

tommy-xr:
## Summary

- add opt-in `reapPrevious` support to `GameServer.launch`
- record each SDK launch in a checkout-scoped temporary lease keyed by requested port and random runtime instance ID
- gracefully stop, then terminate the detached process tree only while the live port still reports the exact leased instance ID
- preserve mismatched/unleased runtimes and retain the normal next-free-port behavior
- document the safe reap lifecycle for long-running playtest/play-through campaigns

This addresses the owner feedback on closed PR #793: reaping is not a process-name/default operation and cannot select another checkout's runtime. It is opt-in for a stable, caller-owned port.

## Reproduction

On current main, two separate owned SDK launches both requested port 17860. The first remained healthy at 17860 and the second remained healthy at 17861 (`bothAlive: true`), demonstrating that a dead automation owner would silently accumulate a runtime.

The new negative-first real-process test then failed on main because the replacement bound 8787 rather than reclaiming the orphan's 8786.

## Safety

- lease namespace includes the canonical checkout path
- a candidate must match requested port and the live `/v1/health` random instance ID
- graceful `/v1/shutdown` includes that exact ID
- force termination occurs only after an immediate second identity check
- mismatched IDs receive no shutdown request and are covered by unit and real-process tests
- no process-name scan or broad debug-runtime kill is used

## Verification

- `npm test`: 27 passed, 0 failed (190 e2e skips)
- focused real-process reap suite: 2 passed, 0 failed
  - exact prior orphan was reaped and the same port reused
  - mismatched lease/runtime stayed healthy while the replacement used the next port
- full SDK e2e: 213 passed, 1 failed, 3 conditional skips (217 total, 45m26s)
  - sole failure: existing `a loaded corpse does not replay its death` assertion
  - isolated branch run failed identically (30.60s)
  - untouched origin/main c07099a failed identically (46.55s)
  - all recorded corpse frames were Dead, sleeping, zero velocity, and zero position/pose error
- Prettier check: passed
- `git diff --check`: passed
- push hook `cargo fmt --all -- --check`: passed

No visual capture is applicable; this changes SDK process lifecycle only.

All issue-owned test runtimes were shut down or reaped and ports 8786-8789 were clear afterward. Pre-existing runtimes on unrelated ports were not targeted.

Fixes #786

### DIFF
--- a/.claude/skills/play-through/SKILL.md
+++ b/.claude/skills/play-through/SKILL.md
@@ -346,8 +346,11 @@ campaign just launches mission 0 with no frontier to load, which is harmless.
    and creates the ledger on the first iteration, keeps the roll after), then
    `show` → read the frontier + NEXT action (goal, tweak, `DARK_ASSET_PATH`).
 2. **Resume:** build the runtime from the **`fix_branch`** (so accrued fixes are
-   in), launch it **with the campaign's `DARK_ASSET_PATH`**, and `POST /v1/load
-   {file: frontier.save}` — or launch the first mission fresh at iteration 0.
+   in), launch it **with the campaign's `DARK_ASSET_PATH`** on a stable,
+   campaign-owned port via `GameServer.launch({ mission, port, reapPrevious:
+   true })`, and `POST /v1/load {file: frontier.save}` — or launch the first
+   mission fresh at iteration 0. The exact-ID reap prevents a dead iteration's
+   runtime from accumulating without touching another checkout's runtime.
 3. **Playtest** from here toward the goal (the `playtest` primitive), passing the
    campaign goal + the tweak instructions into the playtest prompt → `data.json`.
 4. **Review** the session (§2). Shallow/invalid → re-playtest with guidance.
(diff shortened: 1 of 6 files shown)


tommy-xr:
out of sync with other e2e fixes

PR #865, opened 2026-08-10
fix(physics): honor authored frobbable shapes

### LINKED ISSUE #864: fix(physics): honor authored sphere colliders on frobbable props
## Campaign blocker

The Command lower bridge-shield route is sealed at the only shipped WALK | SMALL_CREATURE link from path cell 3799 to 3798. The campaign reached the supported approach near `(-287.500, -7.796, 86.3136)` after using real button 840 / door 839 and descending the ladder.

Exact local frontier (do not upload):

- file: `campaign_cmd25_i24_command1_shield1_off_hp7_20260809.sav`
- size: 13,763,739 bytes
- SHA-256: `70f872731adb360dab6e9e1f22ae6584bf324da91e32e72b0750b3de99f2dbc9`
- configuration: Command 1, bridge shield off, HP 7

## Root cause

Mission object 1520 (`Swarmer Floor Pod`) sits at `(-288.22046, -7.6019087, 87.673706)`. Its authoritative properties are:

- `P$PhysType`: SPHERE
- `P$PhysDims.radius0`: 0.39198092
- `P$PhysDims.radius1`: 0
- `P$FrobInfo.world_action`: SCRIPT

The frobbable entity-creation branch ignores authored PhysType/PhysDims unless the object is a creature or MOVE-frobbable. It instead creates a kinematic model-bounds cuboid. The live collider AABB is x `[-289.22046, -287.22046]`, y `[-8.385871, -6.817947]`, z `[86.673706, 88.673706]`: a 2.0 x 1.568 x 2.0 box that fills the route. Destroying object 1520 makes ordinary crouched locomotion advance from z 86.31 to 91.63, proving the path portal geometry itself is open.

This initially looked like #765 / PR #768, but expanding mantle compression would be unsafe: it only succeeds by phasing the compressed player through this real parented entity. The existing first-candidate abort correctly preserves solid props.

## Expected

Every supported frobbable object with explicit PhysType and PhysDims should instantiate its authored collision shape, offset, dynamics, and collision group. Model bounds remain the selection-only fallback for typeless frobbable objects. No Command or Floor Pod special case.

## Acceptance

- negative-first unit proves frobbable SPHERE + PhysDims selects the authored sphere rather than model bounds
- live Command regression proves object 1520 has its authored sphere-sized dynamic body and the real production route is traversable with the pod present
- ordinary typeless frob selection, MOVE-frobbable, OBB, creature, launched projectile, and dimensionless fallback behavior do not regress
- core warning-clean checks, SDK units, focused/full E2E pass
- visible before/after evidence shows the same route sequence

tommy-xr:
Fixes #864

## Summary

- preserve explicit `P$PhysType` / `P$PhysDims` geometry on frobbable props instead of replacing it with render-model selection bounds
- keep authored movable spheres and MOVE-frobbable props dynamic; support faithful immobile sphere/OBB geometry through a generic kinematic-shape path
- retain model-sized frob/projectile targeting as a zero-mass, contact-free secondary collider while the authored shape alone controls physical contacts
- retain the typeless selection fallback, creature and launched-projectile paths, dimensionless fallback, and character collision filtering
- add negative-first unit coverage for exact sphere radius/offset/mobility plus a live Pod-present Command regression through real button 840 / door 839

## Campaign evidence

Command1 mission object 1520 (`Swarmer Floor Pod`) authors `P$PhysType=SPHERE` and `P$PhysDims.radius0=0.39198092` at `(-288.22046, -7.6019087, 87.673706)`. Because it inherits SCRIPT frobbing, the old frob branch ignored those properties and built a kinematic 2.0 × 1.568 × 2.0 render-model box. That box seals the only WALK | SMALL_CREATURE link from path cell 3799 to 3798.

The live regression leaves Pod1520 alive and solid, opens the real route door with production aim+squeeze, stages the campaign-supported lower pose, and walks through the authored opening with production look+locomotion.

Exact local frontier (not uploaded): `campaign_cmd25_i24_command1_shield1_off_hp7_20260809.sav`, 13,763,739 bytes, SHA-256 `70f872731adb360dab6e9e1f22ae6584bf324da91e32e72b0750b3de99f2dbc9`.

## Safety boundary

This supersedes the initial #765/#768 triage for the Command symptom. Expanding mantle compression was rejected because it only appeared to pass by phasing through a real parented entity and regressed the existing low-wall safety contract. PR #768 remains untouched for its original Hydro scope.

Authored physics and render selection stay separate: the first collider preserves exact physical radius/offset/mobility; the secondary model-bounds collider participates in interaction rays but no physical contact pair. Existing StdDoor targeting, ordinary and hacked replicator pickups, Earth Psi pickups, and Command crossing all pass together.

## Verification

- negative-first focused unit: red (`kinematic` model box), then green (`dynamic` authored sphere with exact radius and offset)
- negative-first live Command scenario: red (z 86.3136 → 86.3136), then green beyond z 89 with Pod1520 present, dynamic, and solid
- focused compatibility matrix: **5/5 passed** — StdDoor, Earth Psi, replicator economy, hacked replicator, Command Pod route
- `cargo test -p shock2vr`: **511 passed**, plus doctests
- `RUSTFLAGS='-Dwarnings' cargo check -p shock2vr -p desktop_runtime -p debug_runtime`: passed
- SDK unit suite: **26 passed**, 189 expected opt-in skips
- exhaustive serial SDK suite: **211 passed**, 3 separately gated SHODAN finale skips, 1 pre-existing loaded-corpse collision-group failure over 47m47s; the sole failure reproduces unchanged on clean `origin/main` and is already addressed by open PR #685
- GitHub Actions: format, Linux, macOS, Ubuntu, Windows, and Android builds all green

## Visual evidence

Command Floor Pod before/after demo

<sub>Identical headless fixed-step sequence on clean `main` and this PR: the inherited render-box collider pins production locomotion at z=86.314; the authored dynamic sphere lets the player pass the still-live, solid Pod1520 and settle on the supported floor at z=89.360. Captured through `/v1/step` + `/v1/screenshot`.</sub>

### Before → after

Command Floor Pod blocked before and supported beyond after

### DIFF
--- a/shock2vr/src/creature/hit_boxes.rs
+++ b/shock2vr/src/creature/hit_boxes.rs
@@ -268,7 +268,7 @@ impl HitBoxManager {
                     let rotation = get_rotation_from_matrix(&joint_xform);
                     // If there is not a physics entity yet, create one
                     if !id_to_physics.contains_key(hit_box_entry) {
-                        let physics_handle = physics.add_kinematic_shape(
+                        let physics_handle = physics.add_kinematic_shared_shape(
                             *hit_box_entry,
                             pos,
                             rotation,
--- a/shock2vr/src/mission/entity_creator.rs
+++ b/shock2vr/src/mission/entity_creator.rs
@@ -1094,34 +1094,39 @@ fn create_physics_representation_with_options(
         }
 
         let qrotation = pos.rotation;
-
-        let _is_sensor = true;
-
-        // let dimensions = v_phys_dimensions
-        //     .get(id)
-        //     .map(|d| d.size)
-        //     .ok()
-        //     .or_else(|| {
-        //         id_to_model.get(&id).and_then(|model| {
-        //             model.bounding_box().map(|bbox| bbox.max - bbox.min)
-        //         })
-        //     })
-        //     .unwrap_or(default_size);
-
-        // TODO: Add dynamic rigid body for some items?
-        // let shape = if let (Ok(phys_type), Ok(dimensions)) =
-        //     (v_phys_type.get(entity_id), v_phys_dimensions.get(entity_id))
-        // {
-        //     match phys_type.phys_type {
-        //         PhysicsModelType::OrientedBoundingBox => PhysicsShape::Cuboid(abs_dimensions),
-        //         PhysicsModelType::Sphere => {
-        //             PhysicsShape::Sphere(dimensions.radius0.abs().max(dimensions.radius1.abs()))
-        //         }
-        //         _ => panic!("unhandled physics type: {:?}", phys_type),
-        //     }
-        // } else {
-        //     PhysicsShape::Cuboid(abs_dimensions)
-        // };
+        let authored_physics = match (
+            v_phys_type.get(entity_id).ok(),
+            v_phys_dimensions.get(entity_id).ok(),
+        ) {
+            (Some(phys_type), Some(dimensions)) => {
+                let shape = match phys_type.phys_type {
+                    PhysicsModelType::ORIENTED_BOUNDING_BOX => Some((
+                        PhysicsShape::Cuboid(vec3(
+                            dimensions.size.x.abs(),
+                            dimensions.size.y.abs(),
+                            dimensions.size.z.abs(),
+                        )),
+                        false,
+                    )),
+                    PhysicsModelType::SPHERE => Some((
+                        PhysicsShape::Sphere(
+                            dimensions.radius0.abs().max(dimensions.radius1.abs()),
+                        ),
+                        true,
+                    )),
+                    _ => None,
+                };
+                shape.map(|(shape, is_sphere)| (shape, dimensions.offset0, is_sphere))
+            }
+            _ => None,
+        };
+        let mut frob_group = CollisionGroup::entity();
+        if v_hud_select
+            .get(entity_id)
+            .is_ok_and(|hud_select| hud_select.0)
+        {
+            frob_group = CollisionGroup::selectable();
+        }
 
         let rigid_body_handle;
         // Is a creature - so we need special handling for their bounding box
@@ -1146,6 +1151,37 @@ fn create_physics_representation_with_options(
                 dynamics_options,
             );
             physics.set_enabled_rotations(entity_id, false, false, false);
+        } else if let Some((shape, offset, is_sphere)) = authored_physics {
+            // FrobInfo controls interaction, not collision geometry. When a
+            // visible prop also carries an explicit physical model, preserve
+            // that authored shape instead of replacing it with the render
+            // bounds used only for selection. SPHERE is Dark's simulated
+            // moving model unless the object is explicitly immobile; MOVE
+            // frobs likewise remain dynamic for either supported shape.
+            if frob_info.world_action.contains(FrobFlag::MOVE)
+                || (is_sphere && !launched_object_is_immobile)
+            {
+                rigid_body_handle = physics.add_dynamic(
+                    entity_id,
+                    pos.position,
+                    qrotation,
+                    offset,
+                    shape,
+                    frob_group,
+                    false,
+                    dynamics_options,
+                );
+            } else {
+                rigid_body_handle = physics.add_kinematic_shape(
+                    entity_id,
+                    pos.position,
+                    qrotation,
+                    offset,
+                    shape,
+                    frob_group,
+                    false,
+                );
+            }
         } else if frob_info.world_action.contains(FrobFlag::MOVE) {
             let shape = PhysicsShape::Cuboid(abs_dimensions * 1.0);
             rigid_body_handle = physics.add_dynamic(
@@ -1161,13 +1197,6 @@ fn create_physics_representation_with_options(
                 dynamics_options,
             );
         } else {
-            let mut group = CollisionGroup::entity();
-            if let Ok(hud_select) = v_hud_select.get(entity_id) {
-                // HACK: Remove pick bias around fluidics computer
-                if hud_select.0 {
-                    group = CollisionGroup::selectable();
-                }
-            }
             // This collider is the *model bounding box*, built only so the
             // object can be frobbed and raycast - it is not an authored
             // collision volume. Dark makes an object physical by giving it a
@@ -1178,7 +1207,7 @@ fn create_physics_representation_with_options(
             // 2.2 x 4.0 x 2.5 box the player must stand inside - and wedges
             // the capsule against its faces with no way out (#801).
             if v_phys_type.get(entity_id).is_err() {
-                group = group.non_solid_to_characters();
+                frob_group = frob_group.non_solid_to_characters();
             }
             rigid_body_handle = physics.add_kinematic(
                 entity_id,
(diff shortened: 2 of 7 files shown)


tommy-xr:
Integration testing of the five item-outline-pass PRs (#742, #865, #878, #1150, #1152 — branch `integration/item-outline-pass`, full SDK e2e suite) found three e2e regressions that reproduce on this branch **alone** (clean on base `f1e2a899` and on the other four branches):

- `medsci-sight-occlusion` — hidden monster target no longer moves when the occluding door opens
- `melee-impact-sound` — a weapon resting on a surface emits 3 spurious `hmetmet3/4` collision sounds (`3 !== 0`)
- `vr-cyber-interface` — world stops simulating while the cyber interface is open (moved 0.042)

All three plausibly share a root cause: the authored-shape kinematic bodies / interaction-only secondary collider changing contact generation and AI-visible geometry. Recommend holding this PR until they're fixed. Also: if #1152 lands first, renaming its `add_kinematic_shape` to `add_kinematic_shared_shape` (one caller in `hit_boxes.rs`) lets this PR land without the `physics/mod.rs` conflict; otherwise see the three-layer resolution on `integration/item-outline-pass` (caece22f).

tommy-xr:
Rebased onto current `main` and fixed the three integration regressions. Force-pushed.

## Rebase

`main` now carries #1152, whose `physics/mod.rs` added an `add_kinematic_shape` taking a `SharedShape` — a name collision with this PR's `PhysicsShape`-taking one. Resolved in three layers, so neither caller had to change shape:

- `add_kinematic(…, size, …)` → delegates with `PhysicsShape::Cuboid(size)` (no double-sanitize)
- `add_kinematic_shape(…, shape: PhysicsShape, …)` → this PR's signature verbatim; converts Capsule/Cuboid/Sphere to a `SharedShape` (capsule assert kept, cuboid size / sphere radius sanitized) and delegates down
- `add_kinematic_shared_shape(…, shape: SharedShape, offset, …)` → #1152's function, body verbatim, renamed; its one caller in `creature/hit_boxes.rs` updated

`git range-diff` confirms nothing else changed across the rebase.

## Root causes

**1 — `melee-impact-sound` (3 spurious `hmetmet3/4`).** Not the held Wrench: a *second* Wrench across the room fell and clattered onto the deck. Reading the authored shape for every frobbable object, and simulating any authored sphere, put loose items on point-sized colliders. An item's authored sphere really is a marble (the Wrench's is 0.046 wu across); the original engine rests it on a shelf through its own support/sleep bookkeeping, which this port does not model, so in rapier it starts unsupported and sinks through whatever it was authored standing on. Measured in command2 on the branch vs clean `main`: 8 named props in free fall where `main` had none, and the Korenchin Log and Modify Soft V2 never stop — velocity climbing 10.3 → 27.5 wu/s as they leave the level. Turning non-MOVE fixtures dynamic did the same to wall fixtures.

Fix: narrow the authored-shape path to what #864 actually needs. A **fixture** (non-MOVE frob) keeps its authored shape, kinematic, exactly where it was authored, instead of the oversized render-model box. **Carryable MOVE frobs are untouched** and keep the dynamic model-bounds body they have always had. After the fix the command2 falling-body set matches `main` exactly.

The Command Floor Pod is therefore kinematic rather than dynamic. That is all #864 needed — the bug was the 2.0 × 1.568 × 2.0 render box sealing the 3799 → 3798 link, not the pod's mobility; the pod is still present, solid and route-opening, and the live crossing regression is unchanged and green.

**2 — `vr-cyber-interface` (world "frozen", moved 0.042).** Same cause. The test identifies the spawned debug item as the last dynamic body; with hundreds of props newly dynamic it picked a different body. Fixed by 1, no test change.

**3 — `medsci-sight-occlusion` (hidden monster never resumes).** Not a behavior regression — the AI resumes correctly on this branch. Instrumented the whole scenario against clean `main` and this branch: after the door opens both republish `AILastKnown`, both go `Chase` → `MeleeAttack`, both close on the player.

The final assertion was `notEqual(AILastKnown, hiddenLastKnown)` while the player reappeared at the *same* control point it was last seen from — where a republished last-known and a stale one read identically. On `main` it passed only because the 5-frame poll happened to sample the player mid-fall (`y -0.30` vs `-0.36`); this PR shifts the settle by ~2 cm and the sample lands the other side of that race.

Amended it to measure what it claims: reacquire a couple of units away and assert the published coordinate tracks the player. Verified the amended test passes on unmodified `origin/main` as well.

## Also fixed

`P$PhysType` and `P$PhysDims` are inherited independently, so a fixture can declare a shape its dimensions never describe — medsci2 ships one whose authored OBB `size` is all zeros. Read literally that became the 1 cm cube the collider sanitizer floors it at, i.e. a solid fixture the player walks through. A degenerate authored shape now falls back to the model bounds, the same rule a fixture with no `P$PhysDims` already followed. Negative-first unit test included (red without the guard).

## Verification

- `cargo fmt --all`; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`: clean
- `cargo test -p shock2vr`: **1093 passed**, 0 failed
- SDK unit suite: **32 passed**, 0 failed
- The three regressions, each red before / green after on this branch: `medsci-sight-occlusion` 2/2, `melee-impact-sound` 1/1, `vr-cyber-interface` 2/2
- Plus `frobbable-physics` 1/1, `earth-psionic-training` 3/3, `vr-melee-contact` 3/3, `flat-melee-animation-event` 1/1
- Full serial `npm run test:e2e` was still running when this was written; it will need a look before merge

## Review follow-ups (not applied here — they are on the pre-existing diff, and each wants its own decision)

An adversarial review pass over the whole branch raised these; none block the regressions above, and I left them alone rather than widen the change:

- **Authored sphere radii are half the authored value.** The importer divides `radius0/radius1` by 2, and `live_creature_shape` compensates with a `* 2.0` that its comment scopes deliberately to live creatures ("loose props keep their existing parser/physics semantics"). The fixture path inherits the un-doubled value. Worth deciding explicitly, but note doubling the Floor Pod's radius may re-seal the very route #864 opens — so it needs its own PR with the Command crossing re-run.
- **`USE_AMMO` frobs.** `internal_frob_move` treats carryable as `!SCRIPT && (MOVE || USE_AMMO)`; the fixture predicate here only excludes `MOVE`. Mirroring the existing predicate would be more consistent (behaviour is unchanged from `main` either way, since a `USE_AMMO`-only frob took the kinematic branch there too).
- **Degeneracy guard vs sanitizer floor.** The new guard rejects exactly zero; the sanitizers floor at 0.01 (size) / 0.005 (radius), so a value in between still reaches the floor. Comparing against those minima would close the gap.
- **`frob_group` vs `entity()`.** The authored collider hardcodes `CollisionGroup::entity()` while the interaction cuboid gets `frob_group`, so a `PropHUDSelect` fixture ends up with one `ENTITY` and one `SELECTABLE` collider — partially undoing the pick-bias hack the same function computes just above. On `main` the single model-bounds collider used `frob_group`.
- **`set_collision_group` / `resize_kinematic_cuboid` rewrite every collider on a body.** No current caller reaches a fixture, but the first one that does would make the ray-only interaction cuboid solid.
- **Simplicity:** the authored `PhysType`/`PhysDims` → `PhysicsShape` derivation now exists three times in `entity_creator.rs` (launched projectile, fixture, non-frobbable), each subtly different; and `secondary_cuboid_full_size` is `cuboid_full_size` with a `skip(1)`.

Note the PR description above still describes the pod as a dynamic simulated body — that line is stale after this change.

PR #867, opened 2026-08-10
feat(debug): expose entity containment details

### LINKED ISSUE #776: feat(debug): expose HasRefs and incoming Contains links in /v1/entities/:id
## Summary

The debug runtime's entity detail endpoint (`GET /v1/entities/:id`) omits `P$HasRefs` and incoming `Contains` links. Container loot (no world presence by design) is therefore indistinguishable from a genuinely world-placed item that failed to get a render/collider.

## Why it matters

This gap caused two successive wrong root-cause guesses in #569: world-suppressed container loot (authored `Contains` link + `HasRefs=false`) looked identical to world-placed ammo in the endpoint output, so the report concluded "world ammo has no colliders" when the 0-body items were all inside lockers/corpses/crates.

## Suggested change

In the entity detail response (`runtimes/debug_runtime/src/main.rs`), include:

- `has_refs` (from `P$HasRefs`, defaulting true when absent)
- incoming `Contains` links (container entity id + name), mirroring how `cargo dq entities <mis> <id>` shows links in both directions

Playtest/triage automation can then classify "no physics body" as expected (contained) vs. a real bug (world-placed) without dropping to `cargo dq` against mission-file IDs.

tommy-xr:
## Summary

- expose top-level `has_refs` from live `PropHasRefs`, preserving the Dark default of `true` when absent
- populate entity-detail `incoming_links` with live incoming `Contains` edges so contained loot identifies its source container
- type the new field in the SDK and cover contained/world-placed controls with a real MedSci scenario

## Reproduction

On current `origin/main` (`607c1ce`), `MissionCore::entity_detail` hardcoded `incoming_links` to an empty vector and never read `PropHasRefs`. The negative-first MedSci scenario failed with the corpse-contained Psi Amp reporting `has_refs` as `undefined` instead of `false`; the endpoint therefore could not distinguish contained loot from a broken world-placed item.

## Verification

- `cargo fmt --all -- --check` — pass
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` — pass
- `cargo test -p shock2vr` — 514 passed
- `npm test` in `tools/shock2-sdk` — 26 passed, 190 opt-in skipped
- focused `entity-containment-detail.e2e.test.ts` — pass standalone before commit, in the full suite, and after final rebuild/rebase
- full `npm run test:e2e` — 210 passed, 3 failed, 3 skipped; all 23 mission-load tests and all containment/detail consumers passed. One failure passed on isolated rerun; the two persistent unrelated failures reproduced identically in a clean detached `origin/main` worktree (`a loaded corpse does not replay its death`, `losing sight of the player triggers a search of the last-known position`).

Fixes #776


tommy-xr:
this functionality is already on main

PR #894, opened 2026-08-11
fix(gameplay): respawn ecology spawns named by a mission object

tommy-xr:
## Problem

In the original game a killed training droid on Earth is put back on its stand by an
ecology a few seconds later. In our port it never came back: every ecology tick logged

```
WARN TrapSpawn could not resolve archetype DopeyDroid
```

(now observable since #888 made the droids actually die).

Two bugs, both on the path from "the ecology wants a droid" to "one droid stands there".

## Bug 1 - the spawn archetype never resolved

`MissionCore::spawn_ecology_entity` resolved `PropSpawn`'s authored archetype name
through `create_entity_by_template_name` → `template_name_to_template_id`, and
`create_template_name_map` (`shock2vr/src/mission/mission_core.rs:6027`) builds that map
from the **gamesys only**.

earth.mis's spawn markers name `"DopeyDroid"`, which is not a gamesys template — it is a
concrete **mission object** (earth.mis object 597, a `Training Droid` + `Docile` + `Silent`
archetype parked off-map) that retail's `TrapSpawn` looks up by object name:

```
Ecology 599 (P$EcoType 41, period 8s, min/max 1)
    --SwitchLink--> Spawn 598 (P$EcoType 41, SELF_MARKER, object_names ["DopeyDroid", ...])
Droid 547 (P$EcoType 41) stands on that marker.
```

**Fix**: index this mission's own objects that author a **unique** `P$SymName`
(`create_mission_object_name_map`) and use it as a fallback **after** the gamesys lookup,
**only in the ecology spawn path**:

- Scoped to `spawn_ecology_entity`, not to the shared `create_entity_by_template_name`, so
  no other by-name creation path (SpawnSFX, tweq emitters, the replicator) changes
  behavior.
- Own `P$SymName` only — an object that merely inherits its archetype's name, or is named
  only by the mission's `OBJ_MAP`, is not a distinct archetype.
- Ambiguous names are dropped, not resolved arbitrarily, mirroring the existing
  `unique_gamesys_template_names`.
- Resolving to a mission object is logged, since cloning a concrete object also clones its
  authored links.

## Bug 2 - the ecology never counted the droid it spawned

With bug 1 fixed the droid came back, and then kept coming back: **one more droid per
stand every 8 s, forever**, past the authored `max_count` of 1.

`TriggerEcology::population` (`shock2vr/src/scripts/trigger_ecology.rs:41`) decided
"is this a physical population member (a creature) or one of my own markers?" from
`entity_class_template_id`, i.e. `RuntimePropCanonicalTemplateId` — the **nearest negative
ancestor**. For an object whose metaprops sort ahead of its archetype that ancestor is a
**metaprop**: both earth's `DopeyDroid` (597) and the third stand's authored droid (567)
resolve to `Docile` (-1073), which does not descend from `Physical` (-11). So those
creatures were invisible to the census and their ecology respawned one every period.
Runtime instrumentation of the census:

```
ECODBG entity EId(49.0)  eco 41 class Some(-4015) physical true    <- authored droid 547
ECODBG entity EId(750.0) eco 42 class Some(-1073) physical false   <- a spawned droid
```

**Fix**: judge membership from the entity's **concrete** `PropTemplateId` ancestry, which
does descend from `Physical`, falling back to the canonical class id only when this
mission's hierarchy does not know the concrete template (an entity carried in from another
mission, whose positive id belongs to that mission's id space). Markers stay excluded:
`Spawn` (-976) and `Ecology` (-975) descend from `Traps` → `Object`, never `Physical`.

## Verification

Headless, deterministic (`/v1/step` at fixed 60 Hz), killing the basic-training droid at
t=0 and stepping five 8 s periods — before, one droid accumulated per stand per period
(4 → 13 droids); now:

```
t=8s   669 DopeyDroid 597 [88.0, 24.2, 187.0]   <- respawned on the stand
t=40s  669 DopeyDroid 597 [87.8, 24.1, 187.1]   <- still exactly one, three periods later
```

Before/after capture of the same sequence — the boxed stand stays empty on `main`, and
holds a fresh protocol droid with the fix:

before/after

| before (main) | after (fix) |
| --- | --- |
| before | after |

(The droid in the foreground is present in both runs and is unrelated to the change.)

## Tests

Both fixes are **negative-tested first** — each test fails without its fix:

- `tools/shock2-sdk/test/earth-training-droid-respawn.e2e.test.ts` — kill the droid, step
  past two periods, assert exactly one live protocol droid on the stand (fails on `main`:
  `0 !== 1`), then step three more periods and assert the level still holds exactly two
  template-597 droids: the parked archetype and the one respawn (fails without the census
  fix: 11 droids).
- `counts_a_spawn_whose_canonical_class_is_a_metaprop` — the census unit regression.
- `mission_object_name_map_tests` — positive-ids-only, own-name-only, unique-name-only.

Validation:

- `cargo test -p shock2vr` — 544 passed.
- `SHOCK2_E2E=1`: all 23 missions load, plus `death-links`, `security-ecology`,
  `shodan-avatar-ecology`, `monster-death`, `spawn-drop-to-floor`,
  `earth-training-narration`, `earth-psionic-training` and the new test — 30 passed /
  0 failed.
- Full e2e suite (on the first fix): 229 tests, 221 pass, 2 fail, 6 skipped — the two
  failures are in the known pre-existing/flaky set; every test covering the systems this
  touches passes.

### DIFF
--- a/shock2vr/src/mission/mission_core.rs
+++ b/shock2vr/src/mission/mission_core.rs
@@ -495,6 +495,10 @@ pub struct MissionCore {
     #[allow(dead_code)]
     pub template_to_entity_id: HashMap<i32, WrappedEntityId>,
     pub template_name_to_template_id: HashMap<String, EntityMetadata>,
+    /// This mission's own objects (positive ids) that author a unique
+    /// `P$SymName`, indexed by lowercased name - the fallback an ecology spawn
+    /// uses when the gamesys has no archetype under the authored name.
+    mission_object_name_to_id: HashMap<String, i32>,
     pub obj_map: HashMap<i32, String>,
     pub world: World,
     pub player_handle: PlayerHandle,
@@ -696,6 +700,7 @@ impl MissionCore {
         // This is important for creating entities based on template name
         let (template_name_to_template_id, unique_gamesys_template_names) =
             create_template_name_map(game_entity_info);
+        let mission_object_name_to_id = create_mission_object_name_map(&entity_info_rc);
 
         world.add_unique(GlobalEntityMetadata(template_name_to_template_id.clone()));
         world.add_unique(Time::default());
@@ -1253,6 +1258,7 @@ impl MissionCore {
             id_to_bitmap,
             id_to_particle_system: HashMap::new(),
             template_name_to_template_id,
+            mission_object_name_to_id,
             scene_objects: scene,
             animated_lightmaps,
             physics,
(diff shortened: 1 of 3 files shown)


tommy-xr:
⚠️ There is a bug that dopey droids will continue to spawn, not respecting the ecology max limit

Issue #797, opened 2026-08-02
[gameplay] Collected audio-log disc remains in the world and stays frobbable


tommy-xr:
After collecting a Deck 3 audio log (Korenchkin 're: Miracles') — which correctly plays and enters the log MFD — the disc object remains in the world and can still be frobbed (no duplication observed, but the pickup should consume the world object as in retail SS2).

Low [gameplay].

campaign: engineering-through-shodan · none · legacy · seed 1378752978

PR #723, opened 2026-07-29
fix(sdk): aim from the live crouched camera

tommy-xr:
Fixes #721

## What changed

- report the debug runtime's actual live camera offset in `/v1/info`, including the crouched eye height
- make SDK `aimAt` and `lookAtWorldPoint` default to that reported camera height while retaining an explicit world-unit override
- add a negative-first SHODAN scenario that discovers Delacroix log 4 by stable `template_id === 290`, stages the authentic hidden platform, and collects it through the production flat camera + squeeze path
- cover crouched camera-origin selection in the SDK unit suite

## Root cause

The log, its `LogDiscScript`, and its selectable collider were healthy. The debug runtime reported a standing `camera_offset` even while the production controller was crouched, and the SDK independently defaulted to the standing 1.6-world-unit eye height. The campaign's crouched interaction therefore verified and aimed one ray while `FlatInteraction` frobbed from another.

The original `eyeHeight: 1.2` reproduction also mixed coordinate spaces: `PLAYER_CROUCH_EYE_HEIGHT` is 1.2 SS2 feet, which is 0.48 runtime world units after `SCALE_FACTOR`. Pointing from the real 0.48-world-unit crouched eye collected the log before any script or geometry change, confirming the mismatch.

## Negative-first evidence

Before this change, the new scenario reached the visible/selectable log surface but failed:

```text
AssertionError: production squeeze should collect Delacroix deck 9/log 4
```

After the runtime and SDK use the same live camera offset:

```text
✔ crouched flat interaction collects SHODAN Delacroix log 4
```

## Validation

- `npm test` — 27 passed, 154 skipped
- focused `shodan-log4-interaction.e2e.test.ts` — passed
- `cargo fmt --all -- --check`
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`
- full SDK E2E — running; results will be added

## Visual verification

Crouched SHODAN log 4 interaction before and after

<sub>Matched fresh `shodan.mis` runs discover template 290 dynamically, stage the same crouched hidden platform, call `aimAt` without an eye-height override, then pulse the production squeeze input. Before, the standing aim origin points the real crouched camera into darkness and collects nothing. After, the live 0.48-world-unit camera origin centers/highlights the disc, collects deck 9/log 4, and opens its reader. Captured headlessly via `/v1/step` + `/v1/screenshot` with deterministic fixed-timestep stepping.</sub>

### Aim — before → after

[...]

PR #899, opened 2026-08-11
test(sdk): cover crouched SHODAN log interaction

### LINKED ISSUE #721: bug(gameplay): SHODAN Delacroix log 4 cannot be frobbed
## Campaign configuration

- Play-through seed: `1546344805`
- Scenario: SHODAN finale
- Tweak: Loot everything
- Assets: System Shock 2 25th Anniversary (`DARK_ASSET_PATH=/Users/bryan/ss2-25th`)
- Campaign stack includes the accepted #708 revision-3 descent and the other open SHODAN fixes.

## Reproduction

1. Load `shodan.mis` and traverse the authored central-tower route to the stacked upper floor.
2. While crouched, use the production jump channel from the interior of that floor. The Dark sparse-body transition lands stably on the hidden `y=-28` platform beneath it.
3. Discover the remaining audio log by stable template identity:

   ```ts
   const log = (await game.entities.list({ filter: "Audio Log", limit: 80 }))
     .entities.find((entity) => entity.template_id === 290);
   ```

   Mission object/template `290` has `PropLog { deck: 9, log: 4 }`.
4. Move into normal frob range and aim the production flat interaction ray at its visible selectable surface.
5. Pulse the normal squeeze interaction:

   ```ts
   await game.player.aimAt(log, { hitbox: "surface", visibility: "required", eyeHeight: 1.2 });
   await game.input.set("right_hand.squeeze", 1);
   await game.step({ frames: 1 });
   await game.input.set("right_hand.squeeze", 0);
   await game.step({ frames: 5 });
   ```

## Actual

- Player is stable at approximately `{31.9264, -27.3150, 28.5662}`.
- Log is at `{32.0, -26.2, 28.4}`.
- Visibility ray reaches the selectable log surface with no blocker; measured target distance is `0.2006`.
- Physics reports the log as an enabled kinematic body in the `selectable` collision group.
- Repeated production squeeze pulses do nothing: the entity remains alive and `info.player.collected_logs` remains `[9/1, 9/2, 9/3]`.

## Expected

The normal flat crosshair + squeeze interaction should run `LogDiscScript`, remove/collect the world log, and append deck 9/log 4, as it does for the other SHODAN Delacroix logs (including the existing durable log-2 E2E).

## Impact

The campaign can continue by skipping the log, so this is not yet a progression blocker, but the authored collectible cannot be looted and the active "Loot everything" play-through requirement cannot be completed faithfully.

Please reproduce with a dynamically discovered entity (never a hardcoded runtime ID), add a negative-first SHODAN E2E, and keep the fix focused on the real interaction/geometry cause.

tommy-xr:
Fixes #721

## Summary

- restore the focused SHODAN regression that was left behind when #723 was superseded by #800
- discover Delacroix log 4 by stable `template_id === 290`, stage the authentic crouched hidden platform, and use normal flat `aimAt` + squeeze interaction
- assert the debug runtime reports the live 0.48-world-unit crouched camera, the production ray selects the disc, deck 9/log 4 enters `collected_logs`, and the reader opens

## Root cause and existing fix

The log, collider, `PropLog`, and `LogDiscScript` are healthy. The original reproduction aimed through debug automation from a stale 1.6-world-unit standing eye while the production crouched interaction ray originated at 0.48 world units. PR #800 merged the runtime/SDK alignment in `0e36fb11`: `/v1/info` reports `game.player_eye_height() / SCALE_FACTOR`, and the SDK aim helpers consume that live `camera_offset` by default.

PR #723 carried the unique SHODAN log-4 regression but was closed as superseded; its closure comment explicitly kept #721 open until the scenario was ported. This PR adds only that missing coverage against current `main`.

## Negative-first evidence

The same fresh-mission sequence was replayed against exact pre-fix parent `9acf17c1` and current `55c39cac`:

| | Pre-fix | Current |
|---|---:|---:|
| Reported camera offset Y | 1.6 | 0.48000002 |
| Actual production crouched eye | 0.48 | 0.48 |
| Collected 9/4 | no | yes |
| Reader panel | none | template 290 |

The new live-camera assertion fails on the pre-fix runtime, and its production squeeze leaves 9/4 uncollected. On current `main`, two fresh runs passed with different launch-local runtime IDs.

## Verification

- focused `shodan-log4-interaction.e2e.test.ts` — passed; live camera `[0, 0.48000002, 0]`, dynamic target selected, 9/4 collected, reader opened
- existing SHODAN Delacroix log-2 production jump/aim/squeeze E2E — passed
- existing MedSci log-reader collection/persistence E2E — passed
- SDK fast suite — 26 passed, 209 E2E-gated skips
- SDK world-aim unit suite — 12 passed
- focused media/camera/crosshair Rust tests — 3 passed
- `cargo fmt --all -- --check`; `git diff --check`

## Player-observable verification

Crouched SHODAN log 4 interaction before and after

<sub>Matched fresh `shodan.mis` runs use dynamic template-290 discovery, the same crouched platform staging, SDK `aimAt(surface)` with no eye-height override, and the same one-frame production squeeze pulse. The pre-fix build aims into darkness and collects nothing; current `main` highlights the disc, records deck 9/log 4, and opens Delacroix's “Re: Stakes” reader. Captured headlessly through deterministic fixed-timestep stepping.</sub>

### Current result

Current reader result

### Before → after

Stale standing aim versus live crouched aim

[Full-resolution matched captures](https://gist.github.com/tommy-xr/6577220bbf4bc0796bcd327a7e68be71).

## Adjacent tracked scope

Current `main` still leaves collected discs rendered and physical (`template 290` body count `1 → 1`), as the media intentionally shows. That global retail-consumption behavior is tracked by #797 and its existing open PR #838; it is not duplicated here.

---

## Rebased onto #917 and reworked

This PR originally asserted that the reader **opens on the frob that collects the disc**:

```js
assert.equal((await game.ui.state()).active_panel?.template_id, 290,
  "the normal log script should open Delacroix log 4 in the reader");
```

That pins behavior the original does not have — collecting a log files it in the PDA and reading it is an explicit act — so merging it as-is would have made the correction in #917 a test-breaking change, with the next person unable to tell a bug from a contract.

Rebased onto #917 and reworked to the new contract, which widens what the scenario proves:

- the log is collected **unread**
- **no panel is raised** by collecting
- the disc **loses its world presence** (it is consumed)
- `ReadLastUnreadLog` opens it and **marks it read**
- `CloseActivePanel` dismisses it

The setup this scenario exists for — the crouched landing on the authored hidden platform, the live crouched eye height, and the production camera ray selecting log 4 — is unchanged, which is why this stayed a separate PR rather than being folded into #917.

**Merge order:** #917 first; this rebases onto `main` behind it.

**Re-verified on the reworked scenario:** `SHOCK2_E2E=1 npx tsx --test test/shodan-log4-interaction.e2e.test.ts` — 1 passed, executed not skipped (12.7s), diagnostic `SHODAN log 4: camera [0,0.48,0], aim target 844, collected 9/4`.

### DIFF
--- a/runtimes/desktop_runtime/src/input_mapper.rs
+++ b/runtimes/desktop_runtime/src/input_mapper.rs
@@ -159,6 +159,11 @@ impl DesktopInputMapper {
                 action: InputAction::DebugCalmAll,
             },
             Binding {
+                key: Key::U,
+                modifier: Modifier::None,
+                action: InputAction::ReadLastUnreadLog,
+            },
+            Binding {
                 key: Key::M,
                 modifier: Modifier::None,
                 action: InputAction::ToggleMap,
--- a/shock2vr/src/gui/gui_script.rs
+++ b/shock2vr/src/gui/gui_script.rs
@@ -109,6 +109,12 @@ where
             // playing its clip) fire in both presentations. A gui can veto the
             // open (a content-less log disc) and/or ask for fresh per-open
             // state (the reader's scroll position).
+            // Opened without a frob (the audio-log reader): give the gui the
+            // same fresh per-open state a frob-open would have.
+            MessagePayload::PanelOpened => {
+                self.gui.prepare_state_on_frob(&mut self.state);
+                Effect::NoEffect
+            }
             MessagePayload::Frob => {
                 let frob_effect = self.gui.on_frob(entity_id, world);
                 if !self.gui.opens_on_frob(entity_id, world) {
--- a/shock2vr/src/gui/gui_script.rs
+++ b/shock2vr/src/gui/gui_script.rs
@@ -120,7 +120,8 @@ where
                 if !self.gui.opens_on_frob(entity_id, world) {
                     return frob_effect;
                 }
-                self.gui.prepare_state_on_frob(&mut self.state);
+                // The per-open reset arrives via `PanelOpened`, dispatched by
+                // the `OpenPanel` handler, so both open paths share one hook.
                 Effect::combine(vec![Effect::OpenPanel { entity: entity_id }, frob_effect])
             }
             MessagePayload::GUIHover {
--- a/shock2vr/src/input/actions.rs
+++ b/shock2vr/src/input/actions.rs
@@ -83,6 +83,15 @@ pub enum InputAction {
     /// Toggle the flat-mode automap panel (the original's BIOFULL MAP button /
     /// `M` key). No-op in VR. See `projects/flat-ui-panels.md` §5.
     ToggleMap,
+
+    /// Play back the most recently collected audio log the player has not read
+    /// yet, opening the reader on it (the original's `play_unread_log`).
+    /// Collecting a disc only files it in the PDA, so this is how a log is
+    /// actually read.
+    ReadLastUnreadLog,
+
+    /// Dismiss the open MFD panel (the original closes its overlays on Tab).
+    CloseActivePanel,
 }
 
 impl InputAction {
@@ -120,6 +129,8 @@ impl InputAction {
             InputAction::DebugCalmAll,
             InputAction::DebugForceChase,
             InputAction::ToggleUseMode,
+            InputAction::ReadLastUnreadLog,
+            InputAction::CloseActivePanel,
             InputAction::ToggleMap,
         ]
     }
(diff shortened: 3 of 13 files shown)


tommy-xr:
i don't think this validates the correct thing - we can revisit later if this is useful

Issue #931, opened 2026-08-13
test(e2e): creature-loot reader assertion broken on main since #917


tommy-xr:
`creature-loot.e2e.test.ts` fails on `main` and has since #917 merged. Found while validating an unrelated UI stack; filing rather than leaving it as an unexplained red in everyone's e2e runs.

## Failure

```
not ok - creature loot: a slain creature is lootable, a live hostile is not, and Watts yields the 12451 disc
  error: frobbing the looted disc should open the reader MFD bound to it
  undefined !== 251
  expected: 251
```

The test loots the 12451 disc from Dr. Watts, frobs it, and expects the reader MFD to open **bound to that disc entity**. It now gets `undefined` — no panel bound.

## Bisect

| Commit | Result |
| --- | --- |
| `9033073` refactor(ui): unify 2D canvas descriptions (#840) | **pass** |
| `c9626f5` feat(gameplay): read audio logs on demand instead of on pickup (#917) | **fail** |
| `86432f5` (current main) | **fail** |

Each run was `SHOCK2_E2E=1 node --test --test-concurrency=1 dist/test/creature-loot.e2e.test.js` on an idle machine after a full rebuild, so this is not the load-flakiness that affects the AI/death tests.

## Likely cause

#917 changed audio logs from "read on pickup" to "read on demand". The test's expectation — frob a looted disc, get a reader MFD bound to the disc entity — is exactly the behavior that changed. So either:

- the new on-demand path does not bind the panel to the source entity (a real bug — the reader would open unbound, or not at all), or
- the expectation is simply stale and the test should assert the new contract.

Worth deciding which, since #917 rewrote `log-reader.e2e.test.ts` and `log-pickup-consumption.e2e.test.ts` but left this third test asserting the old behavior — suggesting it was just missed rather than deliberately left.

## Note

CI does not run e2e (opt-in behind `SHOCK2_E2E=1`), which is why this merged green.

PR #943, opened 2026-08-13
fix(vr): gate authored melee contact damage

### LINKED ISSUE #942: fix(vr): wire authored melee weapons to collision damage
## Summary

Authored Wrench entities do not instantiate the implemented `MeleeWeapon` collision script. In VR, a held melee weapon is intentionally nonphysical and projectile-less `WeaponScript::TriggerPull` intentionally no-ops because the physical `MeleeWeapon::Collided` path is expected to deal damage. Releasing the wrench restores physics, but collisions still cannot emit `Damage` because the actual Wrench has only inherited `WeaponScript` entries.

Campaign: Command · no tweak · 25th Anniversary assets · VR · seed `2110635141`. Medium-severity playable finding: it invalidated the obvious wrench solution for the one-HP Ops card glass, but a reachable loaded Gren Launcher1393 provides a retail bypass, so this is not a campaign blocker.

## Player-observed reproduction

1. Launch `command2.mis` with default `--vr`.
2. Production-squeeze world Wrench786 and carry it to Window2 glass82 guarding Ops Override card507.
3. Place/swing the held wrench against the pane and release squeeze so it becomes physical. Repeat with center, above-pane, and below-pane contacts.
4. Observe the wrench collision-resolve against the pane, while Glass82 remains at authored HP1 and no Damage message occurs.

Local reviewed evidence exists but is not uploaded: pre-action glass save SHA-256 `9524a4fb0f7aca6e8375908d52e6c00fa0fbdc06cd9cc0427d70b4d9738e72f5`; campaign frontier SHA-256 `a7ba41788f7cb86fb6f60e318fb3402d345615fc21b61f799d46dbdaf25b58d1`, command2, HP21.

## Root cause

- Gamesys Wrench template `-928` authors/inherits `PropScripts { scripts: ["WeaponScript"] }`; real Wrench786 has no script named `wrench`.
- Entity creation sorts/deduplicates those inherited entries, yielding `WeaponScript` only.
- The script factory implements the literal script name `wrench` as `MeleeWeapon + InternalSwitchHeldModelScript`, but that factory branch is never reached for ordinary authored Wrench entities.
- Projectile-less `WeaponScript` explicitly no-ops in VR and documents that VR melee damage must come from `MeleeWeapon::Collided`.
- A held VR item is made unphysical; squeeze release restores ordinary physics, but without `MeleeWeapon` its contact remains harmless.

No duplicate was found. Merged flat-melee work #319/#321 covers the crosshair/raycast presentation, not this VR collision-script composition.

## Expected

Actual authored melee weapon entities should compose the existing faithful melee collision behavior. A real VR physical impact should emit exactly the intended melee damage/impact sound without hardcoded mission IDs or a debug attack path.

## Acceptance

- Negative-first test proves an ordinary authored Wrench currently lacks the collision damage behavior.
- Production VR interaction with an authored Wrench damages a one-HP breakable target through the real melee path.
- Preserve flat melee, grab/drop/save-load, sound/material response, and non-melee weapon behavior.
- Use gamesys/template/script semantics generically; no Command coordinates or object IDs in production code.
- Add player-visible before/after evidence because weapon impacts are visible.

tommy-xr:
Fixes #942

## Summary

- recognize authored melee weapons generically from `PropLimbModel` and add an internal VR-only, trigger-gated contact-damage script
- keep the real world collider while a melee weapon is held, with controller-driven kinematic physics that ignores the player but contacts world props, actors, and kinematic/fixed breakables
- close the damage window on trigger release or drop, restore ordinary dynamic loose-prop physics on drop, and limit each target to one hit per pull
- deal the weapon's **authored** contact-stim damage (the Wrench's `StimSource(WeaponBash, 6.0/9.0, Contact)`) through `stim_response::contact_stim_damage`, the same data path AI melee uses, so a VR swing respects receptrons/armor instead of a flat placeholder
- keep the contact body across a save/load, level transition, and `Effect::GrabEntity`, not only on a fresh world grab
- preserve the legacy literal `wrench` script for the Maintenance Tool template (`-2949`); the player Wrench is template `-928` and continues through `WeaponScript`

## VR result

VR Wrench trigger/contact demo

<sub>The gloved VR hand, the real authored Wrench and the shipped one-HP `Window 2` pane in `command2` are all in frame for the whole swing. With the trigger <b>idle</b> the wrench sweeps physically through the pane and the pane survives; with the production VR trigger <b>held</b> the same contact emits `Damage` → `Slay`, the pane is removed and the keycard behind it is exposed. Captured headlessly in default VR via `/v1/step` + `/v1/screenshot` at a deterministic fixed timestep.</sub>

### Before → after

origin/main versus fixed VR Wrench contact

Both halves are the *same* fresh-launch request sequence against separately-built runtimes. On `origin/main` the armed contact produces no messages at all and the pane survives; on this branch the runtime trace records `Damage` then `Slay` for the pane and the alcove opens.

The camera is aimed with the SDK's `input.lookAtWorldPoint`, which composes pawn rotation with the real eye height from `/v1/info`, rather than hand-tuned `head.look` yaw/pitch — the earlier revision of these visuals framed empty scenery because the raw channel is pawn-local.

## Campaign evidence

Reviewed play-through configuration: **Command · no tweak · 25th Anniversary assets · VR · seed 2110635141**.

The deterministic regression (`tools/shock2-sdk/test/vr-melee-contact.e2e.test.ts`) uses the same authored interaction chain from `command2.mis`:

1. discover and physically grab the real player Wrench (mission object/template id `786`, inheriting Wrench `-928`) through `VirtualHand`
2. sweep it through the authored one-HP `Window 2` pane (mission object/template id `82`) with the trigger idle and verify the pane remains
3. leave contact, pull the production VR trigger, swing back into the pane, and verify runtime `Damage` + `Slay`
4. release and drop against a second authored one-HP pane, verify the Wrench is dynamic again and the pane remains

A second scenario, `a melee weapon carried through a save/load keeps its contact body and still damages`, grabs the Wrench, saves, loads, and asserts the restored weapon is still a kinematic contact body and still breaks a pane. It fails on the parent commit with `bodies: []`.

No mission ids exist in production code; stable mission ids are used only by the E2E to discover shipped fixtures because runtime entity ids shuffle every launch.

## Validation

- `RUSTFLAGS="-D warnings" cargo test -p shock2vr` — 580 passed
- `npm test` (SDK TypeScript + unit suite) — 26 passed, 213 opt-in E2E skipped
- `SHOCK2_E2E=1 npm run test:e2e` — 240 tests, 230 passed, 7 skipped, 3 failed. All three are unrelated to this change: `security-ecology` and `trap-unlock` failed with `fetch failed` under suite load and **pass in isolation**; `creature-loot` fails **identically on `origin/main`** (verified by rebuilding the runtime at `origin/main` and re-running) and is tracked as #931.
- targeted real-runtime SDK E2E — 1 passed: `VR authored melee damages only during a trigger-gated physical contact window`
- deterministic `origin/main` vs feature capture against fresh private VR runtimes: baseline pane survives with no pane `Damage`/`Slay`; feature emits `Damage` + `Slay` and removes it

## Compatibility

- flatscreen `WeaponScript` firing/wield behavior is unchanged; the internal contact script is inert outside VR — physical contact is never the damage trigger in flat mode
- authored grab/drop transforms and collision sounds remain in their existing paths
- guns keep their current VR trigger behavior; the melee physics path is selected only for entities with an authored limb model
- attack-window state is intentionally transient across save/load; restored sessions begin harmless until a new trigger pull (the *collider* now survives the load — only the armed window resets)

## Review follow-ups

Found while reviewing this change, filed rather than folded in:

- #951 — flat melee's own timing problem: damage applied on the `TriggerPull` edge rather than on a swing animation event, plus the legacy `wrench` script's unconditional contact damage. Pre-existing on `main`.
- #954 — the held weapon keeps `ENTITY` membership, so the VR hands' selection ray can now hit the weapon the hand is holding.
- #955 — the attack window stays open for as long as the trigger is held, and each target can be hit only once per pull.
- #956 — the VR contact path sends `impact: None`, losing the death-ragdoll directionality the flat swing provides.

### DIFF
--- a/shock2vr/src/scripts/melee_weapon.rs
+++ b/shock2vr/src/scripts/melee_weapon.rs
@@ -26,6 +26,11 @@ impl MeleeWeapon {
 /// edge is that production attack gesture: contacts before it, after release,
 /// and after dropping the weapon are harmless. A target can be damaged only
 /// once per pull even if the physical swing chatters across several contacts.
+///
+/// Physical contact is *never* the damage trigger outside VR - a flat swing is
+/// `WeaponScript`'s aimed short-range raycast, so bumping a wielded wrench into
+/// scenery must do nothing. The whole script is therefore inert in flat mode
+/// rather than guarding individual messages.
 pub struct TriggeredMeleeWeapon {
     attack_active: bool,
     hit_entities: HashSet<EntityId>,
@@ -48,8 +53,12 @@ impl Script for TriggeredMeleeWeapon {
         _physics: &PhysicsWorld,
         msg: &MessagePayload,
     ) -> Effect {
+        if !is_vr(world) {
+            return Effect::NoEffect;
+        }
+
         match msg {
-            MessagePayload::TriggerPull if is_vr(world) => {
+            MessagePayload::TriggerPull => {
                 self.attack_active = true;
                 self.hit_entities.clear();
                 Effect::NoEffect
--- a/shock2vr/src/scripts/melee_weapon.rs
+++ b/shock2vr/src/scripts/melee_weapon.rs
@@ -1,8 +1,13 @@
+use std::collections::HashSet;
+
 use cgmath::{EuclideanSpace, vec3};
 use dark::properties::{CollisionType, PropCollisionType};
-use shipyard::{EntityId, Get, View, World};
+use shipyard::{EntityId, Get, UniqueView, View, World};
 
-use crate::{physics::PhysicsWorld, util::get_position_from_transform};
+use crate::{
+    PresentationMode, mission::GlobalPresentationMode, physics::PhysicsWorld,
+    util::get_position_from_transform,
+};
 
 use super::{Effect, Message, MessagePayload, Script, script_util::play_impact_sound};
 
@@ -15,6 +20,89 @@ impl MeleeWeapon {
     }
 }
 
+/// Contact damage for the player's authored melee weapons (`PropLimbModel`).
+///
+/// Dark opens melee collision only during an attack window. In VR the trigger
+/// edge is that production attack gesture: contacts before it, after release,
+/// and after dropping the weapon are harmless. A target can be damaged only
+/// once per pull even if the physical swing chatters across several contacts.
+pub struct TriggeredMeleeWeapon {
+    attack_active: bool,
+    hit_entities: HashSet<EntityId>,
+}
+
+impl TriggeredMeleeWeapon {
+    pub fn new() -> Self {
+        Self {
+            attack_active: false,
+            hit_entities: HashSet::new(),
+        }
+    }
+}
+
+impl Script for TriggeredMeleeWeapon {
+    fn handle_message(
+        &mut self,
+        entity_id: EntityId,
+        world: &World,
+        _physics: &PhysicsWorld,
+        msg: &MessagePayload,
+    ) -> Effect {
+        match msg {
+            MessagePayload::TriggerPull if is_vr(world) => {
+                self.attack_active = true;
+                self.hit_entities.clear();
+                Effect::NoEffect
+            }
+            MessagePayload::TriggerRelease | MessagePayload::Drop => {
+                self.attack_active = false;
+                self.hit_entities.clear();
+                Effect::NoEffect
+            }
+            MessagePayload::Collided { with }
+                if self.attack_active && self.hit_entities.insert(*with) =>
+            {
+                melee_impact(entity_id, *with, world)
+            }
+            _ => Effect::NoEffect,
+        }
+    }
+}
+
+fn is_vr(world: &World) -> bool {
+    world
+        .borrow::<UniqueView<GlobalPresentationMode>>()
+        .map(|mode| mode.0 == PresentationMode::Vr)
+        .unwrap_or(false)
+}
+
+fn melee_impact(entity_id: EntityId, with: EntityId, world: &World) -> Effect {
+    let damage_effect = Effect::Send {
+        msg: Message {
+            to: with,
+            payload: MessagePayload::Damage {
+                amount: 1.0,
+                impact: None,
+            },
+        },
+    };
+    // Impact sound: the weapon's collision schema (weapontype class tag + hit
+    // material - wrench on metal clangs, on a creature thuds), unless its
+    // collision type opts out.
+    let no_sound = world
+        .borrow::<View<PropCollisionType>>()
+        .ok()
+        .and_then(|v| v.get(entity_id).ok().map(|c| c.collision_type))
+        .is_some_and(|flags| flags.contains(CollisionType::NO_COLLISION_SOUND));
+    let sound_effect = if no_sound {
+        Effect::NoEffect
+    } else {
+        let position = get_position_from_transform(world, entity_id, vec3(0.0, 0.0, 0.0));
+        play_impact_sound(world, entity_id, with, position.to_vec())
+    };
+    Effect::Multiple(vec![damage_effect, sound_effect])
+}
+
 impl Script for MeleeWeapon {
     fn handle_message(
         &mut self,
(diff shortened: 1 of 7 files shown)


shock2vr/src/scripts/melee_weapon.rs
```diff
@@ -24,34 +49,191 @@ impl Script for MeleeWeapon {
...
-                Effect::Multiple(vec![damage_effect, sound_effect])
+            MessagePayload::TriggerPull if is_vr(world) => {
+                self.attack_active = true;
+                self.hit_entities.clear();
+                Effect::NoEffect
+            }
+            MessagePayload::TriggerRelease | MessagePayload::Drop => {
+                self.attack_active = false;
+                self.hit_entities.clear();
+                Effect::NoEffect
```
tommy-xr: Should trigger pull / trigger release have any effect in is_vr? Wondering if clearing hit_entities would be problematic in vr?

shock2vr/src/scripts/melee_weapon.rs
```diff
@@ -24,34 +49,191 @@ impl Script for MeleeWeapon {
...
-            MessagePayload::Collided { with } => {
-                let damage_effect = Effect::Send {
-                    msg: Message {
-                        to: *with,
-                        payload: MessagePayload::Damage {
-                            amount: 1.0,
-                            impact: None,
-                        },
-                    },
-                };
```
tommy-xr, line 36: Pre-existing issue - but for the non-vr (flat) case, i think there is an animation event we should respect and then do a short-range raycast. The damage  application feels weird / unpredictable (and not faithful to the original) in flat mode.

tommy-xr:
The visual evidence for this isn't accurate (not showing anything), wonder if the head position is incorrect? For vr, we might need a utility to help the head look at the hands?

shock2vr/src/scripts/melee_weapon.rs
tommy-xr: Good catch — the asymmetry was real and confusing. `TriggerPull` was guarded by `is_vr` but `TriggerRelease`/`Drop` weren't.

Fixed by hoisting the check to the top of `handle_message`, so the whole script is inert outside VR rather than guarding message-by-message. That also settles your follow-up point: physical contact is never the damage trigger in flat mode — a flat swing is `WeaponScript`'s aimed raycast.

On clearing `hit_entities` being problematic in VR: it's not for the release edge itself, but digging into it surfaced a real gap. Nothing bounds the window or rate-limits pulls, so (a) the window stays open for as long as you hold the trigger, sweeping a permanently-armed weapon across everything it touches, and (b) trigger-spam has no cooldown, while you *can't* hit the same target twice within one pull. Dark's melee window is a bounded swing, not "as long as the button is down". Filed as #955 rather than widening this PR.

shock2vr/src/scripts/melee_weapon.rs
tommy-xr: Agreed, and filed as #951 so this PR stays VR-scoped.

Two distinct problems in there, both pre-existing on `main`:

1. **No swing timing.** The short-range raycast already exists (`weapon_script.rs:106` → `melee_swing`), but it fires on the `TriggerPull` edge rather than on an event in the swing animation — so damage lands before the weapon has visibly moved. That's the unpredictability you're describing.
2. **Contact damage is still live in flat for the legacy path.** `MeleeWeapon` (mapped from the literal `wrench` script, Maintenance Tool `-2949`) damages unconditionally on `Collided` in both presentations. So a dropped/thrown Maintenance Tool still deals contact damage in flat — exactly what this PR's new doc comment declares wrong.

#951 covers both, so flat ends up with a single melee damage source: an animation-timed raycast.

tommy-xr:
## Re-captured the visuals — the old ones were framing empty scenery

You were right, and the cause was the VR camera frame. The capture set `head.look` yaw/pitch by hand, but that channel is **pawn-local**, and in `--vr` the eye sits at `player_position + camera_offset` (~1.04 above the *pawn origin*, not the floor) with default forward `+Z`. So the camera was pointed at a barrel while the captions described a wrench and a pane that were never in shot.

No new utility was needed — the SDK already has the one you were thinking of: **`game.input.lookAtWorldPoint(target)`** (`tools/shock2-sdk/src/game.ts:440`), plus `game.player.aimAt(entity)`. Both compose pawn rotation with the real eye height from `/v1/info`. The E2E now stages through the helper and expresses its sweep relative to the pane, with the held grip's 0.4 offset documented, so this class of bad capture can't recur silently.

The new shot stands off to the side of the pane so the **gloved hand, the wrench, and the target are all in frame for the whole swing**. The before/after is a genuine two-binary comparison: `origin/main` was rebuilt and run through the byte-identical request sequence — baseline emits no pane messages and the pane survives; this branch emits `Damage` → `Slay` and the keycard behind the pane is exposed.

## Two real defects the review turned up

Reviewing the change alongside your comments surfaced two problems worth fixing before this lands:

**1. Damage was a flat `1.0`.** The same wrench swung in flat mode deals `MELEE_DAMAGE = 6.0`, so VR melee was 6× weaker — and the player Wrench `-928` authors the actual Dark mechanism:

```
StimSource(intensity: 6.0, propagator: Contact) -> -3058 (WeaponBash)
StimSource(intensity: 9.0, propagator: Contact) -> -3058 (WeaponBash)
```

`stim_response::contact_stim_damage` already resolves exactly this through the victim's receptrons and is already used by AI melee (`ai_util.rs:432`). The VR path now uses it, so a swing costs what the gamesys says and respects armor/immunities; a contact resolving to no authored damage emits nothing instead of a free 1-point tap. The legacy `wrench` script keeps its historical 1.0 — that's #951's business, not a VR bugfix's.

**2. The contact body did not survive a save/load or level transition.** `set_held_melee` was reachable only from `VirtualHandEffect::HoldItem`, which only a *fresh* world grab emits. Restore goes through `VrInteraction::grab` (returns no effects) and then unconditionally stripped held-item physics — so a wrench carried through a load came back with **no rigid body at all** and silently never reported a contact again. `Effect::GrabEntity` (equip a carried weapon, backpack double-click) had the same hole. Both now attach the contact body.

That one is the path a player actually uses to carry a weapon, and the original E2E passed straight through it because it only covered the fresh-grab case. There's now a scenario that grabs, saves, loads, and asserts both the kinematic body and a real kill — it fails on the parent commit with `bodies: []`.

## Filed rather than folded in

- #951 — flat melee timing + the legacy `wrench` script's unconditional contact damage
- #954 — the held weapon keeps `ENTITY` membership, so the VR hands' selection ray can hit the weapon the hand is holding (and the *other* hand can grab it)
- #955 — the attack window stays open for the whole trigger hold; no per-swing bound, no pull cooldown
- #956 — `impact: None` loses the death-ragdoll directionality the flat swing provides

## Validation

`cargo check -D warnings` (shock2vr, desktop_runtime, debug_runtime) · `cargo test -p shock2vr` 582 passed · SDK `npm test` 26 passed · full `SHOCK2_E2E=1 npm run test:e2e` 230/240 passed, 7 skipped, 3 failed — `security-ecology` and `trap-unlock` failed with `fetch failed` under suite load and **pass in isolation**, and `creature-loot` fails **identically on `origin/main`** (verified by rebuilding the runtime there), tracked as #931.

PR #944, opened 2026-08-13
fix(audio): honor schema volume and pan

### LINKED ISSUE #871: Honor gamesys schema volume/pan in resolve_schema (one-shot sounds play at full clip volume)
Follow-up from the earth.mis training-audio work (#856, #870).

## Current state

`resolve_schema` (`shock2vr/src/mission/mission_core.rs:6270`) only maps a schema name to a randomly-selected sample name via `SoundSchema::get_random_sample`. The parser behind it (`dark/src/gamesys/sound_schema.rs`) reads just the per-schema sample list and each sample's selection `frequency` — nothing else.

In the Dark gamesys, schema objects carry playback parameters we currently ignore entirely: volume (dB attenuation), pan, and related fields (delay/fade/environmental flags). Every schema-resolved one-shot therefore plays at full clip volume.

## Why it matters now

#870 makes entity-sourced `Effect::PlaySound` positional at the source entity (the faithful mitigation for overlapping training narrations — a stale montage segment ~23 units away is now attenuated by distance instead of playing at the player's ears). Distance attenuation is only half of the original loudness model; the other half is the authored schema volume. If specific narrations/effects now land too loud or too quiet, honoring schema volume is the faithful lever — not hand-tuned scale factors.

## Suggested shape

1. Extend `dark/src/gamesys/sound_schema.rs` to parse the schema's volume/pan (and note which other fields exist, even if unparsed).
2. Have `resolve_schema` return sample + volume (struct instead of bare `String`), and apply the volume in the `Effect::PlaySound` / env-sound playback paths (rodio sinks support per-sink volume).
3. Verify against a schema with a known strong authored attenuation by comparing in-game loudness before/after; `GET /v1/audio/recent` (#855) can assert the resolved volume value once it's recorded.

Related knob for ambients (different object class, already parsed but likewise unused for playback shaping): `P$AmbientHacked` radius/volume/`NO_SHARP_CURVE` (`dark/src/properties/prop_ambient_hacked.rs`).

tommy-xr:
## Summary

- parse the retail `P$SchPlayPa` / `sSchemaPlayParams` layout (`flags`, `volume`, `pan`, `initialDelay`, `fade`, all 32-bit) and hydrate schema inheritance
- resolve a schema to its frequency-weighted sample plus authored volume/pan without changing sample weighting
- apply millibel gain at the real rodio sink, apply fixed/ranged schema pan to listener-relative sounds, and keep world-space panning for positional sounds
- carry resolved volume/pan through scripted sounds, speech, and environmental one-shots, and expose applied settings through `/v1/audio/recent`
- preserve #939 live emitter tracking and head-relative listener ears while adding schema gain to positional sinks

Fixes #871

## Reproduction and root cause

`SoundSchema::read` only loaded `SchSamp` and direct symbolic names. `resolve_schema` returned a bare sample string, so the playback path had no authored settings to propagate and every schema one-shot used default sink gain/pan.

The retail `shock2.gam` contains a uniform 20-byte `P$SchPlayPa` record. Representative objective values:

- `linebeep`: volume `-2500`, pan `0`
- `log0219`: inherits volume `-300`, fixed pan `-1000` from `SPEECH_LOGS`
- Earth `trg0001`: inherits volume `-500` from `SPEECH_BRIEFING`
- keypad `bkeypad`: volume `-1500`, fixed pan `-1000`

Negative-first evidence:

- the retail-gamesys integration test was added first and failed on base because `SoundSchema` had no play-param resolution (`no method named resolve`)
- the engine sink test was added first and failed because there was no playback-settings path and no panned sink append

## Rebase reconciliation

Rebased onto current `origin/main` through `e09fc08`, including #939. The generic positional sink still stores `TrackedEmitter<entity>`, refreshes the live source position every update, and derives ears from head rotation. The gain-aware spatial entry point carries both the source entity and schema gain. Only listener-relative/static playback uses the ordinary rodio `Sink` plus channel-volume pan.

## Verification

- `git diff --check origin/main...HEAD`
- `DARK_ASSET_PATH=/Users/bryan/ss2-data-unpacked cargo test -p dark -p engine -p shock2vr`: passed; `shock2vr` 615 passed on the overlap-reconciled tree, with dark/engine suites also green
- `RUSTFLAGS="-D warnings" cargo check -p dark -p engine -p shock2vr -p desktop_runtime -p debug_runtime -p dark_query`: passed; final-head follow-up also passed for dark/engine/shock2vr/debug runtime
- SDK fast: 245 total, 26 passed, 219 E2E-gated skipped, 0 failed
- retail parser/inheritance, millibel conversion, fixed/ranged pan, and exact five-word parser tests: passed
- engine audio tests: 4 passed, including #939 live emitter refresh and #871 rodio gain/channel-pan propagation
- #939 listener-head-rotation and live-runtime-transform tests: passed
- focused live Earth: `/v1/audio/recent` reported inherited volume `-500`, gain `0.5623413`, and `pan_applied=false` for the spatial sink
- focused live keypad: sample `bkeypad`, volume `-1500`, gain `0.17782794`, pan `-1000`, and `pan_applied=true`
- full SDK E2E after audio/spatial conflict reconciliation: 245 total, 235 passed, 7 skipped, 3 failed; both changed audio scenarios and all 23 mission-load smokes passed
  - two failures match known unrelated baselines: creature-loot reader binding and SHODAN assassin corpse drift
  - the third was a transient AI last-known-position null lookup on the pre-advance tree; its required isolated rerun on clean current main passed 1/1, and the final rebased branch passed the same scenario 1/1 after the newly merged patrol fix

## PR visuals

Manager-approved non-renderable rationale: this change only alters audio sample amplitude and stereo channel balance at the sink. Renderer and simulation frames are unchanged, so deterministic GIF/PNG captures would be identical and misleading.

Objective substitutes:

- parsed retail authored values above, including inherited `log0219` values
- the rodio sink/queue test proves linear gain and left/right channel multipliers reach production sink primitives
- live `/v1/audio/recent` observations prove both positional volume and listener-relative volume/pan propagation
- #939 coexistence tests prove tracked emitter position and head-relative ears remain live

## Scope and risk

- positional sounds retain world-space panning; schema pan is visible in telemetry but not applied, matching the original 3D update behavior
- `initialDelay` and `fade` are parsed as part of the exact record layout but are not implemented by this volume/pan fix
- `P$AmbientHacked` ambient-object behavior is intentionally unchanged

### DIFF
--- a/dark/src/gamesys/gamesys.rs
+++ b/dark/src/gamesys/gamesys.rs
@@ -1,9 +1,7 @@
 use std::io;
 
-use rand::{distributions::WeightedIndex, prelude::Distribution, thread_rng};
-
 use crate::{
-    EnvMap, EnvSoundQuery, SoundSchema, SpeechDB, TagDatabase,
+    EnvMap, EnvSoundQuery, ResolvedSoundSchema, SoundSchema, SpeechDB, TagDatabase,
     gamesys::params::{HrmParams, SkillParams, TrainerCostTables},
     properties::{LinkDefinition, LinkDefinitionWithData, PropertyDefinition},
     ss2_chunk_file_reader::{self},
@@ -25,26 +23,17 @@ pub struct Gamesys {
 }
 
 impl Gamesys {
-    pub fn get_random_environmental_sound(&self, query: &EnvSoundQuery) -> Option<String> {
+    pub fn get_random_environmental_sound(
+        &self,
+        query: &EnvSoundQuery,
+    ) -> Option<ResolvedSoundSchema> {
         let tag_query = query.to_tag_query(&self.speech_db.tag_map, &self.speech_db.value_map);
         let result = self.env_tag_map.query_match_all(&tag_query);
         if result.is_empty() {
             return None;
         }
 
-        let sample = result[0];
-        let maybe_samples = self.sound_schema.id_to_samples.get(&sample);
-
-        maybe_samples?;
-
-        let samples = maybe_samples.unwrap();
-
-        let mut rng = thread_rng();
-        let weights = samples.iter().map(|s| s.frequency).collect::<Vec<u8>>();
-        let weight_index = WeightedIndex::new(weights).unwrap();
-        let idx = weight_index.sample(&mut rng);
-
-        Some(samples[idx].sample_name.to_owned())
+        self.sound_schema.resolve_id(result[0])
     }
 
     pub fn speech_db(&self) -> &SpeechDB {
--- a/dark/src/gamesys/sound_schema.rs
+++ b/dark/src/gamesys/sound_schema.rs
@@ -1,11 +1,11 @@
 use std::{collections::HashMap, io};
 
-use rand::{distributions::WeightedIndex, prelude::Distribution, thread_rng};
+use rand::{Rng, distributions::WeightedIndex, prelude::Distribution, thread_rng};
 use shipyard::{Get, View, World};
 use tracing::trace;
 
 use crate::{
-    properties::{PropSymName, PropTemplateId},
+    properties::{PropSchemaPlayParams, PropSymName},
     ss2_chunk_file_reader::ChunkFileTableOfContents,
     ss2_common::{read_i32, read_string_with_size, read_u8, read_u32},
     ss2_entity_info::SystemShock2EntityInfo,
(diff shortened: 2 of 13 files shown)


tommy-xr:
Need to find a good way to test this - what's a good location or area to test these new properties?

tommy-xr:
Rebased onto `main` (de5fe909). One conflict, in `engine/src/audio/mod.rs`, against the frontend menu SFX work that landed since (#1008 line): `main` added `PlayOptions { volume, looping }` + `play_audio_with`, which routed non-spatial playback through the `SpatialSink` core, while this PR moves non-spatial playback onto a plain `Sink` with `ChannelVolume` so authored schema pan can be applied.

Resolved by keeping this PR's `Sink`/pan path and folding looping into the single settings type rather than carrying two overlapping option structs:

- `AudioPlaybackSettings` gains a `looping: bool`; `add_to_sink_with_settings` wraps the source in `repeat_infinite()` when set.
- `PlayOptions` / `play_audio_with` / `add_to_spatial_sink_looping` are gone; `frontend_sfx.rs` now builds `AudioPlaybackSettings` and calls `play_audio_with_settings`. Loudness is unchanged - the default channel gains are `STATIC_CENTER_GAIN` (0.75), the same attenuation rodio applied to the centered spatial mix.
- `frontend_sfx`'s `SoundRecord` also picked up this PR's new audio-log fields (menu sounds have no schema: `volume_millibels`/`pan_millibels` `None`, `pan_applied: false`, `gain` from the settings).

`RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime -p dark -p engine` is clean and `cargo test -p dark -p engine -p shock2vr` passes (982 tests).

PR #950, opened 2026-08-13
feat(vr): render the 25AE authored hands as the player's VR hands

tommy-xr:
## What

Wires the 25AE authored hand poses (#932) into the VR hand renderer, and fixes
the reason nobody could see any of it: **`--vr` drew no hands at all**, in the
glove path too.

### The invisible hands were a frustum bug, not a render bug

`compute_view_matrix` applies the head as a **pawn-local** offset, so the eye
sits only `player_eye_height()/SCALE_FACTOR` (~1.04) above the **pawn origin** —
not above the floor. The debug runtime's default `--vr` rest pose at pawn-local
y `1.4` therefore put the hands ~0.36 **above** the eye (NDC y ≈ 1.58), off the
top of the screen. It contradicted its own comment ("shows the glove hands at
the bottom of the view"). Also worth knowing: default forward is **+Z**, since
pawn and head rotations are both yaw +90° and compose to 180°.

This was expensive to find because `/v1/scene` reported the hands as present,
correctly positioned, opaque and correctly wound — it had no way to say whether
the camera could *see* them. So each object now also reports **`ndc`** (its
origin in the drawing camera's clip space) and **`origin_on_screen`**. One call
now diagnoses it: the broken default reads `ndc.y = 1.98`; a hand behind the eye
reads `ndc: null`.

It is deliberately an **origin** test, not a bounds test — a large object
straddling the frustum edge reads `origin_on_screen: false` while plainly
visible, so the field is named for what it measures and `ndc` is the number to
reason with. A real bounds test needs an AABB on `SceneObject`, which doesn't
exist today.

### Anchoring: measured against the glove, improved but not finished

Two fixes so the hand lands where the glove did and moves less between poses:

- **Missing unit conversion.** `wrist_vector` stepped back `HAND_LENGTH_WORLD` —
  a length in *world* units — through *model-space* geometry, before
  `VIEWMODEL_HAND_SCALE` was applied. That's only ~61% of a hand, so the anchor
  landed mid-palm instead of at the wrist. The glove anchors a 19 cm hand at the
  wrist (`hand_glove::GLOVE_SCALE`); both renderers now agree on the landmark.
- **Per-pose `reach`.** The frame's far point is an extended fingertip on an open
  hand but only a knuckle on a curled one, so a single step-back put the wrist in
  a different place per pose. Each pose declares its far point's distance from
  the wrist in hand-lengths, alongside the authored `roll`.

Measured rather than eyeballed — hand centroid in a fixed framing, against the
glove rendered at the identical transform (the glove holds to 0.7px across poses):

| | mean offset from glove | pose-to-pose spread (sd_x) |
|---|---|---|
| before | 21.3px | 18.1px |
| unit conversion only | 14.7px | 16.9px |
| + per-pose `reach` | **12.8px** | **12.9px** |

So the unit conversion is what aligns the hand with the glove, and `reach` is
what evens the poses out (the worst pose, `Grip`, goes from 28px off to 16px).

**Neither is finished**, and I want to be straight about that: the glove's 0.7px
is the target and this is nowhere near it. The dominant remaining term is that
the source models carry very different amounts of forearm — `ar15_h` (Rest)
covers 268x174px where `atek_h` (Grip) covers 222x102 — which needs geometry
trimming, not anchor tuning. Centroid is also only a proxy for the wrist (a
fist's centroid genuinely sits nearer the wrist than an open hand's), so these
numbers should not be driven to zero.

The glove remains the fallback for a classic install (`PoseLibrary::new`
returning `None` is cached, so a non-25AE install doesn't retry the asset lookup
every frame).

VR hands cycling poses

<sub>Both hands in `--vr` (`debug_minimal`, 25AE install), swaying around <b>the shipped default rest pose</b> — so this exercises the rest-pose fix too, not just the pose library — and cycling Rest → SemiClosed → Grip → Rest. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

Default `--vr` view with no `/v1/control/input` at all — glove fallback (classic install) above, authored hands (25AE) below:

glove vs authored hands

## Testing

- `cargo test -p shock2vr hand_pose_library` — 6 tests. Two are new and were
  **negative-tested**: `the_wrist_anchors_a_real_hand_back_from_the_far_point`
  fails against the old unconverted step-back.
- `cargo fmt --check` and `RUSTFLAGS="-D warnings" cargo check -p shock2vr
  -p desktop_runtime -p debug_runtime` clean.
- Verified visually in `cargo dbgr --vr --mission debug_minimal` against both a
  25AE install and a classic install (glove fallback).

## Cross-engine review (`/xreview`) — findings addressed

Two independent reviewers (a Claude subagent and Codex) plus a dedicated
duplication pass. 0 Critical/High. Both engines independently raised the first
two, which is the high-confidence signal:

- **[AGREED] Partial 25AE install rendered *invisible* hands.** `PoseLibrary::new`
  activated if *any* pose loaded, but `render_hand` draws nothing when the
  requested pose is missing and the caller never falls back. A mod-layered or
  SCP/SHTUP install shipping `atek_h` but not `ar15_h` would make the hands
  vanish the moment the player opened their hand. Now all-or-nothing.
- **[AGREED] `/v1/scene` reported nonsense NDC near the eye plane.** `w > 0.0`
  admits objects sitting almost exactly on it, then divides by ~0 — producing
  values like `ndc.y = -12122363` that read as data, in the very field added to
  stop that misreading. Now gated on the near plane (where the GPU clips
  anyway), which also covers the screen-space HUD quads.
- **Size tuning silently moved the anchor.** `VIEWMODEL_HAND_SCALE` was primary,
  its doc invited tuning, and `wrist_vector` divided by it — so making the hands
  bigger slid the anchor along the finger axis. Inverted: the measured length is
  primary, the scale derives from it. Render is byte-for-byte unchanged.
- **Two mutually-exclusive `Option` renderer params** admitted meaningless
  states; replaced with a `HandRenderer` enum (six call sites → four).
- **Media didn't exercise the rest-pose fix** — it drove custom hand positions.
  Recaptured at the shipped defaults.

I also implemented an explicit `Material::draws_in_screen_space` predicate for
the NDC fix, then **reverted it after measuring that it changed nothing** the
near-plane guard didn't already handle (118 of 219 objects on `medsci1.mis`,
identical either way). Engine-wide trait surface for a case I couldn't produce
isn't worth carrying.

Duplication pass: `PoseLibrary::render_hand` does mirror `GloveRenderer::render_hand`'s
shape closely, and `anchor_of`'s basis construction overlaps
`util::get_rotation_from_forward_vector` (pre-existing). Both are left as
follow-ups rather than folded into this change.

## Notes / follow-ups

- **Known visual artifact, not fixed here: the source models drag weapon-model
  forearm into view.** The `ar15_h` island spans 1.0742 units of which only
  0.4068 is hand — the remaining ~31 cm (after scaling) renders as an
  open-ended hollow cone. `SemiClosed` (`atek_h` index 0) is the worst: a black
  sleeve wedge slices through the palm and fingers, visible in the GIF. Only
  `Grip` is clean. The module docs anticipate a "common-cuff trim" and this
  ships without one; the real fix is trimming geometry behind the anchored
  wrist during hand-island extraction, which is geometry work rather than
  anchor tuning. Flagging it rather than burying it — the hands are usable at
  the default rest pose but not yet clean in every pose.
- `AssetCache` keys by `importer.type_id()`, so any new view of an already
  imported type still needs its own newtype wrapper.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -511,6 +511,15 @@ The project supports experimental flags for gating in-progress features during d
   action reloads the current level in place to exercise this. See
   `projects/loading-screen.md`. Without it, transitions are synchronous and unchanged.
 
+- **`vr_hands_25ae`**: in VR, draw the player's hands using the 25th Anniversary
+  Edition's authored first-person hands (`hand_pose_library`) instead of the
+  SteamVR glove, snapping between `Rest` / `SemiClosed` / `Grip`. Requires a 25AE
+  install; a classic install has none of these models and falls back to the glove
+  regardless. **Unfinished** - the source models drag weapon-model forearm
+  geometry into view (`SemiClosed` has a sleeve wedge through the palm), and the
+  per-source handedness is not yet declared, so hands can appear mirrored. Off by
+  default; the glove is the shipped behaviour.
+
 - **`nav_bridges`**: reconnect the AI navigation mesh for full-map pathfinding. The
   shipped mission data partitions the walk graph into per-area islands with no links
   between them (original AI pathfinding was area-local); this flag synthesizes
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -249,6 +249,15 @@ For debugging visual/rendering changes without a full interactive session:
    # data rather than screenshots. A translucent prop with depth_write=false
    # came from PropRenderAlpha; with depth_write=true it is the material's own
    # transparency.
+   #
+   # Each object also reports `ndc` (its ORIGIN in the drawing camera's clip
+   # space) and `origin_on_screen`. Use these before theorising about culling or
+   # materials when something "renders nothing": `ndc.y > 1` is off the top of
+   # the screen, `ndc: null` is behind the eye, `|ndc.x| > 1` is off the side.
+   # `position` is world-space and says nothing about visibility. Note it is an
+   # origin test, not a bounds test - a large object straddling the frustum edge
+   # reads `origin_on_screen: false` while plainly visible, so read `ndc` for
+   # how far off it is rather than trusting the boolean as a verdict.
    curl "http://127.0.0.1:8080/v1/scene?transparent=true"
    curl "http://127.0.0.1:8080/v1/scene?entity_id=246"
 
--- a/runtimes/debug_runtime/src/commands.rs
+++ b/runtimes/debug_runtime/src/commands.rs
@@ -282,7 +282,7 @@ pub struct SceneListResult {
 }
 
 /// One scene object as submitted to the renderer
-#[derive(Debug, Serialize)]
+#[derive(Debug, Serialize, Clone)]
 pub struct SceneObjectSummary {
     pub entity_id: Option<u64>,
     pub name: Option<String>,
@@ -297,6 +297,19 @@ pub struct SceneObjectSummary {
     pub clear_depth: bool,
     /// Front-face winding used for culling, or absent when double-sided.
     pub backface_culling: Option<String>,
+    /// The object's ORIGIN in normalized device coordinates for the camera that
+    /// drew this frame: x/y in [-1, 1] across the viewport, z in [-1, 1] between
+    /// the near and far planes. `None` when the origin is at or behind the eye
+    /// plane (w <= 0), where NDC is meaningless.
+    pub ndc: Option<[f32; 3]>,
+    /// Whether the object's ORIGIN falls inside the view frustum. Named for
+    /// what it actually tests: it is not a visibility verdict. A big object
+    /// straddling the edge reads `false` while plainly visible (the VR hands do
+    /// exactly this), and `true` does not rule out occlusion or backface
+    /// culling. Read it with `ndc`, which says *how far* off and in which
+    /// direction - `ndc.y > 1` is above the top of the screen, `ndc == null`
+    /// is behind the eye. That is the question `position` alone cannot answer.
+    pub origin_on_screen: bool,
 }
 
 /// List of physics rigid bodies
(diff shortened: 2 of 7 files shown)


tommy-xr:
Scoping decision (owner): first pass uses a **single best hand pose/model everywhere** — idle, trigger-pull, and holding items all render the same hand. Pose *switching* is deferred: finding good models to switch between was the hard part of the earlier attempt, and one good hand beats three mediocre transitions.

**Alternative path also on the table**: instead of the 25AE authored hands, reskin the existing glove model to a bare-hand skin and add a simple cylinder forearm matching the current arm — cheaper, keeps the proven glove pipeline, and sidesteps the 25AE rigging/anchoring traps. Whichever lands first, follow-ups (pose switching, point-on-hover for menus per #1012) stack on top.

tommy-xr:
Superseded by the glove route shipped in #1019 (bare-skin reskin of the proven glove rig + the game's own FISTCOMP cuff as the sleeve, hand hidden while a weapon is wielded). Owner's rationale for the route change: upcoming interactions (climbing, holding, procedural auto-grab poses) want full programmatic pose control, which the glove rig gives us and static authored hands don't.

What this issue produced still stands and is on main: the pawn-local eye-offset frustum diagnosis and the `/v1/scene` `ndc`/`origin_on_screen` fields — both were load-bearing for verifying #1019.

The 25AE authored hand models remain a possible future upgrade if/when hand rigging lands — reopen or file fresh against that work rather than tracking it here.

PR #975, opened 2026-08-14
fix(vr): restore released inventory item physics

### LINKED ISSUE #974: fix(vr): restore world presence after re-releasing inventory items
## Summary

Releasing a previously contained `HasRefs(false)` inventory item after a world re-grab can leave the live entity with no hand ownership, no `Contains` link, and no physics body. The item remains in ECS/save data but cannot be targeted, picked up, or frobbed in production VR, and save/load does not restore its world presence.

This blocked MedSci progression after the sole R&D Card was stranded. It is distinct from #583 / PR #743: that PR makes key cards activate on the authentic container click and avoids this workaround, while the generic drop invariant remains broken for other items and release sequences.

## Exact reproduction

Campaign: 25th Anniversary assets, `medsci2.mis`, `--vr`, exact stacked HEAD `9698555b`, seed `976894603`.

1. Open Watts' Desk (mission object 503) through its physical VR container panel.
2. Move the R&D Card (mission object 772; runtime 1017 in this run) into the backpack.
3. Open the Quest-X backpack and squeeze the exact card slot into the left hand.
4. Release the card into the world.
5. Re-grab that physical card, carry it clear of the desk, and release it again.
6. Try to target, grab, or frob the released card; save and reload, then retry.

Preserved trace proves the second grab/carry/release was production input:

- frame 5343: `wielded_entity_id=1017`, inventory location `left_hand`
- frame 5351: exact card still held after carrying clear
- frame 5359: release clears hand/inventory ownership
- frame 5391 onward: entity persists at `[39.93884, -3.73978, 36.10747]`, with no `Contains` link and zero physics bodies

Normal grab/trigger rays then hit only the room sensor. `rdcard` stayed absent, `Note_2_5` stayed incomplete, and modules stayed 17. A named save/load preserved the zero-body stranded state in an isolated exact-HEAD reproduction.

## Expected

Releasing a live held item restores a selectable world presence: residual container ownership is removed, `PropHasRefs(true)` is present, authored physics is recreated, and the item can be ray-selected/re-grabbed immediately and after save/load.

## Actual

The entity becomes handless, uncontained, and bodyless. Production VR has no recovery path. For the sole MedSci2 R&D Card, progression is blocked because MedSci2 has no region-4 reader or alternate card.

## Acceptance

- Add a negative-first regression using a contained `HasRefs(false)` `MOVE` item.
- Exercise container -> hand -> world drop -> world re-grab -> carried move -> second drop.
- Assert the live released entity has no residual `Contains`, has `PropHasRefs(true)`, has a physics body, and is selected by a production-equivalent VR ray.
- Save/load and assert it remains physically selectable.
- Preserve ordinary container transfer, backpack retrieval, weapons, scripted consumables, and key-card behavior.
- Validate in the real MedSci2 Watts desk/R&D Card path under `--vr`.

## Evidence

Run directory: `.claude/skills/play-through/runs/earth-ops-vr-seed-976894603-iter-3-medsci2-replay-02/` (`pt132`–`pt145`; exact-schema report will be sealed after triage).

Frozen campaign runtime/save were not mutated during review. Independent diagnostic runtimes were shut down.

tommy-xr:
Fixes #974

## Summary

- Restore every live hand-released item to world ownership before authored physics recreation.
- Clear residual incoming `Contains` links and insert-or-replace `PropHasRefs(true)` without changing `StoreItem` container drops.
- Add a MedSci2 production-VR carry-cycle guard covering Desk 503 / Gameboy 72 through two releases and save/load.
- Expose live `HasRefs` in entity debug detail so the E2E asserts the actual persisted invariant.

## Root cause

`VirtualHandEffect::DropItem` called `make_physical` while a formerly contained live entity could still carry `HasRefs(false)` or a residual `Contains` link. Authored physics creation correctly declines non-world entities, so clearing the hand could strand the live object with no owner and no selectable body.

The release path now restores the world-ownership invariant first, then recreates authored physics. `StoreItem` remains the separate legitimate-container path and still writes `HasRefs(false)`.

## Scope note after the stack rebuild

This PR was originally cut on top of the earlier #743 keycard branch, and its evidence used R&D Card 772. It now sits directly on #968, and #743 is an independent PR against `main` where a key source registers on the keyring and is consumed — a keycard never reaches this hand/world path at all. The E2E was therefore reworked onto the Gameboy (obj 72), an ordinary physical prop sharing Desk 503, and no longer asserts anything about keycards.

Measured honestly on this base, the reworked scenario **also passes without the fix**: the `GrabEntity` path already restores `HasRefs` and drops `Contains` for every case a real gesture can reach here, so the E2E is a standing guard on the release invariant rather than a failing-before reproduction. The three `restore_live_entity_world_refs` unit tests are what pin the behavior directly (including refusing to resurrect a consumed entity, which the drop path can otherwise be handed after the item is gone).

## Production-VR coverage

The deterministic MedSci2 scenario performs:

1. physical Desk 503 frob and exact Gameboy-72 panel click;
2. physical Quest-X backpack squeeze;
3. first world release;
4. production world re-grab and carried move;
5. second release;
6. required production-equivalent ray selection;
7. named save/load and the same body/ref/link/ray assertions.

After both the second release and reload, the exact live item has one enabled authored entity body, `HasRefs(true)`, no incoming/outgoing `Contains`, and is selected by the production ray.

## Validation

- focused unit regressions: 3 passed (restore refs/remove containment, preserve legitimate container transfer, do not resurrect a consumed item)
- `cargo test -p shock2vr` — 742 passed
- reworked MedSci2 production-VR E2E — pass in ~21s
- `npm test` in `tools/shock2-sdk` — 26 passed

## Visual evidence (captured on the original base)

MedSci2 second release and save-load

<sub>Captured before the stack rebuild, on the earlier #743-based parent where the released card was stranded. It illustrates the failure mode the invariant guards against; it is not a capture of this branch's current base.</sub>

### Before → after

Before and after: missing zero-body item versus selectable authored item

<sub>Before: the released live object is absent above the desk. After: its authored body is visible and ray-selectable after the same release stage. [Hosted media and capture note](https://gist.github.com/tommy-xr/60118a63bff5ece5a52a0b35d222257f).</sub>

### DIFF
--- a/shock2vr/src/mission/mission_core.rs
+++ b/shock2vr/src/mission/mission_core.rs
@@ -440,6 +440,109 @@ fn move_live_entity_into_container(
     was_able_to_drop
 }
 
+/// Restore a live hand-released item to ordinary world ownership.
+///
+/// Inventory items may have an authored `HasRefs(false)` and can retain a
+/// stale `Contains` link across hand/world transitions. Physics creation is
+/// intentionally performed only after this invariant is restored: entity
+/// creation treats an unreferenced object as non-world state and otherwise
+/// declines to give it a selectable body.
+fn restore_live_entity_world_refs(world: &mut World, entity_id: EntityId) -> bool {
+    let is_alive = world
+        .borrow::<EntitiesView>()
+        .is_ok_and(|entities| entities.is_alive(entity_id));
+    if !is_alive {
+        return false;
+    }
+
+    {
+        let mut links_view = world.borrow::<ViewMut<Links>>().unwrap();
+        for links in (&mut links_view).iter() {
+            drop_contains_links_to(links, entity_id);
+        }
+    }
+    world.add_component(entity_id, PropHasRefs(true));
+    true
+}
+
+#[cfg(test)]
+mod released_item_world_refs_tests {
+    use super::*;
+
+    fn contains(container: &Links, item: EntityId) -> bool {
+        container.to_links.iter().any(|link| {
+            matches!(link.link, Link::Contains(_))
+                && link.to_entity_id.map(|wrapped| wrapped.0) == Some(item)
+        })
+    }
+
+    #[test]
+    fn hand_release_restores_refs_and_clears_residual_containment() {
+        let mut world = World::new();
+        let item = world.add_entity(PropHasRefs(false));
+        let container = world.add_entity(Links {
+            to_links: vec![ToLink {
+                link: Link::Contains(4),
+                to_entity_id: Some(WrappedEntityId(item)),
+                to_template_id: 0,
+            }],
+        });
+
+        assert!(restore_live_entity_world_refs(&mut world, item));
+        assert!(
+            world
+                .borrow::<View<PropHasRefs>>()
+                .unwrap()
+                .get(item)
+                .unwrap()
+                .0
+        );
+        assert!(!contains(
+            world
+                .borrow::<View<Links>>()
+                .unwrap()
+                .get(container)
+                .unwrap(),
+            item,
+        ));
+    }
+
+    #[test]
+    fn ordinary_container_transfer_still_removes_world_refs() {
+        let mut world = World::new();
+        let item = world.add_entity(PropHasRefs(true));
+        let container = world.add_entity(Links::empty());
+
+        assert!(move_live_entity_into_container(&mut world, container, item));
+        assert!(
+            !world
+                .borrow::<View<PropHasRefs>>()
+                .unwrap()
+                .get(item)
+                .unwrap()
+                .0
+        );
+        assert!(contains(
+            world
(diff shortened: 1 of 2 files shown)


tommy-xr:
I'm not sure if this is valid - the card should've disappeared when frobbed _or grabbed_ (not ended up in the backpack) - and added to the user's register of cards

tommy-xr:
Owner decision on the open question here: **keycards should register-and-vanish** when picked up (added to the access-card register, world entity destroyed) — not remain as physical items in the backpack. #743 is being reworked to those semantics; when this PR is recascaded onto the restructured stack, its e2e's "card retained in backpack" assertions need rewriting to assert register+vanish instead. The 5-line released-item-physics fix itself is unaffected.

PR #983, opened 2026-08-14
fix(oculus): service Android events during startup

### LINKED ISSUE #910: Startup window is unserviced by the Android event pump (launch-time ANR / black screen)
The per-frame Android event pump added in `fix(oculus): pump Android events to prevent ANR`
lives inside the OpenXR main loop, so it only covers **steady state**. The window between
process start and the first iteration of that loop — `Game::init` (asset/mission load) — is
still completely unserviced, and it is multi-second on device.

### Evidence

While the app was loading on a Quest 3, logcat recorded:

```
am_anr: ... Input dispatching timed out ... Waited 5000ms for KeyEvent
```

This happens *during* `Game::init`, i.e. before the main loop (and therefore before the
pump) exists. It is user-visible in two ways:

- launch-time "shock2quest isn't responding" dialogs, and
- **black-screen manual launches**: pressing controller buttons while the app loads queues
  KeyEvents that nothing drains, so the platform decides the app is wedged.

Reproduced twice on Quest 3.

### Candidate approaches

1. **Periodic pump callback from within init** — thread a "service the platform" callback
   through the load path and call it at load milestones. Simplest, but requires load code to
   cooperate.
2. **Move init off the render thread** — currently blocked: `Game` is not `Send`.
3. **Attach the input queue to a helper thread's looper during load** — the helper drains
   (and finishes-as-unhandled) input while the render thread loads, then hands the queue
   back. Keeps the load path untouched.

### Scope note

The merged pump fixes the steady-state ANR only; this issue tracks the startup window.

tommy-xr:
## Summary

- register the existing bounded Android looper/input pump on the NativeActivity thread before the synchronous initial game load
- service that hook at game/mission load phase boundaries and every 32 cells or entity instantiations
- keep the hook thread-local so background mission parsing cannot access the Android main thread's looper
- add negative-first coverage for hook invocation and worker-thread isolation

## Reproduction and root cause

On the frozen issue base (`e09fc084e2b528da3639932cfa486dce310188c8`), the Oculus runtime called `Game::init` before the first steady-state `android_pump_events` call. That leaves NativeActivity lifecycle and input queues unserviced through the synchronous first mission load, matching the issue's Quest ANR evidence.

The hook test was added first and failed on the base with `E0433: could not find platform in crate engine`. The implementation makes that test pass and keeps the callback scoped to the registering thread.

Every aggregating startup loop that can grow with mission size now has bounded checkpoints (32 cells/entities). The remaining one-shot load phases are bracketed with checkpoints; observed desktop load instrumentation kept the largest one-shot phase below one second.

## Validation

- `RUSTFLAGS='-D warnings' cargo check -p shock2vr -p desktop_runtime -p debug_runtime`
- `cargo test -p engine -p dark -p shock2vr --lib` — 47 + 110 + 665 passed after rebasing current `main`
- current-main SDK fast suite — 26 passed, 229 skipped, 0 failed
- current-main focused mission-load E2E — all 23 missions passed, including `shodan.mis`
- full serial SDK E2E on the frozen base — 249 tests: 241 passed, 1 failed, 7 skipped; all 23 mission loads, security/SHODAN, transitions, VR interaction, and weapon cases passed. The sole failure was the unrelated creature-loot reader assertion addressed by #971.
- [Android APK CI](https://github.com/tommy-xr/shock2quest/actions/runs/31813676996) — passed
- macOS, Ubuntu, Windows, and formatting CI — passed
- `git diff --check`

Six commits landed on `main` during the full E2E run. They were audited individually: only #970 and #977 touch the same files, in disjoint scene-swap and flat-melee-update hunks. Since this hook is unset/no-op in the debug runtime, the current-main all-mission scenario is the focused rerun that directly re-exercises the changed parse/build/init paths.

Fresh Quest validation was not possible: this host has no `adb`, its configured Android SDK/NDK paths are absent, and `cargo-apk` is not installed.

## Visual evidence

This changes event servicing/timing only and does not alter rendered content. The headless debug runtime cannot exercise Android's looper or the OS ANR dialog, so an identical-frame GIF would be misleading. Objective evidence is the negative-first hook/thread-affinity tests, exact load call-order coverage above, scoped builds/tests, and Android CI.

Fixes #910

### DIFF
--- a/dark/src/mission/mod.rs
+++ b/dark/src/mission/mod.rs
@@ -116,6 +116,8 @@ pub struct TextureSize {
 }
 
 const LIGHTMAP_SIZE: u32 = 4096;
+/// Keep a long synchronous mission parse responsive to host-platform queues.
+const LOAD_EVENT_PUMP_CELL_INTERVAL: usize = 32;
 
 pub struct UVCalculationInfo {
     origin: Vector3<f32>,
@@ -147,6 +149,7 @@ pub fn read<T: io::Read + io::Seek>(
     links_with_data: &Vec<Box<dyn LinkDefinitionWithData>>,
     properties: &Vec<Box<dyn PropertyDefinition<T>>>,
 ) -> SystemShock2Level {
+    engine::platform::service_events();
     let table_of_contents = ss2_chunk_file_reader::read_table_of_contents(reader);
 
     let wr_ext = table_of_contents.has_chunk("WREXT".to_string()); // Extended representation
@@ -191,11 +194,16 @@ pub fn read<T: io::Read + io::Seek>(
     let mut cells: Vec<Cell> = Vec::new();
     let mut packer = TexturePacker::<image::Rgb<u8>>::new_rgb(LIGHTMAP_SIZE, LIGHTMAP_SIZE);
     for cell_idx in 0..wr_num_cells {
+        if cell_idx as usize % LOAD_EVENT_PUMP_CELL_INTERVAL == 0 {
+            engine::platform::service_events();
+        }
         let cell = Cell::read(reader, &mut packer, wr_ext, cell_idx, light_size);
         cells.push(cell);
     }
+    engine::platform::service_events();
 
     let bsp_tree = BspTree::read(reader, &cells);
+    engine::platform::service_events();
 
     if wr_ext {
         let _ = read_bytes(reader, wr_num_cells as usize);
@@ -217,6 +225,7 @@ pub fn read<T: io::Read + io::Seek>(
         properties,
         reader,
     );
+    engine::platform::service_events();
 
     let textures = TextureList::read(
         &table_of_contents,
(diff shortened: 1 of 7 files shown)


tommy-xr:
## On-device verification (Quest 3)

Ran this branch (d2015ac2, release APK) on a Quest 3 (`eureka`), mission `medsci1.mis`, two cold launches with clean logcat each time:

| | Launch 1 | Launch 2 |
|---|---|---|
| `SHOCK2QUEST_STARTUP init_ms` | 5282.458 | 3924.744 |
| Reached `FOCUSED`, steady ~90 fps, 0 skipped | ✅ | ✅ |
| `am_anr` in `logcat -b events` | none | none |
| "not responding" / "ANR in" (main buffer) | none | none |
| `android_pump_events: no ALooper for this thread` | absent (pump active) | absent (pump active) |

- The load window really is multi-second on device (~4–5 s), i.e. squarely inside the territory where the unserviced startup previously tripped the 5 s input-dispatch ANR.
- Launch 2 additionally injected 40 `adb shell input keyevent` presses concurrently with the load window to pressure the input queue — still no ANR.
- No panics, no crash output, no black-screen indications; app stayed `topResumedActivity` through a 30 s settle.

Caveats: `adb` key injection goes through the system dispatcher to the focused window, so it exercises the NativeActivity input queue but isn't identical to physical controller presses (the exact #910 repro); and no pre-fix baseline was run in this session — this confirms the fixed build launches clean with the pump demonstrably active, on a device where #910 had reproduced twice before the fix.

Follow-up to retire the `service_events` call sites via async/time-sliced loading: #995.

PR #980, opened 2026-08-14
fix(gameplay): restore saved melee weapon damage

tommy-xr:
## Summary

- resolve a held melee weapon's authored contact stims through its canonical archetype before falling back to its concrete mission id
- keep the existing trigger-gated, one-hit-per-pull VR melee latch unchanged
- cover the exact MedSci campaign path: corpse Wrench 990, save/load, deck transition, physical VR hits, another save/load, and a destructible-world control

## Root cause

The reviewed campaign save correctly restored Wrench 990 with canonical archetype -928, but `TriggeredMeleeWeapon` looked up contact stims only by the concrete `PropTemplateId` (990). The gamesys-wide contact-stim table is keyed by archetype -928, so the restored weapon still made physical contact and shoved creatures while emitting no `Damage`.

## Verification

- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`
- `cargo test -p shock2vr` (742 passed)
- `npm test` in `tools/shock2-sdk` (26 passed, E2E skipped)
- focused saved-campaign VR E2E, repeated on the finalized test (fresh -928 control; Wrench 990 corpse loot; save/load; MedSci2; second save/load; Window2 destruction)
- per-target damage assertions are per *pull*, not per kill: for Monkey 543, OG-Pipe 1293, and OG-Shotgun 1392 the test asserts the authored starting HP, exactly one `Damage` message per trigger pull, and exactly 9 HP lost during the contact pose. It does not swing a target to death, so it makes no kill-count claim.
- existing VR melee + save/load E2Es (5 passed)

## Visual verification

saved campaign Wrench VR comparison

<sub>Same copied frontier, Wrench 990, Monkey 543, teleport, target, trigger edge, and physical 4.0 -> 0.33 hand sweep on exact base and fixed builds. Captured headlessly via deterministic `/v1/step` + `/v1/screenshot` requests.</sub>

fixed saved Wrench contact

### Before -> after

saved Wrench before and after

The exact #755 base physically displaces Monkey 543 but leaves HP at 10; the fixed build emits one `Damage` and changes HP 10 -> 1. Both runs loaded a SHA-matched private copy of the reviewed campaign frontier; the shared save remained read-only.

## Stack

Depends on #755 (`fix/medpatch-healing`, exact head `d2f6eae6`), which in turn sits on #975 -> #968 -> #967.

Fixes #978

PR #989, opened 2026-08-14
fix(ai): let solid entity cover block sight

### LINKED ISSUE #985: fix(ai): let solid entity-backed walls occlude monster sight
## Summary

Monster sight raycasts only against the WORLD collision group, while ranged line-of-fire raycasts against WORLD, ENTITIES, and SELECTABLE. A solid wall, door, or structural prop represented by an entity collider can therefore remain transparent to perception while correctly blocking the projectile.

The result is inconsistent but player-observable: AITargetVisible remains true and AILastKnown tracks the live player through solid cover, keeping the AI at High awareness and following the current player position. Its shot is still suppressed or absorbed by the same cover because the separate line-of-fire gate sees the entity collider.

This is systemic across presentations. The reviewed occurrence was nonblocking because the wall remained projectile-safe and the hostile could be lured away from the required route.

## Reviewed reproduction

Configuration: 25th Anniversary assets, forced VR, stacked HEAD 40885d86c7f4ee8f980249eec02d459ce7c6ef7c, continuous MedSci2 replay07.

1. A normal ecology Blue Monkey is alerted on the south catwalk.
2. Move the player from the exposed corridor to [33.0,-0.356,20.1], deep west behind open Door884 and several solid wall faces. The Monkey is around [43.028,-0.701,-2.345].
3. Remain fully behind the wall for 57 advancing frames.
4. Observe no player Damage. The ranged attack does not produce a damaging clear shot through the cover.
5. Inspect the live awareness state: AITargetVisible remains true throughout and AILastKnown changes every time the hidden player moves, rather than freezing at the last exposed position.
6. The Monkey remains High/RangedAttack and turns or pursues the live through-wall position.

Local reviewed evidence:

- .claude/skills/play-through/runs/earth-ops-vr-seed-976894603-iter-3-medsci2-replay-07/pt399_true_hard_cover.png
- .claude/skills/play-through/runs/earth-ops-vr-seed-976894603-iter-3-medsci2-replay-07/pt400_deep_west_cover.png
- .claude/skills/play-through/runs/earth-ops-vr-seed-976894603-iter-3-medsci2-replay-07/pt402_monkey_shot_wall.png
- .claude/skills/play-through/runs/earth-ops-vr-seed-976894603-iter-3-medsci2-replay-07/pt403_monkey_attack_absorbed.png

This was not authored GotoLoc or debug-pinned awareness. The MedSci2 random ecology generators that produce Blue Monkey, mission objects1161 and1210, have POP_LIMIT/PLAYER_DISTANCE/SELF_MARKER/RAYCAST flags and no GOTO_LOCATION. Only GOTO_LOCATION spawns receive the runtime pinned SetAlertness path.

The campaign runtime and save were not mutated by review.

## Root cause

shock2vr/src/scripts/ai/ai_util.rs uses different collision masks:

- is_player_visible and is_player_visible_in_fov cast only against InternalCollisionGroups::WORLD.
- has_line_of_fire casts against WORLD | ENTITIES | SELECTABLE and resolves proxy entities.
- AnimatedMonsterAI publishes AITargetVisible directly from is_player_visible_in_fov and refreshes last_known_player_pos every visible frame.

The source already carries a TODO to check entities while allowing genuinely transparent surfaces. Until that classification exists, entity-backed structural cover is invisible to sight but solid to bullets.

AITargetVisible is not a stale coarse property: RuntimePropAITargetAwareness is republished whenever the current visibility or last-known position changes. The continuously changing last-known coordinate proves a fresh visible result each frame.

No exact duplicate was found. #481 and its follow-ups cover chase/path stalls and line-of-fire attack gating, not perception through entity-backed cover.

## Expected

Opaque solid entity-backed walls and closed doors should break monster sight exactly as brush WORLD geometry does. Once sight breaks, last-known position should freeze at the last genuinely visible point and normal hidden-time decay/Search behavior should proceed. The AI must still be able to see through surfaces authored as visually transparent when appropriate.

## Acceptance

- Add a negative-first visibility test with player and monster separated by a solid ENTITY or SELECTABLE collider: is_player_visible_in_fov returns false.
- Cover at least a closed standard door and an entity-backed structural wall or large opaque prop.
- Keep WORLD brush occlusion working.
- Define one shared sight-occluder classification that resolves proxies, ignores the observer and target, and distinguishes opaque cover from authored transparent glass rather than blindly treating every selectable icon proxy as a wall.
- In a real-runtime AI scenario, expose the player, move behind solid entity cover, and assert AITargetVisible true -> false, AILastKnown freezes, and no fresh player coordinate leaks through subsequent hidden movement.
- Re-expose the player and assert awareness resumes normally.
- Preserve has_line_of_fire projectile safety, ally obstruction, open-door behavior, psi invisibility, and camera/turret visibility semantics.
- Keep path pursuit toward the frozen last-known point and normal alertness decay/Search transitions intact.

tommy-xr:
## Summary

- make monster/camera/turret sight use an actor-membership ray across world, entity, and selectable collision groups
- classify entity hits through their visible proxy owner so opaque solid cover blocks sight while authored alpha, non-rendered helpers, and interaction-only bounds pass through
- **stop creatures from occluding sight, in any state**: a packmate's capsule hid the player from the AI behind it, so a pack closing on the player blinded everything but its front rank and sight ended up stricter than line of fire
- keep projectile line-of-fire stricter and unchanged: transparent physical glass and small props still stop shots, and a living ally still suppresses the shot
- correct the `ray_cast2_as_actor_with_entity_filter` doc comment: the filter is skipped for *ownerless* colliders, not for world geometry as such
- add negative-first unit coverage plus authored MedSci2 VR scenarios for the south-catwalk door and UBWindow

## Creatures never occlude

The original engine's sight cast tests world geometry plus an explicit list of vision-blocking objects (doors and anything authored to block AI vision); creatures are never in it.

This is decided in `entity_occludes_sight` rather than by dropping `ACTOR` from the collision mask. A corpse keeps its creature identity but swaps `ACTOR` membership for `SELECTABLE`, so a mask-level exemption would only cover the living — an AI would see through a packmate standing up and be blinded by the same body a second later. One rule, every state.

## Before → after

AI entity-cover sight before/after

<sub>Same MedSci2 VR sequence on PR #980 and this fix: the baseline refreshes the live hidden target through the closed entity-backed door; the fixed query publishes no through-cover target. Captured headlessly via deterministic fixed-timestep `/v1/step` + `/v1/screenshot`.</sub>

Labeled before/after still

## Multi-point sight sampling: attempted, then backed out

Sight is still a single ray from the observer's origin to the player's origin. Multi-point sampling was implemented on this branch and **removed again**, because it is a real regression rather than a refinement, and it belongs in its own PR.

What was tried: rays starting at the observer's eye (placed at the same fraction of body height as the player's, from the observer's own authored body size) and aimed at several points on the player — body centre and eye height, each with lateral offsets taken perpendicular to the sight direction, sized from the player's live collider so they shrink when crouching. Visible if any ray is clear. That is a superset of the original engine's geometry, which raises the ray by an authored per-AI eye offset and aims at the player's head rather than the body origin.

Why it was backed out — measured, not assumed:

- `chase-last-seen.e2e.test.ts` and `search-behavior.e2e.test.ts` both **pass on the base commit and fail with sampling enabled**: the chasing AI abandons the last-seen position (1.44 → 8.64 units) and the searcher never begins its search.
- The failures are unchanged when the lateral offsets are halved, and unchanged when the observer eye offset is disabled entirely — so it is the **target-side** sampling, not the eye placement or the lateral width. The authored "no line of sight" pockets these scenarios rely on hide the player's body origin but not their head.

That is a genuine design question — either those pockets need re-authoring or the sampled set needs to be narrower than the player's real silhouette — and it should be settled with the scenarios in hand rather than tuned blind. Filed as follow-up work; this PR ships the two changes that are unambiguously correct.

Also noted while reviewing, deliberately out of scope: `has_line_of_fire` starts at the object origin rather than an eye, so sight and line of fire do not share one viewpoint.

## Validation

Negative-first — each fails with just its own change backed out:

- `opaque_solid_entity_occludes_sight` failed on PR #980 head `dd8e5030` because the entity collider was ignored
- `a_living_creature_does_not_occlude_sight_but_does_block_the_shot` fails without the creature exemption, and asserts the other half too: a living ally still suppresses the shot
- `a_creature_corpse_does_not_occlude_sight_either` is a forward-looking guard (a corpse is already unreachable by an actor-membership ray today, so it passes for two reasons — it catches a future regrouping that puts bodies back in the ray's path)

Plus:

- `cargo test -p shock2vr` — 725 passed
- `cargo test -p shock2vr scripts::ai::ai_util::sight_occlusion_tests` — 9 passed
- targeted e2e after the backout — `chase-last-seen`, `search-behavior`, `ai-reacquire`, `ally-line-of-fire`, `shodan-corpse-drift` — 7 passed, 0 failed
- full `npm run test:e2e` was run on the sampling version to catch the regression above
- rebased onto `main` @ `245cd692`
- reviewed by two independent adversarial reviewers; the corpse/living inversion they found is the reason the exemption lives in `entity_occludes_sight` rather than in the collision mask

Campaign configuration: earth-to-ops custom · Navy · 25th Anniversary · VR · seed 976894603.

Fixes #985

### DIFF
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -4443,8 +4443,11 @@ impl PhysicsWorld {
     }
 
     /// Raycast with living-actor membership while allowing the caller to
-    /// reject individual entity colliders. World geometry (which may not have
-    /// an ECS owner) is always retained. This is the visibility-query path:
+    /// reject individual entity colliders. The filter is consulted only for
+    /// colliders that carry an entity id, so ownerless colliders - world
+    /// geometry among them - are retained without being offered to it; an
+    /// entity-owned collider is subject to the filter whatever it represents.
+    /// This is the visibility-query path:
     /// actor membership naturally skips interaction-only bounds that do not
     /// block creatures, while the entity filter can pass through authored
     /// transparent objects without changing projectile or selection rays.
--- a/shock2vr/src/physics/mod.rs
+++ b/shock2vr/src/physics/mod.rs
@@ -4414,6 +4414,7 @@ impl PhysicsWorld {
             collision_groups,
             entity_to_ignore,
             ignore_sensors,
+            None,
         )
     }
 
@@ -4437,6 +4438,35 @@ impl PhysicsWorld {
             collision_groups,
             entity_to_ignore,
             ignore_sensors,
+            None,
+        )
+    }
+
+    /// Raycast with living-actor membership while allowing the caller to
+    /// reject individual entity colliders. World geometry (which may not have
+    /// an ECS owner) is always retained. This is the visibility-query path:
+    /// actor membership naturally skips interaction-only bounds that do not
+    /// block creatures, while the entity filter can pass through authored
+    /// transparent objects without changing projectile or selection rays.
+    pub fn ray_cast2_as_actor_with_entity_filter(
+        &self,
+        start_point: Point3<f32>,
+        direction: Vector3<f32>,
+        max_toi: f32,
+        collision_groups: InternalCollisionGroups,
+        entity_to_ignore: Option<EntityId>,
+        ignore_sensors: bool,
+        entity_filter: &dyn Fn(EntityId) -> bool,
+    ) -> Option<RayCastResult> {
+        self.ray_cast2_with_memberships(
+            InternalCollisionGroups::ACTOR,
+            start_point,
+            direction,
+            max_toi,
+            collision_groups,
+            entity_to_ignore,
+            ignore_sensors,
+            Some(entity_filter),
         )
     }
 
@@ -4449,6 +4479,7 @@ impl PhysicsWorld {
         collision_groups: InternalCollisionGroups,
         entity_to_ignore: Option<EntityId>,
         ignore_sensors: bool,
+        entity_filter: Option<&dyn Fn(EntityId) -> bool>,
     ) -> Option<RayCastResult> {
         // Guard against degenerate rays. A zero-length direction normalizes to
         // NaN, and a NaN/zero ray direction sends parry's `clip_aabb_line` down
@@ -4493,6 +4524,11 @@ impl PhysicsWorld {
             let data = collider.user_data;
             let maybe_entity_id = EntityId::from_inner(data as u64);
             if maybe_entity_id == entity_to_ignore {
+                return false;
+            }
+            if let (Some(entity_id), Some(entity_filter)) = (maybe_entity_id, entity_filter)
+                && !entity_filter(entity_id)
+            {
                 return false;
             }
             // A degenerate collider (zero-extent / non-finite AABB, e.g.
--- a/shock2vr/src/scripts/ai/ai_util.rs
+++ b/shock2vr/src/scripts/ai/ai_util.rs
@@ -846,6 +846,20 @@ fn resolve_proxy_entity(world: &World, entity_id: EntityId) -> EntityId {
 /// surfaces that remain, fully rendered objects occlude while authored alpha
 /// (the Windows archetype is 0.45) and non-rendered helpers pass sight. Proxy
 /// hitboxes are classified through their visible owner.
+///
+/// Creatures never occlude, in any state. A packmate's capsule must not hide
+/// the player from the AI behind it - a pack closing on the player would blind
+/// everything but its front rank, leaving sight stricter than line of fire -
+/// and the same has to hold once that packmate dies, or an AI would see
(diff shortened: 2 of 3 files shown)


tommy-xr:
Wondering:
- Should the ai raycast check only look at WORLD | SELECTABLE, vs creatures?
- Should we do a couple raycasts (head -> center, head -> corners, or at least head -> midpoint(center, corners))?

PR #1013, opened 2026-08-17
fix(menu): make Quit work on every runtime

tommy-xr:
The main menu's Quit already produced `Effect::GlobalEffect(GlobalEffect::Quit)` -> `Game::should_quit`, but only the desktop runtime ever observed it. On Quest the item did nothing at all.

## What changed

**Oculus (`runtimes/oculus_runtime/src/lib.rs`)** — the actual fix. Once `game.should_quit()` latches, the loop asks OpenXR to exit (`session.request_exit()`, exactly once) and lets the *existing* session-state machine do the teardown: the runtime drives us READY -> STOPPING (where the handler already calls `session.end()`) -> EXITING, and the EXITING arm already breaks `'main_loop`. The request is issued outside the wait/begin/end frame sequence, so the request-failure fallback (break out and finish anyway) can't strand an `xrWaitFrame` with no matching `xrBeginFrame`.

After the loop: ndk-glue runs `main` on a thread it spawned, so returning is not enough — the NativeActivity and the process would linger with nothing rendering (a black void in the headset). So we finish the activity, keep pumping Android's queues until it reports `Destroy` (1s upper bound), release the XR objects explicitly (`process::exit` runs no destructors), and end the process.

**Debug runtime** — still quit-immune on purpose: an automation session must not kill itself. It now *reports* the request as `quit_requested` on `GET /v1/info` (mirrored in the SDK's `FrameSnapshot`), which is what makes the whole path assertable headlessly.

**Desktop** — no change; verified by reading. `should_quit` -> `window.set_should_close(true)` sits in the single `while !window.should_close()` loop, and `--vr` only switches `PresentationMode`, so both desktop presentations already exit.

## Tests

New `tools/shock2-sdk/test/menu-quit.e2e.test.ts` (gated `SHOCK2_E2E=1`) boots the menu in the debug runtime and clicks Quit twice over: once through the flat pointer channels, once through the VR hand ray (`--vr`), asserting `/v1/info` flips `quit_requested` to true **and** the runtime is still serving. Negative-first: without the `/v1/info` change the field is `undefined` and both tests fail (verified). The duplicated menu/VR-panel geometry the two menu tests shared now lives in `test/helpers/frontend-menu.ts`.

No Rust unit test was added: the `MenuAction::Quit` resolution is already covered by `rising_edge_over_quit_activates_it` / `the_left_hand_can_drive_the_menu_too`, and the effect mapping is a one-line match arm the e2e now exercises end to end.

## Verification

- `cargo fmt --all --check` clean
- `cargo test -p shock2vr --lib` — 740 passed, 0 failed
- `cargo apk check` (android, `-D warnings`) clean
- e2e: `menu-quit` (2) + `menu-audio` (3) — 5 passed, 0 failed
- `/xreview` (Claude + Codex + a de-dup pass): 0 Critical/High. Both engines agreed on two Medium teardown issues (exit requested inside the frame sequence; `process::exit` skipping destructors / not detecting `Destroy`) and on dropping a tautological menu refactor and the test duplication — all four are fixed in the second commit.

**Not device-verified.** No adb device was touched for this change, so the Quest exit path is implemented and compiles but has not run on hardware. It can be verified later with the on-device debug server: trigger the Quit click remotely and watch the process disappear from `adb shell ps`.

No screenshots — nothing visual changed.

### DIFF
--- a/runtimes/debug_runtime/src/commands.rs
+++ b/runtimes/debug_runtime/src/commands.rs
@@ -685,6 +685,10 @@ pub struct FrameSnapshot {
     pub mission: String,
     /// True once the retail finale has entered its terminal ending cutscene.
     pub campaign_completed: bool,
+    /// True once a scene has asked the runtime to quit (the main menu's Quit).
+    /// The debug runtime deliberately stays up - an automation session must not
+    /// kill itself - so this is how the quit path is observed headlessly.
+    pub quit_requested: bool,
     pub player: PlayerInfo,
     pub entity_count: usize,
     pub debug_features: Vec<String>,
--- a/runtimes/debug_runtime/src/main.rs
+++ b/runtimes/debug_runtime/src/main.rs
@@ -1890,6 +1890,7 @@ fn capture_frame_snapshot(
         },
         mission: game.scene_name().to_string(),
         campaign_completed: game.campaign_completed(),
+        quit_requested: game.should_quit(),
         player: {
             // Real player state from the active world (position, look, and the
             // held/wielded entities), or zeros when the scene has no player.
--- a/runtimes/oculus_runtime/src/lib.rs
+++ b/runtimes/oculus_runtime/src/lib.rs
@@ -450,6 +450,9 @@ fn main() {
     let mut swapchain = None;
     let mut event_storage = xr::EventDataBuffer::new();
     let mut session_running = false;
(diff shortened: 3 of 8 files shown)


tommy-xr:
Device verification complete (owner, on Quest 3): the menu Quit path exits the app cleanly on hardware. That closes the one item the PR flagged as unverified — the OpenXR request-exit → session teardown → activity finish path works as designed.

Issue #1022, opened 2026-08-17
Quest renders the legacy protocol droid skin instead of the 25AE/SHTUP upgrade


tommy-xr:
On-device (Quest 3, anniversary data layout): the protocol droid renders with the legacy skin, not the upgraded one. Desktop can't catch this class of bug: `~/ss2-data-unpacked` is a classic loose install with no kpf, so every desktop `--vr` capture runs legacy-textured.

**Ruled out — the assets are present and the mounts look right:**
- Device has the full stack: `/sdcard/shock2quest/sshock2.kpf` (434 MB) + all five `mods/*.kpf`.
- The upgrade exists: `shtup.kpf` → `obj/txt16/proto03.dds` (43 KB; also `droid.dds`, `protscrn.dds`); `sshock2ee.kpf` → upgraded `mesh/protonew.bin` (102 KB vs legacy 15 KB).
- The engine mounts the mod stack highest-priority-first (`shock2vr/src/lib.rs` `MOD_ARCHIVES`/`anniversary_family_mounts`), and the loose folder mount is LAST — the device's leftover loose `res/mesh/PROTONEW.BIN` + `res/mesh/txt16/PROTO03.GIF` should not shadow anything.

**Remaining suspects, in order:**
1. **DDS decode**: if `proto03.dds` is an encoding we don't decode (BC7?), the texture load may fail silently and fall back to the legacy GIF. Check what encodings shtup ships vs what the engine's DDS path supports, and make an unsupported-encoding fallback LOUD (log the name + format).
2. **Extension resolution**: the model references `PROTO03` (classically `.GIF`/`.PCX`) — verify the zip-mount lookup actually tries `.dds`/`.png` variants for obj textures (`dark::util::resolve_texture_name` coverage; the txt16 basename-collision gotcha may also apply).
3. **Detection**: confirm `is_25th_anniversary_install()` actually returns true on the device layout (the #991 startup probe logs should say — check logcat).
4. The upgraded `protonew.bin` mesh in `sshock2ee.kpf` may additionally not be picked up by the creature-mesh load path (separate from the skin, same investigation).

[...]

Issue #352, opened 2026-07-04
VR: held weapons use first-person viewmodels (_h meshes with stripped faces) - use world models instead


tommy-xr:
## Problem

When holding/wielding a weapon in VR, the entity's model is swapped to its first-person **view model** (the `_h` mesh) by `InternalSwitchHeldModelScript` (`shock2vr/src/scripts/internal_switch_held_model.rs`) — it changes to the `PropPlayerGun` view model on `Hold` and restores the original on `Drop`.

The `_h` viewmodels are optimized for the flatscreen first-person camera: faces that were never visible from that fixed angle are removed. In VR you can inspect the weapon from any angle, so the missing geometry is visible (holes in the mesh).

## Proposed fix

In VR presentation mode, keep the **world model** when held instead of swapping to the view model, and position/orient the hand gloves relative to the world model. Pieces involved:

- `internal_switch_held_model.rs` — gate the `Hold` model swap on presentation mode (flat keeps the current behavior; its viewmodel path depends on `_h` meshes).
- `shock2vr/src/vr_config.rs` — `HAND_MODEL_POSITIONING` currently keys per-hand grip offsets by model name, mostly `_h` names (`atek_h`, `amp_h`, `wrench_h`, ...). These offsets need re-keying/re-tuning for the world model names.
- `shock2vr/src/hand_glove.rs` (from #351) — the glove hands render at the raw hand transform; once weapons keep world models, the glove pose while holding could switch to a grip pose aligned with the weapon's handle (see `projects/vr-gloves.md` follow-ups).

## Verification

Headless via the debug runtime (`cargo dbgr --vr`): spawn/grab a weapon, orbit the head with `head.look`, screenshot from angles that expose the stripped faces on the `_h` model before, and confirm the world model is intact after.

PR #1023, opened 2026-08-17
feat(vr): wield the 25AE first-person weapon models with fitted grips

tommy-xr:
## VR wields the 25AE first-person weapon models

On a 25th Anniversary install, a VR-grabbed gun now renders its remastered first-person model (`obj/*_h.bin` from `mods/sshock2ee.kpf`) instead of the world model — grip seated in the hand, muzzle vhots adopted for firing. Classic installs are untouched: the classic `_h` meshes really are one-sided flat-camera shells (#352), so without the 25AE archives VR keeps the world-model path. Flat is unchanged in both cases.

VR 25AE weapons

<sub>All eleven guns wielded in VR, cycling. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep) in `debug_weapons --vr`.</sub>

### Before → after

before/after

### Every weapon, as wielded (owner request: per-weapon evidence)

contact sheet

<sub>The small bare hand in each tile is the co-located *reference* hand (left hand parked at the same transform below the grab threshold) marking exactly where the tracked controller is — the grip should land on it. Weapons lying on the floor in some tiles are previously-cycled drops, not part of the wielded model.</sub>

### Pistol firing-hand fix (before → after)

firing hand before/after

<sub>Same wield, same framing: before, the island strip had deleted every `ND-arm_atek.psd` polygon — the pistol's baked firing hand — leaving a holed hand with the pale reference hand showing through; after, the mesh renders complete.</sub>

### What changed

- **Routing** (`shock2vr/src/scripts/internal_switch_held_model.rs`): on `Hold` in VR, if the install is 25AE and the gun's `hand_model` is in the verified set (`vr_config::VR_25AE_VIEW_MODELS`, 11 models), swap to it via `ChangeModel` — same as flat's wield swap; the model's own muzzle vhots come along. Otherwise the existing world-model + vhot-donor path runs. `Drop` restores the world model (already-existing `InternalPropOriginalModelName` path).
- **Grip fitting** (`shock2vr/src/vr_config.rs`): per-model offsets for all 11, hand-fit against an empty-hand reference at a fixed display pose (~720 px/world-unit calibration at the capture plane). The whole 25AE `_h` set is authored barrel-along **−X**, so one shared −90° yaw seats every gun.
- **All baked hand islands kept** (`dark/src/importers/model_importer.rs`, `VR_HELD_MODELS_IMPORTER`): the VR-held model now renders **exactly as authored, baked hands included**. An earlier revision of this PR stripped every arm-material island but the largest to hide spare reload hands — but "one arm island = one hand" is false: `atek_h` splits its **firing hand** (`ND-arm_atek.psd`, its own island) from its sleeve (`ND-arm.psd`), so the strip kept the sleeve and **deleted the pistol's gripping hand** (the defect the contact sheet showed). Spare-hand stripping is deliberately dropped per owner decision — a floating spare hand is the lesser evil vs deleting the real one; it can return later if a reliable classifier turns up and the floating hands bother anyone in practice. The cache-separated importer stays as the seam for per-wield mesh preparation.
- **Bind-pose fix** (`mission_core.rs` ChangeModel): articulated models (FP meshes are hand+arm+gun skeleton sub-objects) rendered unposed/exploded in the world pass; ChangeModel now installs an empty `AnimationPlayer` (bind pose, emits no motion flags/events) for animated models.
- **Flat frozen** (`vr_config::FLAT_WIELD_SWAP_MODELS`): flat's wield-swap filter no longer derives from the (growing) VR grip table — it stays exactly the pre-change set, covered by a unit test. Flat viewmodel render-verified unchanged (pistol + AR15 screenshots vs base).

### Mesh audit: the 25AE `_h` set is closed (classic is not)

`cargo run -p dark --example mesh_audit` over the 25AE `obj/*_h` guns (open edges = edges used by only one polygon):

| model | tris | open edges | (classic `_h` for comparison) |
|---|---|---|---|
| al_h | 880 | **0.0%** | – |
| viro_h | 1036 | **0.6%** | – |
| empgun_h | 1753 | **0.8%** | – |
| amp_h | 1343 | **1.5%** | classic 5.9% |
| atek_h | 2759 | **3.7%** | classic **19.8%** |
| lasehand | 2256 | **4.1%** | – |
| fsn_h (+H1/H2) | 2106 | **4.2–5.1%** | – |
| sg_h | 1535 | **4.3%** | classic **28.0%** |
| sfg_h | 1357 | **4.9%** | – |
| ar15_h | 2378 | **10.5%** | classic **38.3%** |
| gren_h | 957 | 22.6% (visually fine in VR) | – |

### Heading question: resolved by measurement

The prior handoff flagged an unresolved X-axis-gun heading ambiguity in the `_w` set. The `_h` route dissolves it: every 25AE `_h` model is authored barrel-along −X with a real muzzle vhot, so one rotation fits all. Decisive fire test (VR, pistol wielded, hand at world `[-1.9, 2.244, 0.0]`, barrel along hand −Z): bullet impact at `[-11.496, 2.506, 0.018]` — < 0.2° yaw deviation over 9.6 units, and the impact height exactly equals grip offset (+0.12) + muzzle vhot height (+0.139). The projectile leaves the muzzle vhot straight along the rendered barrel. Classic `_w` fallback behavior is unchanged.

### PMNM / #1022 independence

The eleven gun `_h` models are **standalone remastered LGMD meshes — no `PMNM` chunk** — so this path is independent of the high-detail gate (#1022 / PR #1024) and renders remastered on Quest as-is. The **melee** `_h` models (wrench/rapier/shard/psword/pipewrench) are LGMM skinned meshes shipped as classic geometry + appended PMNM chunk; they need a posed-skeleton render path (the PMNM gate itself landed in PR #1024, now on by default), so **VR melee keeps world models in this PR** (recorded follow-up).

### Two-handed weapons (recorded follow-up — out of scope here)

For the second-grip workstream, the `_h` models posed for a support-hand hold: **sg_h** (baked hand on the pump), **empgun_h** (baked hand on the foregrip), **ar15_h** (baked hand at the magwell). **fsn_h / sfg_h / gren_h** bake no hand at all and are natural two-handers. The idle off-hand currently renders as the ordinary bare hand next to them.

### Known gaps / notes

- Psi Sword (`-2291`) resists the harness raycast-grab in `debug_weapons` (its tile shows the empty fist); its wield path is unchanged by this PR (melee fallback).
- `sfg_h` is a bulky device whose grip handle is occluded from the capture angle; its offset places the authored carry handle at the hand.
- The `dark/examples/mesh_audit.rs` scratch tool stays — it is the evidence for the table above and for future model audits.

### Verification

- `cargo fmt --all --check`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean
- `cargo test -p shock2vr --lib`: 814 passed (3 new vr_config tests: every VR view model has a grip entry; flat swap set frozen; case-insensitive lookup)
- e2e: `vr-held-model` **updated** (was the negative test asserting VR keeps `atek_w`; now asserts the `_h` swap + drop-restore, the flat swap unchanged, and that the wielded `atek_h` draws its **full authored mesh** — 7 scene objects; 6 with the old island strip, red before the keep-all-islands fix) — plus `cross-mode-held-load`, `vr-melee-contact`, `medsci-saved-vr-melee`, all `weapon-*`, `energy-weapon-recharge`, and `missions` (23/23 load): **all green**
- Both presentations render-verified: VR per-weapon sweep above; flat pistol/AR15 viewmodel screenshots identical in framing/behavior to base

### DIFF
--- a/dark/examples/mesh_audit.rs
+++ b/dark/examples/mesh_audit.rs
@@ -1,0 +1,66 @@
+//! Scratch tool: report polygon/vertex counts and open (boundary) edges for an
+//! LGMD object `.bin`. A closed solid has every edge shared by exactly two
+//! polygons; a one-sided "viewmodel" mesh leaves many edges used only once.
+
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::{BufReader, Cursor, Read};
+
+fn main() {
+    for path in std::env::args().skip(1) {
+        let mut buf = Vec::new();
+        BufReader::new(File::open(&path).unwrap())
+            .read_to_end(&mut buf)
+            .unwrap();
+        let mut cursor = Cursor::new(buf);
+        let header = dark::ss2_bin_header::read(&mut cursor);
+        let mesh = dark::ss2_bin_obj_loader::read(&mut cursor, &header);
+
+        // Weld positions so an edge shared across sub-objects still matches.
+        let key = |i: u16| {
+            let v = mesh.vertices[i as usize];
+            (
+                (v.x * 4096.0).round() as i64,
+                (v.y * 4096.0).round() as i64,
+                (v.z * 4096.0).round() as i64,
+            )
+        };
+
+        let mut edges: HashMap<(_, _), u32> = HashMap::new();
+        let mut tris = 0usize;
+        for poly in &mesh.polygons {
+            let n = poly.vertex_indices.len();
+            if n < 3 {
+                continue;
+            }
+            tris += n - 2;
+            for i in 0..n {
+                let a = key(poly.vertex_indices[i]);
+                let b = key(poly.vertex_indices[(i + 1) % n]);
+                let e = if a <= b { (a, b) } else { (b, a) };
+                *edges.entry(e).or_insert(0) += 1;
+            }
+        }
+        let open = edges.values().filter(|c| **c == 1).count();
+        let bb = mesh.bounding_box;
+        println!(
+            "  bbox min=({:.3},{:.3},{:.3}) max=({:.3},{:.3},{:.3})",
+            bb.min.x, bb.min.y, bb.min.z, bb.max.x, bb.max.y, bb.max.z
+        );
+        for v in &mesh.vhots {
+            println!(
+                "  vhot {:?} ({:.3},{:.3},{:.3})",
+                v.vhot_type, v.point.x, v.point.y, v.point.z
+            );
+        }
+        println!(
+            "{path}: verts={} polys={} tris={} edges={} open_edges={} ({:.1}%)",
+            mesh.vertices.len(),
+            mesh.polygons.len(),
+            tris,
+            edges.len(),
+            open,
+            100.0 * open as f32 / edges.len().max(1) as f32,
+        );
+    }
+}
--- a/dark/src/importers/model_importer.rs
+++ b/dark/src/importers/model_importer.rs
@@ -119,8 +119,7 @@ fn process_vr_held_model(
     };
 
     let islands = ss2_bin_obj_loader::connected_islands(&obj, is_first_person_arm_material);
-    let drop: std::collections::HashSet<usize> =
-        islands.into_iter().skip(1).flatten().collect();
+    let drop: std::collections::HashSet<usize> = islands.into_iter().skip(1).flatten().collect();
 
     let obj = if drop.is_empty() {
         obj
--- a/dark/src/importers/model_importer.rs
+++ b/dark/src/importers/model_importer.rs
@@ -100,12 +100,14 @@ pub fn is_first_person_arm_material(name: &str) -> bool {
 /// A first-person weapon model prepared for VR wielding, as its own cache
 /// bucket (see [`FirstPersonHand`] for the type_id trap).
 ///
-/// The 25AE `obj/*_h` models bake in the player's hand and forearm, plus
-/// spare hand islands their reload/fire animations pose into view (e.g. the
-/// pistol's off-hand holding a magazine). In VR the model renders in bind
-/// pose from free viewpoints, so those spare hands float disembodied beside
-/// the gun. Keep the largest arm island (the gripping hand and its sleeve)
-/// and drop the rest; non-arm geometry is untouched.
+/// The mesh renders exactly as authored, baked hand and forearm included. An
+/// earlier revision stripped every arm island but the largest to hide the
+/// spare hands some reload/fire animations pose into view - but "one arm
+/// island = one hand" is false: the pistol (`atek_h`) splits its firing hand
+/// (`ND-arm_atek.psd`) and sleeve (`ND-arm.psd`) into separate islands, so the
+/// strip kept the sleeve and deleted the gripping hand. Spare-hand stripping
+/// is deliberately dropped (a floating spare hand is the lesser evil); it can
+/// return if a reliable classifier turns up.
 pub struct VrHeldModel(pub Model);
 
 fn process_vr_held_model(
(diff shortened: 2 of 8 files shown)


tommy-xr:
**Owner correction to this PR's conclusion** (recorded in the handoff doc): the investigation audited only the CLASSIC `_h` models and only checked the 25AE kpf for `_w` upgrades. The 25AE layers in fact ship **remastered `_h` first-person models that are complete** — `sshock2ee.kpf`: `wrench_h` (60,642 B vs classic 6,806), `psword_h`, `rapier_h`, `shard_h`, `pipewrench_h`, `al_h`, `amp_h`; mod layers add `atek_h`, `sg_h`, `fsn_h`.

Revised direction: **VR wield should use the 25AE `_h` models** when an upgraded one resolves (they're authored for close-up viewing), falling back to `_w` on classic installs. Next steps: run `mesh_audit` on the 25AE `_h` set to confirm closure; route VR wield to `_h`-with-fallback; re-examine grip offsets against the `_h` set (they may carry better origins — the X-axis heading question in this PR still applies).

tommy-xr:
I don't understand why we can't use the _h for the melee weapons?

tommy-xr:
also, we should not remove the primary hand - we should keep that in the mesh and make our 'empty' hand model disappear

tommy-xr:
Answering both questions above:

**Why not `_h` for melee (yet):** the gun `_h` models are standalone static LGMD meshes, so the existing held-model render path draws them as-is. The melee `_h` files (`wrench_h`, `rapier_h`, `shard_h`, `psword_h`, `pipewrench_h`) are **LGMM skinned meshes** — classic low-poly geometry with the remastered mesh in an appended `PMNM` chunk, deformed by a skeleton. Using them needs two things this PR doesn't have: (1) a posed-skeleton render path for held items (in bind pose a skinned arm+weapon renders collapsed/exploded — same reason ChangeModel needed the bind-pose player fix here, but worse since the remastered PMNM geometry only makes sense posed), and (2) on Quest, the PMNM parse gate from PR #1024. Nothing structural blocks it — it's the recorded follow-up, sequenced after #1024 lands.

**Primary hand:** it is kept — that's already how this PR behaves. The strip removes only the *spare* reload/fire off-hand islands (e.g. the pistol's floating magazine hand); the largest arm island — the gripping hand and its sleeve — stays in the mesh (`process_vr_held_model` in `dark/src/importers/model_importer.rs`). And the separate "empty hand" glove model already disappears whenever a weapon is wielded (`shows_hand_visual`, the #1019 display rule, unit-tested). So on wield you see exactly the model's own baked hand holding the gun, and nothing else.

tommy-xr:
**Correction to "Primary hand: it is kept" above** — that claim was wrong, and the contact sheet showed it: the **Pistol tile had the baked firing hand stripped**.

**Cause:** the spare-hand strip kept only the *largest* connected arm-material island, assuming one island = one hand. `atek_h` is the one gun where that fails — it has **three** arm islands across two arm materials: the sleeve (`ND-arm.psd`, 445 polys, kept) and the **firing hand** (`ND-arm_atek.psd`, 440 polys) plus more arm geometry (265 polys), both dropped. So the wielded pistol rendered a sleeve with a holed/handless grip, and the pale hand visible in the old tile was the co-located reference hand showing through. No other gun was affected: `ar15_h`/`sg_h`/`empgun_h` have exactly one arm island (nothing stripped), and `gren_h`/`fsn_h`/`sfg_h`/`al_h`/`viro_h`/`amp_h`/`lasehand` have no `ND-arm*` material at all (`viro_h`'s look is as authored — nothing was stripped there).

**Fix (per owner decision):** spare-hand stripping is dropped entirely — the VR-held model renders exactly as authored, baked hands included (`4b78241c`). A floating spare reload hand is the lesser evil vs deleting the real one; stripping can return later behind a reliable classifier. Pinned by an updated `vr-held-model` e2e (wielded `atek_h` must draw its full mesh — red before the fix) and re-verified in the refreshed contact sheet / before-after in the PR body.

Issue #916, opened 2026-08-12
Player-facing HUD message line (the original's overlay text), and what it should be in VR


tommy-xr:
## Problem

The original shows transient on-screen message lines — "Picked up log: \<name\>", pickup strings, cyber-module/XP gains, hacking success/failure, research reports, ammo swaps, and the script-driven objective/story text. This port has no such facility, so those messages currently go to the game log or are dropped, and the player never sees them.

This came up while making audio logs read-on-demand (#N): the original's *only* feedback for collecting a log is that overlay line. Because we cannot draw it, the log disc currently fires the message beep directly as a documented stand-in.

## What the original does

Two behaviors worth reproducing together, because the second is a consequence of the first:

1. **A single funnel for transient HUD lines.** Every one of those events posts through one "add overlay text" entry point rather than each drawing its own UI.
2. **That funnel plays a beep — `linebeep` — unconditionally, per message.** No rate limiting, no deduplication, no priority or type gating. If three lines post in the same instant, it beeps three times. This is the chirp players remember on pickup: it comes from *posting a message*, not from the pickup itself.

A consequence worth noting for fidelity: an ordinary auto-pickup item produces **two** sounds — the message beep, then the item pickup cue. An audio log produces exactly **one** (the beep), because `Audio Log` (template -76) overrides `FrobInfo` to `world_action: SCRIPT` and so never takes the frob path that plays the pickup cue. Verify with:

```
cargo dq templates 76   # Audio Log: world_action: SCRIPT
cargo dq templates 49   # Goodies:   world_action: MOVE
```

`linebeep` is a real schema in the shipped data (template -3827, `sym:linebeep`, parented to `UI_SCH` -3778), from the same UI schema family as the item pickup cue (-2734).

## Proposed work

- A message queue with per-message lifetime, and the beep fired centrally when a message is posted (not by each caller).
- A flat-mode renderer for the lines, in the original's screen position.
- Route the existing callers to it: the log pickup string, generic pickup strings, quest-bit/objective text, research and hacking results, XP gain.
- Once it exists, move the beep out of the audio-log script (see the `LOG_PICKUP_SOUND` comment in `shock2vr/src/scripts/gui/media.rs`), and gate the *item* pickup cue on the object actually carrying the `MOVE` world frob action so discs stay silent apart from the beep.

## Open question: what is this in VR?

[...]

PR #1026, opened 2026-08-18
feat(vr): show narration subtitles in both presentations, head-anchored toast in VR

tommy-xr:
VR narration subtitle toast — the four timed cues of the earth.mis arrival PA playing out on the world panel

<sub>VR presentation: the `trg0001` narration's four timed cues appearing/expiring on the head-anchored world panel (~2x speed). Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

## Flat / VR parity (one shared canvas)

flat vs VR

<sub>The same cue at the same canvas position in both presentations. The VR shot is deliberately taken against a wall well inside the 2 m panel distance — the toast rides the clear-depth overlay group, so near geometry cannot swallow it.</sub>

Implements the VR-UI plan's step 3b: the narration voice-overs (earth/station training PA, the in-mission Polito/SHODAN voice triggers) now show their transcript on screen while they play — in **both** presentations.

## What the research changed

The plan assumed flat already showed these texts and only VR was missing a presentation. It isn't so: the port showed **no** text in either presentation (`earthtext`/`trapmessage` are noop scripts; #916 tracks the missing HUD-line facility). The on-screen training text is a **25th Anniversary subtitle feature**: KEX cue files (`sshock2-subtitles.kpf`, `vbriefs.sub` + `vtriggers.sub`) whose `$key` lines resolve through `base.kpf:localization/loc_english.txt`. The classic 1999 data ships no transcripts at all — so on a classic install this change is a deliberate no-op and narrations stay audio-only, exactly as they were in 1999.

## How it works

- **`subtitles.rs`** — parsers for the KEX `.sub` cue grammar (timed `multisub` entries, untimed `sub` lines) and the `$key = "value"` localization table. 207 cue tracks load on a 25AE install; malformed input degrades to "no cues", never a crash.
- **`trap_sound.rs`** — `TrapSound` (the script all these narrations play through) emits `GlobalEffect::ShowSubtitle { audio_schema }` alongside its `PlaySound`. Ordinary sound traps simply miss the DB lookup; script purity is preserved (pure effect, unit-tested).
- **`subtitle_overlay.rs`** — a Game-level overlay (sibling of `pause_menu`): cue timing on the **scene clock** (freezes under pause), and **one shared 640×480 canvas** deciding placement once — wrapped lines (real glyph measurement), bottom-center, above the HUD vitals cluster (AGENTS.md §3: neither presentation makes any placement decision).
  - **Flat** renders that canvas in screen space (aspect-preserving letterbox, like every flat screen).
  - **VR** renders the *same canvas* on a world panel hung from the shared `FrontendPanelAnchor` (vr-ui-design rule 3): placed once from the head pose the narration starts at, yaw-only/gravity-aligned, **world-locked** while it plays, lazily recentered only after a sustained large deviation — never gaze-glued. At the 2 m panel distance the bottom-of-canvas text lands ~13° below view center, standard subtitle height. The toast rides the renderer's clear-depth overlay group (like the pause dim) so a wall nearer than 2 m can't swallow it.
- **`lib.rs`** — mounts the two extra archives (prefix-scoped, nothing else in `base.kpf` leaks into asset resolution), loads the DB once, resolves schema→sample the same way the sound system does, clears an in-flight narration on scene swaps.

Sound is untouched — the narration audio plays exactly as before, in both presentations.

## Verification

- Unit tests: parser grammar (timings, case-insensitivity, `$key` misses dropped, malformed input), overlay lifetime/dedupe/replacement/clear, TrapSound effect emission (negative-first: all assert behavior that doesn't exist on main).
- SDK e2e `subtitle-overlay.e2e.test.ts` (green): flat run asserts the `subtitle` scene source appears with the narration and expires with the cues; VR run asserts the same narration presents on a world panel within 4 m of the player. `missions.e2e.test.ts`: **all 23 missions load** (the mount change touches the shared load path).
- Both presentations render-verified on earth.mis `trg0001`, including the close-wall occlusion case in VR.
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo test -p shock2vr --lib` 825 green; no new clippy findings.
- Cross-engine review (`/xreview`, Claude + Codex + de-dup pass) — the High finding (VR occlusion) and both Mediums (scene-swap leak, one-frame un-placed anchor) are fixed in the review commit.
- Remaining Codex suggestions addressed in a follow-up commit: `TurnOff` now takes the subtitle down with its sound (`HideSubtitle` effect + unit tests), and the loader logs when cue files are present but `loc_english.txt` is missing instead of silently disabling subtitles. A third suggestion (registering the cue files with a `TEXT_IMPORTER`) was verified moot — no such importer exists; the loader reads raw. Re-validated after the fixes: fmt/check `-D warnings` clean, `cargo test -p shock2vr --lib` 827 green, subtitle e2e green in both presentations.

## Known follow-ups (out of scope here)

- station.mis `BriefTrap*` use the still-noop `TrapMessage` script with an *empty* sound prop — their trigger path needs its own investigation (they play nothing today either).
- The generic HUD message line ("Picked up log: …", `linebeep`) remains #916 — this PR deliberately implements only the narration-subtitle funnel.
- The crate now has four greedy word-wrap helpers (this one is the only glyph-measured one) and a fourth fixed-metrics test font stub — consolidation into `ui/`/`test_support` is a clean follow-up.
- `Localization` flattens `[section]` namespaces (safe for subtitle keys, documented); resolving the Nightdive `strings` stubs through it later needs section-qualified lookup first.

### DIFF
--- a/shock2vr/src/lib.rs
+++ b/shock2vr/src/lib.rs
@@ -1777,7 +1777,11 @@ impl Game {
                     .cues(&sample)
                     .or_else(|| self.subtitle_db.cues(&audio_schema));
                 if let Some(cues) = cues {
-                    self.subtitle_overlay.post(&sample, cues.to_vec());
+                    // Keyed by the *schema* name: `resolve_schema` is a random
+                    // draw for multi-sample schemas, so keying on the sample
+                    // would defeat the repeat-while-playing guard when a trap
+                    // re-fires. (Every loaded track today is a 1:1 schema.)
+                    self.subtitle_overlay.post(&audio_schema, cues.to_vec());
                 }
             }
         }
@@ -1788,6 +1792,10 @@ impl Game {
     /// scene swap goes through here so that hook cannot be forgotten.
     fn set_active_scene(&mut self, scene: Box<dyn GameScene>) {
         self.active_game_scene.on_exit(&mut self.audio_context);
+        // A narration's remaining cues belong to the outgoing scene - they
+        // must not keep drawing over the menu, the loading screen, or the
+        // next level.
+        self.subtitle_overlay.clear();
         self.active_game_scene = scene;
     }
 
@@ -1850,9 +1858,9 @@ impl Game {
         // aim at it), so it is mapped into world coordinates with the same
         // pawn transform the runtime builds its camera from.
         let pawn_to_world = Matrix4::from_translation(pos) * Matrix4::from(rot);
-        // The VR subtitle toast, before the pause panel: while the menu is up
-        // the frozen toast stays a world object behind the menu's dim rather
-        // than fighting the panel.
+        // The VR subtitle toast. Suppressed entirely while the pause menu is
+        // up (its clock is frozen anyway): both draw in the renderer's
+        // clear-depth overlay group, and the menu owns that space.
         if !self.pause_menu.is_open() {
             scene.extend(self.subtitle_overlay.render(
                 &mut self.asset_cache,
--- a/shock2vr/src/lib.rs
+++ b/shock2vr/src/lib.rs
@@ -33,6 +33,8 @@ pub mod quest_info;
 pub mod research;
 mod runtime_props;
 mod scripts;
+mod subtitle_overlay;
+mod subtitles;
 mod systems;
 #[cfg(test)]
 mod test_support;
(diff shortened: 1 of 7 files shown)


tommy-xr:
**Hold — this is not the requested feature, and it is currently on by default.** Owner clarification: the wanted feature is the **on-screen mission text that pops up in station.mis during character creation**, not narration subtitles. Two blocking notes before this merges:

1. **Subtitles must not be on by default.** There is no gating in this PR — `TrapSound` unconditionally emits `ShowSubtitle`, and the overlay ticks/draws unconditionally. The only incidental gate is that the DB is empty on classic installs. Smallest fix per repo idiom: guard behind `options.experimental_features.contains("subtitles")` (same pattern as `teleport`, `nav_bridges`, `ragdoll`). The #1027 dev-params registry is a poor fit — it is float-only, for live-tuned numbers, not feature switches.

2. **This is not a prerequisite for the station text.** Roughly two-thirds of the diff (`subtitles.rs` cue-track/`loc_english.txt` parser, the `sshock2-subtitles.kpf`/`base.kpf` mounts, the audio-schema-keyed effects, the `TrapSound` coupling) is narration-specific. The reusable third — canvas → flat screen-space / VR head-anchored panel, overlay depth group, scene-swap clear — already exists on `main` in `ui/panel_anchor.rs` + `UiCanvas`, as `pause_menu.rs` demonstrates. Critically, this pipeline is a **no-op on classic (1999) installs**, while the chargen text comes from `CHARGEN.STR`, which ships with both.

The station chargen text is a from-scratch feature: load `CHARGEN.STR` via `STRINGS_IMPORTER`, surface `TourReward::text_key` (`player_stats.rs:358`, currently dead) and/or implement `EarthText`/`TrapMessage` for real (`scripts/mod.rs:917,919` are `NoopScript` today), add a generic `GlobalEffect::ShowMessage { text, duration }`, and render through a small shared overlay that works on classic data.

Suggested path for this PR: gate subtitles off by default, then either merge as opt-in or split out a presentation-neutral `message_overlay` (text + duration, no sample keys, no cue tracks) that both the station-text feature and the subtitle DB plug into.

tommy-xr:
will hold off on this for now, needs better UI treatment

Issue #1029, opened 2026-08-19
TrapMessage: the message source is P$UseMsg, a key into USEMSG.STR


tommy-xr:
`trapmessage` is still `NoopScript` (`shock2vr/src/scripts/mod.rs:919`) with the comment *"What prop for message? Where to load string?"*. Investigating the character-creation mission text (#PR feat/mission-text) answered both halves:

**The property is `P$UseMsg`.** It is currently unparsed (it shows under *Unparsed Properties* in `cargo dq`) and is not in `dark/src/properties/`. It is a per-entity variable-length string, laid out like the other string properties (entity id, chunk length, string length, NUL-terminated bytes). Raw decode of the `P$UseMsg` chunk in `eng2.mis`:

| entity | value |
| --- | --- |
| 130 | `Comp_Online` |
| 393 | `Broken: "This lift is malfunctioning. Maintenance has been notified"` |
| 450 | `Botcoming` |
| 689 | `OutOfOrder` |

**The string table is `res/strings/USEMSG.STR`**, which contains exactly those keys, e.g.

```
OutOfOrder:"This lift has been taken offline for repairs."
```

(Entity 393's value is an author pasting the whole STR line in rather than the key alone - worth tolerating: if the value contains `:"` treat it as literal text, otherwise as a key.)

**Why it is now cheap to implement.** `GlobalEffect::ShowMessage { text, duration }` plus the shared `MessageOverlay` (flat screen-space / VR head-anchored panel) land with the character-creation debrief work, so `TrapMessage` needs only: a `PropUseMsg` parser, and a script that on `TurnOn` resolves the key against `USEMSG.STR` and returns `ShowMessage`. In eng2 the three elevator buttons switch-link into entity 689, so "press the broken lift button, read *This lift has been taken offline for repairs.*" is the e2e.

Separately: `earthtext` (`scripts/mod.rs:917`) has **no** text property at all - `earth.mis` entity 449 carries only `PropObjectSound`/`PropScripts` - so its string was hardcoded in the original script and is not recoverable from the data the same way.

PR #1030, opened 2026-08-19
feat(gameplay): show the character-creation debrief as its own screen

tommy-xr:
System Shock 2 shows a page of text when a character-creation training tour
ends — "Your stint aboard the UNN Gallo is finished… You've gained +2
Strength." Nothing implemented it here, in either presentation. This adds it,
as its own screen.

## The data discovery

Mission text had no known source in the data, so this started with `cargo dq`
on `station.mis`.

**The chargen markers carry no text.** The three `S$ChooseMission` entities in
`station.mis` (template −327 "Marker") hold exactly six properties:
`DestLevel`, `DestLoc`, `Position`, `Scale`, `Scripts`, and **`P$CharGenRo`** —
the tour index, 0/1/2. No text property, no link to one. So the string is not
authored on the trigger: it is *selected in code* from (career, training year,
tour), which is precisely the 27-entry table `player_stats::REWARDS` already
encodes, each row carrying the `Mission1..Mission27` key of
`res/strings/CHARGEN.STR` in `TourReward::text_key` — a field documented as
"display only … so a future UI can show the message" and, until now, read by
nobody. `CHARGEN.STR` ships in `sshock2.kpf`/`res/strings` on **both** classic
and anniversary installs, so the feature is not 25AE-only.

## A screen, not an overlay

Per review direction, the debrief is an **interstitial**, not a HUD layer — the
original gives this moment a whole page, so it gets one:

- **`DEBRIEF.PCX`** backdrop, **`DEBRIEFR.BIN`** for the Continue button's rect
  (one LTRB rect, `425,401 210x74`, via `dark::importers::UI_LAYOUT_IMPORTER`),
  **`DEBRIEF.STR`** for its label. The body is the tour's `CHARGEN.STR` page,
  word-wrapped on real glyph widths into the backdrop's text panel, drawn at the
  font's authored size (`text_native`) like every other screen's body text.
- **`DebriefScene`** is a sibling of `GameOverScene`: one shared `UiCanvas`
  built once in 640×480 canvas pixels, presented screen-space on flat and on the
  shared frontend world panel in VR, with the same `FrontendMenu` driving the
  pointer, hover and click in both. Neither presentation makes a placement
  decision of its own (AGENTS.md §3).
- **`GlobalEffect::ShowDebrief { text_key, then }`** mirrors `PlayCutscene`: the
  tour trigger hands it the transition it wanted, the page stands in for the
  mission scene, and **Continue** dispatches the follow-on — the departure
  cutscene, then the level change. Both effects now go through one
  `Game::show_over_mission`, which saves the replaced scene (so the follow-on
  carries career, training year, inventory and vitals) and, if the stand-in
  cannot be built, runs the follow-on rather than stranding the player. A
  missing or blank `CHARGEN.STR` page therefore just departs, as before.
- The page is selected from the grant the trigger is *about to make*, and is
  suppressed for a training year already granted — so it can never claim a stat
  change `apply_tour_reward` would refuse.

**`GlobalEffect::ShowMessage` and the message overlay are gone.** They were
introduced by the earlier revision of this PR and nothing else used them; a
screen needs none of the stand-down rules an overlay did (pause menu, loading
screen). Issue #1029's message trap (`P$UseMsg` → `USEMSG.STR`) will want a
transient HUD line — that is a different presentation from this page, and is
better added with its consumer than kept alive here unused.

## What it looks like

debrief screen demo

<sub>The tour ends, the page comes up and waits, Continue lights on hover and departs into the shuttle clip. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

flat vs vr debrief screen

<sub>The same tour (Marine, year 1, tour 0 -> `Mission1`) in both presentations: same wrap, same rows, Continue in the same place - the VR shot differs only by its pointer ray, because both draw the one canvas.</sub>

There is no before/after pairing: nothing rendered this text on `main`, in
either presentation.

## Verification

- `cargo fmt --all --check`; `RUSTFLAGS="-D warnings" cargo check -p shock2vr
  -p desktop_runtime -p debug_runtime` clean.
- `cargo test -p shock2vr --lib`: 1176 pass, including the new scene tests
  (Continue's rect and its rising-edge click emitting exactly the deferred
  effect, the wrap and its paragraph rows) and the script test pinning the page
  in front of the departure clips.
- New `tools/shock2-sdk/test/debrief.e2e.test.ts` — negative-first: nothing on
  `main` lands on a `debrief` scene. It completes a tour in `station.mis`,
  asserts the page is that tour's own `CHARGEN.STR` entry (`/v1/ui`
  `debrief_text`), that it *waits* rather than expiring, and that Continue
  departs — in **flat** and in **VR**, through the same canvas rect. The VR
  case also asserts the page is on the panel in front of the head.
- `station-flow.e2e.test.ts` now dismisses the page the way a player does
  (shared `dismissDebrief` helper); all 23 missions still load
  (`missions.e2e.test.ts`).
- Driven end to end on the debug runtime, flat and `--vr`: tour → page →
  Continue → `shuttle1.avi` → `station.mis` with Strength 3 and
  `training_year_2` set, i.e. the preserved-state contract holds.

## Reviewed

`/xreview` — both engines (Claude + Codex) plus a dedicated duplication pass.

**[AGREED] by both engines, fixed:** `show_over_mission` (and, before it,
`PlayCutscene`) would clear a pending transition and then save the *loading*
scene as the outgoing state — wiping quest bits, held items and vitals. It now
refuses exactly as `begin_transition` does, and the transition the player is
waiting for is what lands.

Other findings applied: the `ShowDebrief` arm was a near-verbatim copy of
`PlayCutscene`'s (now one `show_over_mission`); the screen hand-rolled a button
the shared `FrontendMenu` helpers already resolve; the body scaled a bitmap font
1.45× against the repo's fidelity default (now `text_native`); an over-tall page
silently dropped the row naming the grant (now truncated with a one-shot
warning) and an over-wide word could run off the panel (rows are fitted);
`debrief_key` recomputed a grant the caller already had, and now also returns
nothing for an already-granted year. Test-side duplicates — the flat and VR
canvas-click sequences — moved into `helpers/frontend-menu.ts`.

Declined, with reasons: the measured word wrap is the fifth greedy wrapper in
the tree (the other four count characters); converging them belongs in the
shared font layer, not here. Codex wanted the page enqueued only on
`apply_tour_reward`'s own `true`, which would mean emitting it from the effect
applier and so losing the ability to wrap the transition; the pre-effect guard
covers every reachable case (two tour markers cannot fire in one frame in-game),
and the applier stays authoritative for the stats themselves. `CHARGEN.STR` also
carries `MissionPic1..N` → `D001..D006` for the backdrop's empty circular
picture panel; that is a follow-up, not part of "a page of text".

### DIFF
--- a/runtimes/debug_runtime/src/commands.rs
+++ b/runtimes/debug_runtime/src/commands.rs
@@ -610,6 +610,9 @@ pub struct UiStateResult {
     /// while the interface is up in VR. Aim a controller at a canvas rect by
     /// mapping it through this.
     pub panel_pose: Option<shock2vr::game_scene::DebugUiPanelPose>,
+    /// The character-creation debrief page currently on screen, verbatim;
+    /// `None` when the debrief screen is not up.
+    pub debrief_text: Option<String>,
 }
 
 /// A single carried item.
--- a/runtimes/debug_runtime/src/main.rs
+++ b/runtimes/debug_runtime/src/main.rs
@@ -1319,6 +1319,9 @@ fn process_command(
             }
         }
         RuntimeCommand::GetUiState { reply } => {
+            // Game-level, so it is reported whether or not the active scene
+            // exposes debug UI state.
+            let debrief_text = game.active_debrief_text().map(str::to_owned);
             let result = game
                 .debug_scene()
                 .map(|scene| {
@@ -1331,6 +1334,7 @@ fn process_command(
                         ammo_cycle: ui.ammo_cycle,
                         pointer: ui.pointer,
                         panel_pose: ui.panel_pose,
+                        debrief_text: debrief_text.clone(),
                     }
                 })
                 .unwrap_or(commands::UiStateResult {
@@ -1341,6 +1345,7 @@ fn process_command(
                     ammo_cycle: None,
                     pointer: None,
                     panel_pose: None,
+                    debrief_text,
                 });
             if reply.send(result).is_err() {
                 tracing::warn!("Failed to send ui state - receiver dropped");
--- a/shock2vr/src/game_scene.rs
+++ b/shock2vr/src/game_scene.rs
@@ -140,6 +140,13 @@ pub trait GameScene {
         false
     }
 
+    /// The character-creation debrief page this scene is showing, if it is the
+    /// debrief screen. Debug introspection only (`/v1/ui`), so a test can
+    /// assert which page came up.
+    fn debrief_text(&self) -> Option<&str> {
+        None
+    }
+
     /// Whether the in-game pause menu may open over this scene.
     ///
     /// True for scenes that run a simulation the player would want frozen -
--- a/shock2vr/src/lib.rs
+++ b/shock2vr/src/lib.rs
@@ -1950,8 +1950,16 @@ impl Game {
     /// stand-in chaining into another keeps what the first captured - the empty
     /// world it would capture now is exactly what must not win.
     fn show_over_mission(&mut self, scene: Option<Box<dyn GameScene>>, follow_on: GlobalEffect) {
-        // Any pending transition is abandoned, as for the frontend swaps.
-        self.pending_transition = None;
+        // The same refusal `begin_transition` makes, for the same reason: with a
+        // transition in flight the active scene is the loading screen, and
+        // saving *that* as the outgoing state wipes quest bits, held items and
+        // vitals. Nothing in-game can reach this (the loading scene runs no
+        // scripts), but an external caller can, so refuse rather than corrupt -
+        // the transition already under way is what the player is waiting for.
+        if self.pending_transition.is_some() {
+            warn!("Ignoring a scene standing in for the mission: a level transition is in flight");
+            return;
+        }
         let preserved = match self.preserved_scene_state.take() {
             Some(carried) => carried,
             None => self.save_active_scene(),
--- a/shock2vr/src/lib.rs
+++ b/shock2vr/src/lib.rs
@@ -57,7 +57,7 @@ mod wielded_weapon;
 pub mod zip_asset_path;
 
 use scenes::{
-    CutscenePlayerScene, DeveloperScene, GameOverScene, LoadGameScene, MainMenuScene,
+    CutscenePlayerScene, DebriefScene, DeveloperScene, GameOverScene, LoadGameScene, MainMenuScene,
     SceneInitResult, create_initial_scene, load_mission_from_save_data, resolve_ending_cutscene,
 };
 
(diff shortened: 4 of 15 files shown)


tommy-xr:
Rebased onto main (through #1043).

**Conflicts and how they were resolved** — all in `shock2vr/src/lib.rs`, where main grew three new `Game` fields and a new `GlobalEffect` arm alongside this branch's:

- **`Game::new` / struct fields** — took both: main's `hit_feedback` + `head_pose` (#1038) beside this branch's `message_overlay`.
- **`GlobalEffect` match** — took both arms: main's `PlayerHit` and this branch's `ShowMessage`.
- **`ShowDeveloper` (#1028)** — main open-coded the `pending_transition = None` + `set_active_scene` pair that this branch factors into `replace_with_frontend_scene`. Converged it onto the helper (commit `ad85acc4`), so opening the Developer *screen* also clears mission text in flight, like every other frontend swap.
- **`util.rs` render sources** — took both `HIT_FEEDBACK` and `MISSION_TEXT`.

`git range-diff prerebase-1030...` confirms nothing was dropped: the test commit is byte-identical, the feature commit differs only in rebased context, and the stand-down commit is 5 lines smaller solely because a `set_active_scene` reordering hunk is now redundant.

**Interaction check** (`/v1/scene` source ordering, station.mis, both presentations):

| state | `mission_text` rows | sources present |
| --- | --- | --- |
| debrief showing | 7 | `entity`, `mission_text` |
| + hit vignette (#1038) | 7 | `hit_feedback` @137, `mission_text` @144 |
| pause overlay, root page | **0** | `hit_feedback`, `pause_dim` |
| pause overlay, **Developer page** (#1028) | **0** | `hit_feedback`, `pause_dim`, `frontend_pointer` |

- The pause stand-down survived the rebase and **already covers the Developer page**, because that page lives inside the overlay (`PauseMenuPage::Developer`) and `pause_menu.is_open()` stays true — no new gate needed.
- The page composites correctly with the hit vignette: both are in the clear-depth group and the text is submitted *after* the rim, so the debrief stays legible over a hit.
- No duplicate head-tracking guard to converge: the panel reuses the shared `FrontendPanelAnchor`, whose zero-quaternion guard is main's, and this branch adds no eye-pose maths of its own.

**Validation on the rebased branch**: `cargo fmt --all --check`; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`; `cargo test -p shock2vr --lib` (885 pass); e2e `mission-text` (flat + VR), `player-hit-feedback`, `pause-menu`, `dev-menu`, `dev-params` — 16/16 pass; `missions.e2e` — 23/23 missions load.

tommy-xr:
this text should be shown as a separate pane (we should make a mission/scene for it as an interstial) - using debrief.pcx / debrief.str / debriefr.bin

Issue #956, opened 2026-08-13
fix(vr): melee contact damage sends no DamageImpact, losing the death-ragdoll reaction


tommy-xr:
Found reviewing #943.

The flat melee swing sends a full impact so the victim's death ragdoll reacts directionally (`shock2vr/src/scripts/weapon_script.rs`):

```rust
impact: Some(DamageImpact { direction: aim.forward.normalize(), point: hit_point.to_vec(), bone: None }),
```

The VR contact path (`melee_weapon.rs`, `melee_impact`) sends `impact: None`, so VR kills crumple with no directionality — even though the physical contact is exactly where that data would come from.

The blocker is that `MessagePayload::Collided { with }` (`shock2vr/src/scripts/mod.rs:226`) carries only the other entity — no contact point or normal.

## Proposed fix

Either carry the contact point/normal through `Collided` from the physics contact manifold, or synthesize `direction` from the held weapon's kinematic delta between frames.

Issue #955, opened 2026-08-13
fix(vr): bound the melee attack window to a swing instead of the whole trigger hold


tommy-xr:
Found reviewing #943 (and raised in review on the PR itself).

`TriggeredMeleeWeapon` (`shock2vr/src/scripts/melee_weapon.rs`) sets `attack_active = true` on `TriggerPull` and only clears it on `TriggerRelease`/`Drop`. `hit_entities` is cleared on the same edges.

Consequences:

1. **The window never closes while the trigger is held.** A player holding the trigger sweeps a permanently-armed weapon and damages every distinct entity it touches, including scenery and doors. Dark's melee window is a bounded swing, not "as long as the button is down".
2. **A target can be hit only once per pull**, so you cannot hit the same enemy twice without a full release/re-pull.
3. **No rate limit between pulls** — trigger-spam has no cooldown.

## Proposed fix

Bound the window with a short timer from `TriggerPull` (advanced in `Update`), clearing `hit_entities` per swing rather than per hold, and add a minimum interval between armed pulls. State stays transient — no save/load serialization needed.

Ideally shares its timing source with the flat animation-event work in #951 so both presentations agree on when a swing lands.

PR #1032, opened 2026-08-19
feat(vr): wield the 25AE melee _h models, with melee damage proven through the production path

### LINKED ISSUE #1031: fix(vr): melee _h wield must not move the weapon's contact collider
Follow-up found while wielding the 25AE melee `_h` models in VR (plan step 3d).

## The trap

VR melee damage is a physical contact test against the **held entity's rigid body** (#942, #978, #984). That body is placed by `VirtualHand`'s `SetPositionRotation`, which applies the model's `HAND_MODEL_POSITIONING` grip:

```
position = hand_position + hand_rotation * adjust.offset
rotation = hand_rotation * adjust.rotation
```

So a grip offset is **not** a render tweak — it moves the damage volume too.

The remastered melee `_h` models are skinned meshes with a whole baked arm: seating their fist in the palm needs a ~0.7-unit model-space shift. Applied as a grip offset, that carries the wrench's authored 4.6 cm contact sphere ~0.7 units off the weapon, and the wield looks perfect while dealing **no damage at all**.

Measured on `feat/vr-25ae-melee` before the fix: `medsci-saved-vr-melee.e2e.test.ts` — a shipped MedSci1 Wrench, one trigger-gated pull at a Blue Monkey — emitted **0 Damage messages** where it must emit exactly one for 9 HP. The same test is green at the branch base.

## The invariant

Alignment of a wielded first-person model is a **render-only** correction. It belongs in the model's local transform (`Model::apply_local_transform`), never in the hand grip, which the contact collider rides.

Guns are unaffected in practice (they do no contact damage), but the same rule applies if that ever changes.

tommy-xr:
Wields the four melee weapons with the 25AE remastered first-person `_h` models — the model's own baked hand replaces our glove — and, per the owner's "a pretty wrench that does nothing is NOT done", proves melee damage still works through the production path.

Plan step **3d** (`vr-ui-plan.md`). Fixes #1031.

## The wield

| weapon (gamesys) | model | install |
| --- | --- | --- |
| Wrench (−928) | `wrench_h` | 25AE; classic keeps `wrench_w` |
| Electro Shock (−24) | `rapier_h` | 25AE; classic keeps its world model |
| Crystal Shard (−28) | `shard_h` | 25AE; classic keeps its world model |
| PsiSword (−2291) | `psword_h` | 25AE; classic renders nothing (it authors no world model) |

`pipewrench_h` ships in the 25AE layer but has **no gamesys template** (KEX drives it from a squirrel script we don't run), so it is not wieldable and is not listed.

Melee `_h` are LGMM **skinned** meshes, unlike the standalone-LGMD guns of #1023: the bind pose renders as a splayed mid-swing arm, so the wield holds the player-melee idle clip's final frame (`from_completed_animation` — a static pose that emits no motion flags, events, or root velocity).

melee wield contact sheet

cycling the four melee models in VR

Captured per-weapon in `debug_weapons --vr` (`DebugCycleWeapon` → raycast-grab → wield). Full-resolution stills for each weapon are in [the same gist](https://gist.github.com/tommy-xr/0262feac71550040b3c1cc41d6b82702).

## The damage bug (#1031)

The first cut of this branch seated each arm with a **grip offset**. That looked right and silently disarmed the weapon: `VirtualHand::SetPositionRotation` applies the grip to the **entity**, and the held entity's rigid body *is* the VR melee contact collider (#942/#978). A ~0.7-unit alignment offset carried the Wrench's authored 4.6 cm contact sphere clean off the weapon.

Measured, not guessed:

| build | `medsci-saved-vr-melee` — one trigger-gated pull at a Blue Monkey |
| --- | --- |
| base (`d63707af`) | 1 Damage message, 9 HP lost ✅ |
| this branch, grip-offset cut | **0 Damage messages** ❌ |
| this branch, shipped | 1 Damage message, 9 HP lost ✅ |

The fix: each melee `_h` keeps **its world model's physical grip**, and the arm is seated by a render-only model-space correction through `Model::apply_local_transform` — the existing seam, already documented as "composed inside the entity transform, which the render path re-sets every frame". `vr_config::melee_wield_pose_correction` cancels the grip, applies the authored yaw, and cancels the posed fist joint, so the rendered result is grip-independent while the collider never moves.

Damage proof through the production path (shipped command2 1-HP pane, shipped Wrench, real VR trigger + physical swing — no injected damage message):

melee damage before/after

## Also fixed (found by review)

The PsiSword authors no world model, so the wield **materializes** its first one — and the drop had no inverse, leaving the first-person arm mesh lying in the world forever and persisting through save/load (reproduced live). `Effect::ClearModel` makes the swap symmetric; a new e2e covers grab and release.

## Owner review: hand orientation, and where the hitbox is

Both findings were real; both are fixed here (commits `a52ea56a`, `781a41f8`).

### 1. The baked arm was not on the hand's arm axis

Measured against the frame the tracked hand actually defines - `hand_forearm`
hangs the empty hand's sleeve along the hand's **+Z**, and the fingers/aim run
down **-Z** - the shipped seating put the posed forearm here:

| model | forearm (fist -> elbow) in hand space | off the hand's arm axis | correcting axis |
| --- | --- | --- | --- |
| `wrench_h` | (-0.348, 0.102, 0.122) | **71.3 deg** | (0.28, 0.96, 0.00) - a yaw |
| `rapier_h` | (-0.366, 0.073, 0.067) | **79.8 deg** | (0.20, 0.98, 0.00) |
| `shard_h` | (-0.366, 0.073, 0.067) | **79.8 deg** | (0.20, 0.98, 0.00) |
| `psword_h` | (-0.356, 0.070, 0.069) | **79.3 deg** | (0.19, 0.98, 0.00) |

So the arm left the wrist *sideways* instead of running back toward the elbow.
All four are nearly identical because the four rigs share the same posed arm
(only the weapon joint differs).

**This is not a different authoring axis from the guns.** The 25AE guns of
\#1023 are standalone LGMD meshes authored barrel-along -X, and one -90 yaw maps
-X onto the hand's -Z aim axis - for a gun that single degree of freedom is the
whole job. A melee `_h` is a skinned *arm*, so seating it is a 3-DOF problem
(arm onto the arm axis, plus a roll that decides where the weapon points), and
one authored yaw can only satisfy one of the two. The shipped 120 deg satisfied
the weapon and paid for it with the arm.

The correction is now derived from the rig instead of authored:
`melee_wield_alignment` rotates the posed forearm onto +Z, then rolls about it
until the weapon lies in the hand's vertical plane, standing up out of the fist.
Forearm error after: **< 0.5 deg** for all four (unit-tested).

melee arm before/after

### 2. The hitbox was not on the weapon - it is now

Answering the question directly: `Model::apply_local_transform` sets
`local_transform` on **every scene object of the model**, so the render
correction displaced the *whole* wield - arm and weapon geometry alike - while
the collider stayed at the entity transform. The suspicion was right.

Measured in hand-local space (collider = the held entity's body; the Wrench's
authored contact volume is a 4.6 cm sphere):

| model | contact sphere | rendered weapon head | gap |
| --- | --- | --- | --- |
| `wrench_h` | (0, 0, -0.40) | (0.013, 0.591, -0.556) | **0.61** (and 0.29 from the *nearest* point of the shaft) |
| `rapier_h` | (0, 0, 0) | (-0.032, 1.096, -0.416) | **1.17** |
| `shard_h` | (0, 0, 0) | (-0.154, 1.184, -0.339) | **1.24** |
| `psword_h` | (0, 0, 0) | (-0.159, 1.142, -0.290) | **1.19** |

The damage volume was floating in clear air next to the weapon, and the damage
e2e passed anyway because it drives the collider. (Note `/v1/scene` cannot see
this: it reports `SceneObject::transform`, not `transform * local_transform`,
so it printed the entity position for both.)

The fix follows from the mechanism the previous commit established. Because the
correction *cancels* the grip, the grip entry moves no pixels for a melee `_h` -
it only places the body. So the wield now **computes** that placement where it
poses the arm (`melee_contact_offset`) and stores it per entity as
`RuntimePropVrGripOffset`, which the grip lookup prefers; the correction cancels
that same value. Both halves come from one posed arm, so the weapon joint
renders exactly on the collider **for any rig** - a property, not four
hand-measured constants somebody has to keep up to date. (The four melee entries
are gone from `HAND_MODEL_POSITIONING` entirely, which also restores the
180-degree projectile rotation they had before VR wielded them.) Verified live,
one frame, hand at a known pose:

| model | measured body offset from the hand |
| --- | --- |
| `wrench_h` | (0, 0.811, -0.032) |
| `rapier_h` | (0, 1.160, 0.168) |
| `shard_h` | (0, 1.201, 0.316) |
| `psword_h` | (0, 1.146, 0.315) |

the collider framed - the camera aims straight at the contact sphere

Each panel points the camera *at the contact sphere*: the weapon head is what
is in the crosshair.

**Known limitation, filed as #1039:** the authored contact volume is a ~5 cm
sphere, so this arms the weapon's *head* and nothing else along its length.
Covering the whole blade needs a collider shaped like the weapon (a capsule
fist->head), which is a bigger change than putting the existing one in the right
place.

**One consequence, fixed here:** the held entity now sits on the weapon head, so
a release used to respawn the loose prop out there - a dropped Crystal Shard
materialized 1.2 units above the palm and fell. `DropItem` puts it back on the
hand first (measured: release at hand+1.22 -> hand+0.41, where every other drop
already lands).

Two asset-free tests hold both halves: `melee_wield_seats_the_forearm_on_the_hands_arm_axis`
(forearm error < 0.5 deg) and `the_rendered_weapon_head_lands_on_the_contact_collider`,
which composes the entity transform with the render correction and asserts the
weapon joint lands on the collider and the fist on the hand, for every shipped
rig. `cargo run -p shock2vr --example melee_grip` prints the offsets from the
live assets.

One test change follows: `medsci-saved-vr-melee`'s gesture placed the *hand* at
the sweep distance, which only connected while the collider happened to sit
0.4 along the fingers. It now measures the held weapon's grip offset live and
backs the hand off by it, so the sweep means what its comment always said -
bring the weapon's contact volume to 0.33 units from the torso. Same gesture,
same assertions, no injected damage. The measurement asserts its own shape
(~0.8 out, standing up out of the fist) so the staging cannot quietly
compensate for a misplaced collider, and it takes its hand rotation from the
SDK's level-horizon `lookQuat` - shortest-arc from -Z rolls up to ~180 degrees
for directions near world +z, which is exactly where these targets sit, and
that roll now carries an 0.8-unit offset.

### Review of this round

`/xreview` again (Claude subagent + Codex + the de-dup pass). Both engines
independently landed on the same structural point - the render half was
computed from the live rig while the collider half was pasted constants - which
is what the `RuntimePropVrGripOffset` derivation above replaced. Also applied
from review: the joint indexing is guarded (a short/modded rig logs and wields
unseated instead of panicking the frame), the drop-position regression above,
the tautological-staging and level-horizon fixes in the e2e, and #1039 filed for
the tip-only contact volume. `vr_config` stays a private module (two items are
re-exported for the example).

## Flat is unchanged

`FLAT_WIELD_SWAP_MODELS` is untouched and still frozen by its unit test; the `_h` route is gated on VR **and** `is_25th_anniversary_install()`. The flat wield e2e (`vr-held-model.e2e.test.ts`) is green.

## Review

`/xreview` run on the full branch (Claude subagent + Codex CLI + a dedicated de-dup pass), with the contact sheet attached for visual verification. Both engines independently agreed on two findings, both applied here: use `Model::apply_local_transform` instead of a new `AnimationPlayer` field (which is what the earlier draft did), and the PsiSword drop asymmetry. Both engines independently verified the correction's matrix algebra against `virtual_hand`'s forward composition. A remaining Medium — the correction bakes the *right* hand's grip at wield time — is now guarded by a unit test asserting the four melee grips are hand-symmetric.

One test change is bundled and deserves a callout: `vr-melee-contact`'s sweep now steps **2** frames per sample instead of 1. The Wrench's contact volume is a 4.6 cm sphere and the sweep's last sample lands it *on* the pane plane, so at one frame per sample the same swing passed or failed purely on how much physics free-ran between HTTP requests — measured: adding one read per sample flipped the result. Two frames gives the contact a real overlap window; the gesture is otherwise unchanged, and the test still asserts the same damage.

## Validation

- `cargo fmt --all --check`; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`
- `cargo test -p shock2vr --lib` — 817 pass
- SDK e2e (`SHOCK2_E2E=1`): `medsci-saved-vr-melee` (still exactly **1 Damage / 9 HP** per pull, plus the three-pull OG-Shotgun kill and the world pane), `vr-melee-contact` (3), `vr-held-model` (5, incl. the new PsiSword drop and the flat swap), and the whole `vr-*` + `weapon-*` set — **49 tests, all green** (including 23/23 missions)
- `missions.e2e.test.ts` — 23/23 missions load

## Follow-ups (not in scope)

- The empty hand is the pale 25AE bare-skin hand while the `_h` meshes bake a dark-skinned hand in a black sleeve — the two hands read as different characters. Newly visible because the wield now shows the *model's* hand.
- #955 (attack window bounded to a swing) and #956 (missing `DamageImpact`) remain open, as does the two-handed grip thread.

### DIFF
--- a/dark/src/motion/animation_player.rs
+++ b/dark/src/motion/animation_player.rs
@@ -73,6 +73,12 @@ pub struct AnimationPlayer {
     /// `AnimationInfo::cancel_root_motion`). Set for first-person viewmodels,
     /// whose entity transform is re-anchored to the camera every frame.
     cancel_root_motion: bool,
+    /// Rigid model-space transform composed onto every posed joint after
+    /// animation (`get_transforms` returns `post * joint`). Used by the VR
+    /// melee wield to align the posed arm's fist and weapon axis with the
+    /// entity's grip frame without moving the entity (and its contact
+    /// collider). `None` for every ordinary player.
+    post_transform: Option<Matrix4<f32>>,
 }
 
 impl AnimationPlayer {
@@ -87,6 +93,7 @@ impl AnimationPlayer {
             blend_state: None,
             rotation_pos: 0.0,
             cancel_root_motion: false,
+            post_transform: None,
         }
     }
     pub fn from_animation(animation_clip: Rc<AnimationClip>) -> AnimationPlayer {
@@ -101,6 +108,7 @@ impl AnimationPlayer {
             blend_state: None,
             rotation_pos: 0.0,
             cancel_root_motion: false,
+            post_transform: None,
         }
     }
 
@@ -120,6 +128,7 @@ impl AnimationPlayer {
             blend_state: None,
             rotation_pos: 0.0,
             cancel_root_motion: false,
+            post_transform: None,
         }
     }
 
@@ -165,6 +174,7 @@ impl AnimationPlayer {
             blend_state,
             rotation_pos: 0.0,
             cancel_root_motion: player.cancel_root_motion,
+            post_transform: player.post_transform,
         }
     }
 
@@ -237,6 +247,7 @@ impl AnimationPlayer {
             blend_state,
             rotation_pos: 0.0,
             cancel_root_motion: player.cancel_root_motion,
+            post_transform: player.post_transform,
         }
     }
 
@@ -245,6 +256,17 @@ impl AnimationPlayer {
     pub fn with_root_motion_cancelled(player: &AnimationPlayer) -> AnimationPlayer {
         let mut new_player = player.clone();
         new_player.cancel_root_motion = true;
+        new_player
+    }
+
+    /// A copy of `player` whose posed joints are all pre-multiplied by
+    /// `transform` (see the `post_transform` field).
+    pub fn with_post_transform(
+        player: &AnimationPlayer,
+        transform: Matrix4<f32>,
+    ) -> AnimationPlayer {
+        let mut new_player = player.clone();
+        new_player.post_transform = Some(transform);
         new_player
     }
 
@@ -265,6 +287,7 @@ impl AnimationPlayer {
             blend_state: player.blend_state.clone(),
             rotation_pos: player.rotation_pos,
             cancel_root_motion: player.cancel_root_motion,
+            post_transform: player.post_transform,
         }
     }
 
@@ -401,6 +424,7 @@ impl AnimationPlayer {
                                 blend_state,
                                 rotation_pos: raw_end_pos - current_clip.num_frames as f32,
                                 cancel_root_motion: player.cancel_root_motion,
+                                post_transform: player.post_transform,
                             },
                             motion_flags,
                             events,
@@ -438,6 +462,7 @@ impl AnimationPlayer {
                                 // slice is emitted on its first tick.
                                 rotation_pos: 0.0,
                                 cancel_root_motion: player.cancel_root_motion,
+                                post_transform: player.post_transform,
                             },
                             motion_flags,
                             events,
@@ -468,6 +493,7 @@ impl AnimationPlayer {
                         blend_state,
                         rotation_pos: raw_end_pos,
                         cancel_root_motion: player.cancel_root_motion,
+                        post_transform: player.post_transform,
                     },
                     motion_flags,
                     events,
@@ -523,7 +549,7 @@ impl AnimationPlayer {
         if maybe_current_clip.is_none() {
             let animated_skeleton =
                 ss2_skeleton::animate(skeleton, None, &self.additional_joint_transforms);
-            return animated_skeleton.get_transforms();
+            return self.apply_post_transform(animated_skeleton.get_transforms());
         }
 
         let (rc_animation_clip, is_looping, is_last_anim) = maybe_current_clip.unwrap();
@@ -580,7 +606,16 @@ impl AnimationPlayer {
             }
         }
 
-        animated_transforms
+        self.apply_post_transform(animated_transforms)
+    }
+
+    fn apply_post_transform(&self, mut transforms: [Matrix4<f32>; 40]) -> [Matrix4<f32>; 40] {
+        if let Some(post) = self.post_transform {
+            for transform in transforms.iter_mut() {
+                *transform = post * *transform;
+            }
+        }
+        transforms
     }
 
     fn compute_transforms_for_clip(
--- a/dark/src/motion/animation_player.rs
+++ b/dark/src/motion/animation_player.rs
@@ -73,12 +73,6 @@ pub struct AnimationPlayer {
     /// `AnimationInfo::cancel_root_motion`). Set for first-person viewmodels,
     /// whose entity transform is re-anchored to the camera every frame.
     cancel_root_motion: bool,
-    /// Rigid model-space transform composed onto every posed joint after
-    /// animation (`get_transforms` returns `post * joint`). Used by the VR
-    /// melee wield to align the posed arm's fist and weapon axis with the
-    /// entity's grip frame without moving the entity (and its contact
-    /// collider). `None` for every ordinary player.
-    post_transform: Option<Matrix4<f32>>,
 }
 
 impl AnimationPlayer {
@@ -93,7 +87,6 @@ impl AnimationPlayer {
             blend_state: None,
             rotation_pos: 0.0,
             cancel_root_motion: false,
-            post_transform: None,
         }
     }
     pub fn from_animation(animation_clip: Rc<AnimationClip>) -> AnimationPlayer {
@@ -108,7 +101,6 @@ impl AnimationPlayer {
             blend_state: None,
             rotation_pos: 0.0,
             cancel_root_motion: false,
-            post_transform: None,
         }
     }
 
@@ -128,7 +120,6 @@ impl AnimationPlayer {
             blend_state: None,
             rotation_pos: 0.0,
             cancel_root_motion: false,
-            post_transform: None,
         }
     }
 
@@ -174,7 +165,6 @@ impl AnimationPlayer {
             blend_state,
             rotation_pos: 0.0,
             cancel_root_motion: player.cancel_root_motion,
-            post_transform: player.post_transform,
         }
     }
 
@@ -247,7 +237,6 @@ impl AnimationPlayer {
             blend_state,
             rotation_pos: 0.0,
             cancel_root_motion: player.cancel_root_motion,
-            post_transform: player.post_transform,
         }
     }
 
@@ -256,17 +245,6 @@ impl AnimationPlayer {
     pub fn with_root_motion_cancelled(player: &AnimationPlayer) -> AnimationPlayer {
         let mut new_player = player.clone();
         new_player.cancel_root_motion = true;
-        new_player
-    }
-
-    /// A copy of `player` whose posed joints are all pre-multiplied by
-    /// `transform` (see the `post_transform` field).
-    pub fn with_post_transform(
-        player: &AnimationPlayer,
-        transform: Matrix4<f32>,
-    ) -> AnimationPlayer {
-        let mut new_player = player.clone();
-        new_player.post_transform = Some(transform);
         new_player
     }
 
@@ -287,7 +265,6 @@ impl AnimationPlayer {
             blend_state: player.blend_state.clone(),
             rotation_pos: player.rotation_pos,
             cancel_root_motion: player.cancel_root_motion,
-            post_transform: player.post_transform,
         }
     }
 
@@ -424,7 +401,6 @@ impl AnimationPlayer {
                                 blend_state,
                                 rotation_pos: raw_end_pos - current_clip.num_frames as f32,
                                 cancel_root_motion: player.cancel_root_motion,
-                                post_transform: player.post_transform,
                             },
                             motion_flags,
                             events,
@@ -462,7 +438,6 @@ impl AnimationPlayer {
                                 // slice is emitted on its first tick.
                                 rotation_pos: 0.0,
                                 cancel_root_motion: player.cancel_root_motion,
-                                post_transform: player.post_transform,
                             },
                             motion_flags,
                             events,
@@ -493,7 +468,6 @@ impl AnimationPlayer {
                         blend_state,
                         rotation_pos: raw_end_pos,
                         cancel_root_motion: player.cancel_root_motion,
-                        post_transform: player.post_transform,
                     },
                     motion_flags,
                     events,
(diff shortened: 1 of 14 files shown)


tommy-xr:
Both of your findings were real. Measured, then fixed — details and stills are now in the PR body; the short version:

**1. Hand orientation.** The tracked hand's own frame is defined by `hand_forearm`: it hangs the empty hand's sleeve along the hand's **+Z**, with the fingers/aim down **-Z**. Against that axis the shipped seating put the posed forearm:

| model | off the hand's arm axis | correcting axis |
| --- | --- | --- |
| `wrench_h` | **71.3°** | (0.28, 0.96, 0.00) — a yaw |
| `rapier_h` | **79.8°** | (0.20, 0.98, 0.00) |
| `shard_h` | **79.8°** | (0.20, 0.98, 0.00) |
| `psword_h` | **79.3°** | (0.19, 0.98, 0.00) |

The arm left the wrist sideways instead of running back toward the elbow. All four are nearly identical because the four rigs share one posed arm; only the weapon joint differs.

It is **not** a different authoring axis from the guns, and it is not a wrist rotation in the idle pose I posed them to. The #1023 guns are standalone LGMD meshes authored barrel-along -X, and one -90° yaw maps -X onto the hand's -Z aim axis — for a gun that single DOF *is* the whole job. A melee `_h` is a skinned **arm**, so seating it is 3-DOF (arm onto the arm axis, plus a roll deciding where the weapon points). One authored yaw can only buy one of the two, and the 120° I picked bought the weapon and paid with the arm.

The alignment is now derived from the rig rather than authored: rotate the posed forearm onto +Z, then roll about it until the weapon lies in the hand's vertical plane, standing up out of the fist. Forearm error after: **< 0.5°** on all four, unit-tested. Before/after and a refreshed contact sheet (with an empty-hand panel at the same pose, for comparison) are in the body.

**2. Hitbox vs visible weapon — your suspicion was correct.** `Model::apply_local_transform` sets `local_transform` on *every* scene object of the model, so the render correction moved the whole wield — arm **and** weapon geometry — while the collider stayed at the entity transform. In hand-local space:

| model | contact sphere | rendered weapon head | gap |
| --- | --- | --- | --- |
| `wrench_h` | (0, 0, -0.40) | (0.013, 0.591, -0.556) | **0.61** — and 0.29 from the *nearest* point of the shaft |
| `rapier_h` | (0, 0, 0) | (-0.032, 1.096, -0.416) | **1.17** |
| `shard_h` | (0, 0, 0) | (-0.154, 1.184, -0.339) | **1.24** |
| `psword_h` | (0, 0, 0) | (-0.159, 1.142, -0.290) | **1.19** |

The Wrench's 4.6 cm damage sphere was floating in clear air beside the weapon — and yes, the damage e2e stayed green because it exercises the collider. (`/v1/scene` can't show this either: it reports `SceneObject::transform`, not `transform * local_transform`, so it printed the entity position for both.)

The fix falls out of the mechanism the earlier commit set up. Since the correction *cancels* the grip, a melee `_h`'s grip entry moves no pixels at all — it only places the body. So it is now set to exactly where the rendered weapon head lands, and visual and physical coincide by construction; the live measurement matches the table to the millimetre on all four. Two asset-free tests keep the halves from drifting apart, and `cargo run -p shock2vr --example melee_grip` re-measures the constants if the rigs ever change.

**Damage still holds:** `medsci-saved-vr-melee` is green — exactly **1 Damage / 9 HP** per pull, plus the three-pull OG-Shotgun kill and the authored world pane. Its gesture needed one honest update: it staged the *hand* at the sweep distance, which only connected while the collider happened to sit 0.4 along the fingers; it now measures the held weapon's grip offset live and backs the hand off by it, so the sweep does what its comment always claimed — bring the weapon's contact volume to 0.33 units from the torso. Also green: `vr-melee-contact` (3), `vr-held-model` (5), and the whole `vr-*` + `weapon-*` set (48 tests), plus 817 lib tests and `-D warnings`.

Still open from before, unchanged: the empty hand is the pale 25AE bare-skin hand while the `_h` meshes bake a dark-skinned hand in a black sleeve — the two read as different characters (visible in the contact sheet's last panel).

tommy-xr:
Follow-up to the measurements above: `/xreview` (Claude + Codex, plus the de-dup pass) went over the fix and both engines independently landed on the same structural complaint, which was fair — so the second half is now built differently, and three further defects came out of it.

**The collider placement is now derived, not measured.** My first cut computed the *render* correction live from the posed skeleton but pinned the *collider* to four hand-copied constants — so a modded or updated `_h` would move the drawn weapon and leave the damage volume behind, with every test still green (the drift test compared two pasted copies of the same numbers). The wield now computes the contact offset where it poses the arm and stores it as `RuntimePropVrGripOffset`, which the grip lookup prefers; the correction cancels that same value. Both halves come from one posed arm, so the weapon joint renders exactly on the collider **for any rig** — a property rather than a measurement someone has to maintain. The four melee entries are gone from `HAND_MODEL_POSITIONING` entirely, which as a side effect restores the 180° projectile rotation they had silently changed. The coupling test is now a composed-matrix assertion (entity transform × render correction ⇒ weapon joint on the collider, fist on the hand) for every shipped rig.

**A drop regression I had missed.** Because the held entity now sits on the weapon head, releasing one respawned the loose prop out there: a dropped Crystal Shard materialized 1.2 units above the palm and fell. `DropItem` now puts it back on the hand first — measured: release at hand+1.22 → hand+0.41, which is where every other drop already lands.

**Two smaller ones:** the joint indexing is guarded (a short or patched rig logs and wields unseated instead of panicking the frame), and the e2e's staging no longer quietly compensates for a misplaced collider — it asserts the measured offset's shape (~0.8 out, standing up out of the fist) and takes its hand rotation from the SDK's level-horizon `lookQuat`, since shortest-arc-from-−Z rolls up to ~180° for directions near world +z, which is exactly where those staged targets sit.

**One limitation I deliberately did not fix here, filed as #1039:** the authored contact volume is a ~5 cm *sphere* (Wrench 0.046, Electro Shock 0.052, Crystal Shard 0.050), so this arms the weapon's **head** and nothing else along its length — a swing that connects mid-blade still passes through. Arming the whole weapon needs a collider shaped like it (a capsule fist→head, geometry the wield already computes). That is a real change to the held body's shape, with knock-on effects on how the weapon shoves actors, so it wants its own pass rather than riding along here. Happy to take it next if you'd rather have it in this PR.

**Validation after all of that:** `medsci-saved-vr-melee` still measures exactly **1 Damage / 9 HP** per pull (plus the three-pull OG-Shotgun kill and the world pane); `vr-*` + `weapon-*` = **49 e2e tests green**, including 23/23 missions; 817 lib tests; `cargo fmt --all --check` and `-D warnings` clean.

Issue #1034, opened 2026-08-19
vr: weapons with no muzzle vhot fire from the grip, and classic world models are barrel-along-Z


tommy-xr:
Split out of #1033 / PR for `feat/vr-projectile-origin`, which made VR projectiles leave the model's muzzle vhot along the model's -X barrel. Two pre-existing gaps that fix does **not** close (behavior there is unchanged, not newly broken):

## 1. Four 25AE view models ship with zero vhots

`empgun_h`, `gren_h`, `fsn_h`, `al_h` carry no vhot at all (verified by decoding `mods/sshock2ee.kpf`), so `create_projectile` takes its documented fallback and fires from the **model origin - the grip**.

Measured in `debug_weapons --vr`, EMP rifle (`empgun_h`):

| hold | result |
|---|---|
| hand out at `z = -2.5`, clear of the body | `EMP Shot` flies the full 11 units to the wall |
| hand in at the body | `Blue Pulse` + `EMP Explosion` spawn at the weapon origin and **detonate immediately** - the player EMPs themselves |

The grenade launcher (`gren_h`) survives the same close hold because its projectile is slow and physical, but it too leaves from the grip rather than the tube.

Possible fixes: derive a fallback fire point from the model's -X bbox extreme, or give the spawn the same kind of clearance the flat branch uses (`FLAT_MUZZLE_CLEARANCE`).

## 2. Several classic-install world models are authored barrel-along-Z

VR falls back to world models when the 25AE view models are unavailable. `weapon_script`'s -X barrel axis is wrong for these (engine-space half-extents, from the mesh headers):

| model | half-extents (x, y, z) | vhots |
|---|---|---|
| `atek_w` | 0.11, 0.46, **0.75** | 1, at -Z |
| `ar15_w` | 0.20, 0.85, **3.00** | 2: type 1 at the -Z tip, type 0 at the breech |
| `sg_w` | 0.23, 0.47, **3.01** | 1 |
| `gren_w` / `viro_w` / `al_w` | Z-long | none |

So on a classic install those six fire 90 degrees off the rendered barrel. (`laser`, `empgun`, `fsn_w`, `sfg_w`, `amp_w` are X-authored and are fine.)

## 3. `vhots.first()` vs the loader's type sort

[...]

PR #1035, opened 2026-08-19
fix(vr): fire every VR weapon from its muzzle vhot down its own barrel

### LINKED ISSUE #1033: vr: laser bolt does not leave the laser pistol's muzzle vhot
## Symptom

In VR, the laser pistol's bolt does not come out of the barrel. It leaves the weapon off-axis and, held anywhere near the body, spangs on the player's own collider a fraction of a unit out instead of reaching the target.

## Diagnosis (headless, `debug_weapons --vr`)

The VR hand grabs the pistol (25AE view model `lasehand`), is posed pointing along -X, and fires. Measured from the runtime:

| | value |
|---|---|
| muzzle vhot, world | `(-0.495, 2.336, -1.936)` |
| rendered barrel axis | `(-1, 0, 0)` |
| bolt travel | 12.00 degrees off the barrel |
| bolt origin | 0.16 world units off the muzzle |
| miss at 11 units of range | 2.5 world units |
| fired from a hold near the body | spangs on the player 0.43 units out |

Root cause: `vr_config`'s per-model `projectile_rotation` carries a 12-degree fudge for `laser`/`lasehand`. It dates from when VR wielded the `laser` **world** model, which has zero vhots, so the bolt spawned at the grip and the fudge nudged it. Since #1023 VR wields `lasehand`, which does carry a real muzzle vhot at the -X barrel tip, so the fudge is pure error - applied to both the spawn point and the travel direction.

The surrounding construction is what allows this: `create_projectile` reaches the muzzle by applying the hand grip rotation a second time and undoing it on the spawn point, leaving `projectile_rotation` as a per-weapon escape hatch.

## Expected

A VR weapon's projectile leaves the model's muzzle vhot travelling along the model's rendered barrel, with no per-weapon aim correction.

tommy-xr:
Fixes #1033.

## The bug

In VR the laser pistol's bolt did not come out of the barrel. Held anywhere near the body it crossed in front of the player and spanged on their own collider instead of reaching the target.

Measured headlessly in `debug_weapons --vr` (VR hand grabs the pistol, is posed pointing along -X, fires):

| | before | after |
|---|---|---|
| travel vs. rendered barrel | **12.00 degrees off** | **0.00 degrees** |
| spawn vs. muzzle vhot | **0.16 world units off** | **< 0.01** |
| miss at 11 units of range | **2.5 world units** | 0 |
| fired from a hold at the body | **spangs on the player, 0.43 units out** | flies the full 11 units to the wall |

VR laser bolt leaving the muzzle

<sub>AFTER: the VR hand holds the laser pistol in front of the camera in `debug_weapons --vr`, aims down -X and fires; the bolt leaves the visible muzzle and flies the full 11 units into the wall. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

### Before -> after: the bolt vs. the barrel

before/after bolt path with the weapon in frame

<sub>Same deterministic sequence on both builds - same grab, same hand pose, same trigger frame - with the frames of the bolt's flight blended into one image. Left: the pre-fix bolt starts detached from the gun and angles away. Right: the fixed bolt emerges from the muzzle and recedes along the barrel.</sub>

<sub>**Read this one as "does it come out of the gun", not as an angle.** The camera necessarily looks *down* the barrel, which foreshortens the 12 degrees; a genuinely side-on framing isn't available, because `debug_weapons` only has geometry toward -X and turning to get one fills the frame with void. The panel below is the decisive evidence.</sub>

### Before -> after: where it lands

before/after impact point on the wall

<sub>The wall at the moment of impact (16 steps after firing), magnified. Dashed yellow marks where the rendered barrel points; the blue circle marks the spang. Before, the impact is ~1.9 world units off that line - the bolt spawns at z 0.22 and drifts +0.17 per step to z 2.18 by the wall. After, z holds at 0.28 from muzzle to wall: no drift at all. The residual y difference between the two (2.14 vs 2.29) is the 0.16-unit muzzle-vhot offset the old code also got wrong.</sub>

## Root cause

`create_projectile`'s VR branch reached the muzzle by applying the hand grip rotation a **second** time and undoing it on the spawn point:

```rust
position       = grip_rotation.invert() * vhot
root_transform = weapon_transform * grip_rotation * projectile_rotation
```

`grip_rotation` is already baked into `weapon_transform` by `virtual_hand`. The construction only landed on the barrel because every gun grip happens to be a -90 degree yaw — and it left `projectile_rotation` as a per-weapon escape hatch. The laser pistol carried 12 degrees of it, from the era when VR wielded the **`laser` world model, which has zero vhots**: the bolt spawned at the grip and the fudge nudged it. Since #1023 VR wields `lasehand`, which does carry a real muzzle vhot at the -X barrel tip, so the fudge became pure error — applied to *both* the spawn point and the direction.

## The fix

One faithful, shared mechanism, no per-weapon special cases: spawn at the model's muzzle vhot, travelling along the model's **-X** — the axis Dark's first-person gun models author their barrel on, where each of them puts its muzzle vhot, and the axis every grip entry's -90 degree yaw exists to point out of the hand.

```rust
root_transform: weapon_transform
    * Matrix4::from_translation(muzzle_vhot)
    * Matrix4::from(barrel_axis_from_forward())   // yaw -90: +Z travel axis -> model -X
```

`projectile_rotation` is **deleted** rather than retuned, so no weapon can carry a private aim correction again, and a `vr_config` test pins every VR-wieldable gun's grip to that convention.

**Documented fallback**: a model with no vhot fires from the model origin (the grip). This covers `empgun_h`, `gren_h`, `fsn_h`, `al_h` (which ship with zero vhots) and the classic world models — unchanged from before, and tracked in #1034.

**Flat is untouched** — it returns earlier, from the crosshair ray. Verified: `world-aim`, `muzzle-flash`, `projectile-trail`, `close-range-projectile`, `slow-projectile-spang`, `ally-line-of-fire` and all `weapon-*` e2e tests pass unchanged.

## The psi amp

The backlog also listed "vr: psi amp cry doesn't come out (maybe reversed?)". **It is not reproducible on current `main`** — I measured the amp firing from its `amp_h` muzzle vhot exactly along the barrel (0.00 degrees deviation) in four hand poses including rolled and off-axis ones, and in both hands. It was almost certainly fixed by #1023, which moved VR from the offset-less `amp_w` world model to `amp_h` with a fitted grip. Rather than leave that to rot, this PR pins it: the psi-amp e2e case is new here and passes on the pre-fix build too.

## Verification

Negative-first, as required:

- **e2e (`tools/shock2-sdk/test/vr-muzzle-origin.e2e.test.ts`)** — drives the production VR hand (grab, pose, trigger) and asserts, against the weapon's live transform, that the shot's heading is within 1 degree of the rendered barrel, that its flight path passes within 0.05 units of the muzzle vhot laterally, and that its axial distance matches the elapsed frames of flight (so a spawn displaced *along* the barrel is caught too). **Confirmed failing on pre-fix code** at exactly `12.00 degrees`; green after.
- **Rust unit tests** — `a_vr_shot_leaves_the_muzzle_vhot_along_the_barrel` checks the returned `CreateEntity` numerically in an awkward yawed/pitched/translated pose; plus the vhot-less fallback, and the `vr_config` grip-convention test.
- `cargo fmt --all --check`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`, `cargo test -p shock2vr --lib` (817 pass), and `missions.e2e.test.ts` (all 23 missions load).

## Review

`/xreview` (Claude + Codex + a dedup pass) found no ship-blocking correctness issue in the production change, and both engines agreed on two test-quality points, both applied: the origin assertion could not detect an along-barrel displacement, and the test duplicated `test/helpers/vr-hand.ts`'s quaternion math for the third time. The Claude reviewer additionally decoded the shipped meshes and showed the doc/test were certifying the -X barrel convention over world models where it is false — corrected in the second commit, and filed as #1034 along with the vhot-less view models and the `vhots.first()` vs. type-sort hazard.

### DIFF
--- a/shock2vr/src/scripts/weapon_script.rs
+++ b/shock2vr/src/scripts/weapon_script.rs
@@ -33,6 +33,17 @@ const MELEE_DAMAGE: f32 = 6.0;
 /// guns for now; per-weapon loudness is a follow-up.
 const GUNSHOT_NOISE_RADIUS: f32 = 50.0 / SCALE_FACTOR;
 
+/// Rotation taking the projectile's +Z travel axis onto a Dark gun model's
+/// barrel, which is authored along the model's **-X**. Every VR grip entry's
+/// -90 degree yaw exists to point that same axis out of the hand, and every
+/// wieldable model's muzzle vhot sits at its -X extreme (see the
+/// `vr_config` grip table and its `gun_grips_aim_the_barrel_out_of_the_hand`
+/// test), so this one rotation aims every VR weapon - ballistic, energy and
+/// psi alike - down its own rendered barrel.
+fn barrel_axis_from_forward() -> Quaternion<f32> {
+    Quaternion::from_angle_y(Deg(-90.0))
+}
+
 /// The weapon entity's current world position (from its live transform).
 fn weapon_world_position(world: &World, entity_id: EntityId) -> Option<cgmath::Vector3<f32>> {
     let v_transform = world.borrow::<View<RuntimePropTransform>>().ok()?;
@@ -361,6 +372,10 @@ pub(super) fn create_projectile(
         }
     }
 
+    // VR: the weapon is a physical object in the hand, so the shot is defined
+    // entirely by the *rendered* weapon - it leaves the model's muzzle vhot
+    // travelling down the model's barrel. Both are read from the weapon's own
+    // live transform, so no per-weapon aim correction is involved.
     let v_transform = world.borrow::<View<RuntimePropTransform>>().unwrap();
     let v_vhots = world.borrow::<View<RuntimePropVhots>>().unwrap();
 
@@ -368,37 +383,29 @@ pub(super) fn create_projectile(
         .get(entity_id)
         .map(|vhots| vhots.0.clone())
         .unwrap_or_default();
-    let vhot = vhots
-        .get(0)
+    // The fire point is the model's first vhot: every wieldable gun/amp model
+    // authors its muzzle there, at the -X tip of the barrel. Documented
+    // fallback for a model that carries no vhot at all (the classic `laser`
+    // and `lasehand` meshes, for instance): the model origin, i.e. the grip -
+    // the shot still leaves along the barrel, just from the hand.
+    let muzzle = vhots
+        .first()
         .map(|v| v.point)
         .unwrap_or(point3(0.0, 0.0, 0.0));
 
     let transform = v_transform.get(entity_id).unwrap();
 
-    let adjustments = vr_config::get_vr_hand_model_adjustments_from_entity(
-        entity_id,
-        world,
-        // TODO: I guess we don't care about handedness for now,
-        // since it only affects the flipping of the weapon... but truly we should consider it.
-        vr_config::Handedness::Left,
-    );
-
-    let rotation = adjustments.rotation;
-    let projectile_rotation: Matrix4<f32> =
-        vr_config::get_projectile_rotation_from_entity(entity_id, world).into();
-    let rot_matrix: Matrix4<f32> = rotation.into();
-    let inv_rot_matrix: Matrix4<f32> = rotation.invert().into();
-
-    // Adjust the vhot position to be in the same coordinate space as the weapon
-    let position = inv_rot_matrix.transform_point(vhot);
-
     Effect::CreateEntity {
         template_id: projectile_template_id,
-        position,
+        position: point3(0.0, 0.0, 0.0),
         // HACK: Not sure why we need to do this, but seems projectile
         // models are rotated 90 degrees
         orientation: Quaternion::from_angle_y(Deg(90.0)),
-        root_transform: transform.0 * rot_matrix * projectile_rotation,
+        // Projectile velocity is `root_transform * (0, 0, magnitude)`, so put
+        // the muzzle at the origin and aim +Z down the barrel.
+        root_transform: transform.0
+            * Matrix4::from_translation(muzzle.to_vec())
+            * Matrix4::from(barrel_axis_from_forward()),
         options: CreateEntityOptions {
             force_visible: true,
             ..CreateEntityOptions::default()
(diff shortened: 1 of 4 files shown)


tommy-xr:
## Restacked on `main` (past #1032)

Rebased onto `origin/main` (now `8fbdd8ae`, the melee `_h` wield). All conflicts were in `shock2vr/src/vr_config.rs`; `weapon_script.rs`, `helpers/vr-hand.ts` and `vr-muzzle-origin.e2e.test.ts` applied untouched.

**How the two changes met.** They only *look* like they collide. #1032 deleted the four melee grip entries so the wield can compute the body placement per-rig (`RuntimePropVrGripOffset`), and noted that as a side effect this "restored their 180-degree projectile rotation". This PR deletes `projectile_rotation` outright — so that side effect no longer exists to restore, and it never mattered: melee weapons do contact damage and never reach `create_projectile`. Both intents are preserved, neither is weakened:

- **#1032 kept whole.** `melee_wield_alignment` / `melee_contact_offset` / `melee_wield_pose_correction`, `RuntimePropVrGripOffset` precedence in the grip lookup, and all four of its tests survive unchanged — except the one assertion in `melee_view_models_have_no_static_grip` that pinned `get_vr_projectile_rotation_from_model` to 180°, which is dead with the function it called. The test's real claim (a melee `_h` must have **no** static grip entry) is untouched. The stale trailing sentence in the table comment went with it.
- **#1035 kept whole.** No per-weapon projectile fudge; VR fires from the muzzle vhot down the model's authored −X barrel.
- **One genuine merge, in the new test.** `gun_grips_aim_the_barrel_out_of_the_hand` iterates `VR_25AE_VIEW_MODELS`, which #1032 grew by the four melee rigs. It now filters `MELEE_VIEW_MODELS` out — they are exempt twice over: they fire nothing, and they deliberately have no table entry, so asserting a barrel axis on their (absent) grip would just assert the identity rotation.
- `let default = ...` is gone, as on `main` — its only consumer was the deleted `wrench_h` entry.

`git range-diff` against the pre-rebase tip: commit 1 identical (`=`); commits 2 and 3 differ **only** by the resolutions listed above. No content silently dropped.

**Re-validated on the rebased branch**

| check | result |
| --- | --- |
| `cargo fmt --all --check` | clean |
| `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` | clean |
| `cargo test -p shock2vr --lib` | 820 passed, 0 failed |
| `vr-muzzle-origin` e2e | 2/2 pass |
| `medsci-saved-vr-melee`, `vr-melee-contact`, `vr-held-model` | 9/9 pass — #1032's canonical **1 Damage / 9 HP** still measured |
| weapon / aim / projectile / psi e2e (15 files) | 17/17 pass |
| `missions.e2e.test.ts` | 23/23 pass |

**The PR's own numbers still hold**, measured after the rebase:

| shot | barrel deviation | lateral miss of the muzzle vhot | axial distance after 2 frames |
| --- | --- | --- | --- |
| Laser Pistol bolt | **0.000°** | **0.0000** | 0.8000 (frame step 0.4000) |
| Psi Amp cryo bolt | **0.000°** | **0.0000** | 0.4000 (frame step 0.2000) |

Nothing in the melee change shifted them.

Issue #1039, opened 2026-08-19
VR melee: arm the whole weapon, not just its head (contact volume is a ~5cm sphere)


tommy-xr:
## What

A VR-wielded melee weapon's damage volume is the held entity's authored physics shape, and for every melee weapon that shape is a small **sphere**:

| weapon | `PropPhysDimensions` |
| --- | --- |
| Wrench | SPHERE, radius 0.046 |
| Electro Shock | SPHERE, radius 0.052 |
| Crystal Shard | SPHERE, radius 0.050 |

PR #1032 put that sphere where it belongs — on the rendered weapon head (it used to float up to 1.2 units off the weapon entirely) — but a ~5 cm ball only arms the *head*. The rest of a 0.8-1.2 unit blade or shaft is inert: a swing that connects mid-blade passes through the target without a contact, and the player has no way to tell which part of the weapon is live.

## Why it is not just "make the sphere bigger"

A bigger sphere is a bigger *ball*, so it starts hitting things the weapon is nowhere near (and the held body also shoves actors — `medsci-saved-vr-melee` asserts bounded actor displacement). The shape has to follow the weapon: a **capsule spanning fist -> weapon joint**, which is exactly the geometry the wield already computes (`vr_config::melee_contact_offset` derives the head from the posed rig; the fist is the other end).

## Sketch

- On the melee `_h` wield, replace the held body's collider with a capsule from the posed fist to the posed weapon joint, radius ~ the authored sphere.
- Restore the authored shape on drop (the release path already rebuilds authored physics via `make_un_physical` + `make_physical`).
- `physics::resize_kinematic_cuboid` is the existing precedent for reshaping a live kinematic collider.

## Verification

- The existing melee damage e2e must keep measuring exactly 1 Damage / 9 HP per pull (no double hits along the capsule — `TriggeredMeleeWeapon` already dedupes per victim per trigger pull).
- Add a sweep that connects **mid-blade** and asserts damage; it fails today.
- Re-check `assertBoundedActor` in `medsci-saved-vr-melee`: a long capsule has more leverage on an actor than a point.

Found during review of #1032 (both review engines flagged the tip-only placement); deliberately out of scope there, which was about getting the volume onto the weapon at all.

tommy-xr:
Correction to this issue's premise, found while tracing the owner-reported "hurtbox not working in VR" (draft PR #1045).

[...]

PR #1045, opened 2026-08-19
feat(debug): world-space melee contact-volume overlay, and why VR melee misses

### LINKED ISSUE #1048: fix(vr): melee needs a contact that BEGINS after the trigger pull, so an already-touching swing does nothing
Measured on `feat/hurtbox-debug-view` (#1045) with the new melee contact-volume overlay and with `Collided` restored to the message trace.

## What happens

`TriggeredMeleeWeapon` (`shock2vr/src/scripts/melee_weapon.rs:76-83`) damages only on a `Collided` message, and `Collided` is dispatched only from Rapier's `CollisionEvent::CollisionStarted` edge (`shock2vr/src/mission/mission_core.rs:2182-2194`). Rapier does not re-emit `Started` for a contact that never ended.

So if the weapon's contact volume is **already intersecting the victim** when the trigger is pulled, no new `Started` is generated and **no damage is ever dealt**, for as long as the hold lasts.

## Evidence (medsci2, VR, shipped Blue Monkey 543, gamesys Wrench -928)

Park the contact volume on the monkey's torso, dwell 30 frames, *then* pull the trigger and hold for 30 frames:

```
ALREADY-TOUCHING: hp 10 -> 10   messages to victim = ["Collided","Collided","Collided"]
```

Three contacts are generated and delivered to the script - they just arrive before the window opens. HP does not move.

The same weapon and the same victim, armed from out of contact and then swung in, takes its authored 9 HP:

```
FRESH-SWING (12 samples, 1 frame each): hp 10 -> 1   messages = ["Damage"]
```

## Why the harness never caught it

Every existing melee e2e arms the trigger from several units away and then approaches, which always produces a fresh `Started`. In a headset the weapon is a 1.83-unit slab (see the companion issue) that is very often already touching whatever you are about to hit.

## Suggested fix

On `TriggerPull`, also damage whatever the weapon's collider currently overlaps (a Rapier intersection query at the pull edge), not only contacts that begin afterwards. Note this is **orthogonal** to the spring-driven physical-melee direction: simulating the weapon toward the hand instead of teleporting it does not by itself make an existing contact re-fire.

Related: #955 (bound the attack window to a swing), #1039, #1045.

tommy-xr:
**A debug overlay for melee contact volumes, toggleable from Developer options — plus the diagnosis of why VR melee misses.**

Nothing in the codebase could show **where** a melee weapon's damage volume actually is, or **how big** it is. This adds that instrument, and uses it to answer the question with numbers. It changes **no gameplay behavior** — the overlay is off by default and nothing else in the frame is touched.

The fix for what it reveals is **#1053** (physical kinematic melee: a spring-driven weapon body with a collider that envelops the weapon, fitted to the `_h` geometry the player actually sees). This PR is the instrument for that work, not a fix for it.

melee contact-volume overlay

> ### That box really is that wrong — it is the defect, not a debug-draw bug
>
> The red wireframe is read **straight out of Rapier** (`debug_entity_collider_volumes` reports what is actually simulated, deliberately *not* a recomputation of what the wield intended — a recomputation would agree with the renderer and hide exactly this). Its placement is correct: it is drawn at the collider's own live isometry. What it shows is a genuinely broken volume:
>
> - **It is a cuboid of the *world* model's bounding box**, floor-clamped to **0.2 units per axis** — `entity_creator.rs:1086`, taken because a Wrench is a frobbable `FrobFlag::MOVE` item. Measured: **1.83 × 0.39 × 0.20** world units. The shipped Blue Monkey's entire actor capsule is **1.20 × 1.80 × 1.20**, so *the wrench's damage volume is as long as the creature it hits is tall* — roughly 3-4× the rendered weapon. The 0.20 axis is exactly the clamp (`min_size = 0.5 / SCALE_FACTOR`), so the thin axes are inflated on top of a bounding box that already fits a long thin wrench badly.
> - **It is derived from `_w` while the player sees `_h`.** The gamesys Wrench declares `PropModelName("wrench_w")` and `PropLimbModel("wrench_h")`. The collider is built once at entity creation from the **world** model; the VR wield (#1023/#1032) renders the **first-person** model, and `set_held_melee` changes body type, active collision types and groups — never the shape. Two different meshes with different proportions and origins, which is why the box looks unrelated to the wrench in your hand. **This is a distinct defect from the sizing**, and distinct again from tunneling.
> - It is **not** the authored `PropPhysDimensions` sphere (radius 0.046) that #1039 and #1032's text both describe. That path is reachable only by launched projectiles and non-frobbable objects. #1039's premise is wrong in the *opposite* direction — a correction is posted there.
>
> In the clip: the volume resting inside a Blue Monkey with the trigger off; the trigger **pulled** while it is still buried (HP stays 10); then a swing armed out of contact (HP 10 → 1). Captured headlessly (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).

contact that damages vs contact that does not

### The toggle, where the owner asked for it

Developer screen toggle in flat and VR

<sub>Render-verified in **both** presentations per AGENTS.md §3: the row lands in the same place, reads the same, and the switch actually flips from the screen — with the pointer flat, with the controller ray in VR (`melee_volumes=1` reported by `GET /v1/dev-params` after each click).</sub>

## What this adds

- **`PhysicsWorld::debug_entity_collider_volumes`** — live world-space geometry (ball / capsule / cuboid + isometry) of every collider on an entity's primary body, read from Rapier rather than recomputed.
- **`mission::melee_debug`** — draws those volumes as a world-space wireframe, reusing `dark::hit_box`'s existing line builders (the same ones behind `cargo dv --debug-hitboxes`) rather than growing a parallel debug-draw system. Emitted in the **shared world pass**, so flat and VR render the identical overlay — required, since the failure was reported in VR.
- **`DevParamKind::Bool`, and the overlay as a Developer-screen switch.** The registry's own comment always anticipated this kind; it is added properly rather than smuggled in as a 0/1 float, so the menu renders **On/Off** instead of "1.00", either arrow flips it, and `GET /v1/dev-params` reports `kind: "bool"`. The overlay is the first consumer (`melee_volumes`). No rebuild to toggle it; on the Quest, no redeploy.
- **`shape` + `shape_extents` on `GET /v1/physics/bodies`** (and the SDK type). Every field there said where a body *is*; none said how big it is, so "how large is this damage volume?" could only be eyeballed from a screenshot. One AABB for every shape kind, so the number means the same thing for a ball, a capsule and a cuboid. This is what produced the measurements above.
- **`Collided` restored to the message trace**, carrying the entity collided with. While it was filtered out, a swing that never touched its target and a swing that touched it but was disallowed from damaging looked **identical** — both showed nothing, since only the resulting `Damage` was visible. Bounded so it cannot displace event history: contacts dedupe per *pair* within 30 frames, and on ring overflow the oldest **contact** is dropped ahead of the oldest anything else. That bound is not theoretical — tracing contacts without it made `medsci-saved-vr-melee`'s Damage assertion fail under load, because the `Damage` it counted had been evicted before the test read it.

## Diagnosis — measured, and it corrects the earlier one

I set out to prove the tunneling hypothesis (a kinematic weapon teleported by `set_next_kinematic_position` with CCD off, contacts evaluated at the end pose only). **The measurements falsified it as the operative cause.**

**Tunneling does not bite at human speeds.** Sweeping the volume through a shipped one-HP pane, straddling the pane plane so no pose lands on it by construction:

| collider travel per 60 Hz step | ≈ speed | pane destroyed |
| --- | --- | --- |
| 0.10 | 6 m/s | yes |
| 0.30 | 18 m/s | yes |
| 0.60 | 36 m/s | yes |
| 1.00 | 60 m/s | **no** |
| 2.00 | 120 m/s | **no** |

The threshold is 36-60 m/s of *collider* travel — far above anything an arm produces, precisely because the volume is so oversized. A 12 m/s swing at a creature, one sample per frame with no dwell, lands its authored 9 HP every time. So the teleport model is a real structural fragility but **not** why swings missed. An interim CCD fix was therefore deliberately **not** taken: it patches a gap that only opens above 36 m/s.

**The actual cause is edge-only semantics (#1048).** `TriggeredMeleeWeapon` damages only on `Collided`, which is dispatched only from Rapier's `CollisionStarted` edge; Rapier never re-emits `Started` for a contact that never ended. So if the volume is already inside the victim when you pull the trigger, nothing happens for the whole hold:

```
ALREADY-TOUCHING  hp 10 -> 10   messages to victim = ["Collided","Collided","Collided"]
FRESH SWING       hp 10 ->  1   messages to victim = ["Damage"]
```

The `Collided` entries prove this is not a "no contact generated" failure: contact **is** generated and **is** delivered to the script — it simply arrives outside the window. And it bites in a headset precisely because of the oversized volume: a 1.83-unit box sitting 0.81 units out of your palm is touching something most of the time. The harness never reproduced it because every existing melee e2e arms the trigger several units away and then approaches, manufacturing a fresh edge.

### Generated-but-not-forwarded, or never generated?

**Neither, as a blanket answer.** Contacts are generated, delivered, and honoured whenever they begin inside the window. The losing case is "never generated" only in the narrow sense that no *new* `Started` exists to route — the physics is correct; the semantics discard it.

### The flat/VR conditional is benign

`TriggeredMeleeWeapon::handle_message` returns `NoEffect` unless `is_vr(world)`. That gates the contact path **on** in VR and **off** in flat — the opposite of "works flat, diverted in VR" — and it is deliberate: flat melee is `weapon_script::flat_melee_hit`'s aimed raycast, so bumping a wielded wrench into scenery must do nothing. Every measurement above was taken in `--vr`. No presentation conditional causes the miss.

### What this hands #1053

- The contact→damage path is **correct** and needs no change; a spring-driven body should light it up essentially unchanged.
- **But the spring alone does not fix #1048**: simulating rather than teleporting still never re-emits `Started` for a contact that persists across the trigger pull. The pull edge must also damage what the weapon already overlaps, whatever drives the body.
- Solver impulse against actors is already suppressed while contact generation is retained (`CollisionGroup::held_melee`) — exactly the seam a physical design wants.
- Per-limb hitboxes are **unreachable** by melee contact (`CollisionGroup::hitbox()` is raycast-only and generates no contacts), so a creature presents only its actor capsule. VR melee cannot do locational damage at all (#956).

## Tests

- **882 lib tests pass**, including new coverage for the `Bool` kind (normalization, the overlay defaulting to off), the shape/size round-trip (`PhysicsShape::Cuboid`'s authored full size must not come back as Rapier's half-extents), and the trace's eviction invariant (a negative case: under plain FIFO the `Damage` is the first thing dropped).
- **`tools/shock2-sdk/test/vr-melee-swing.e2e.test.ts`** — a continuous swing (one sample per 60 Hz frame, no dwell) lands the authored 9 HP, with the already-touching case pinned beside it. It re-reads the live victim on every placement rather than staging against a torso sampled once (measured drift over the wind-up alone was 1.13 units, over half the contact volume), and asserts the resting contact is specifically **with the wrench**, since a creature touches the floor constantly. Its two known-bug assertions encode today's behavior deliberately, each with a `TODO(#1048)`/`TODO(#1049)` naming what to flip — so the suite stays green and the fix announces itself by failing.

## Validation

- `cargo fmt --all --check` ✅
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` ✅
- `cargo test -p shock2vr --lib` ✅ 882 passed
- SDK e2e: `dev-params`, `dev-menu`, `pause-menu`, `vr-melee-swing`, `vr-melee-contact`, `medsci-saved-vr-melee`, `weapon-*`, `vr-held-model`, `vr-weapon-handedness`, `missions` (23/23) ✅

Reviewed with `/xreview` (two independent engines plus a dedicated duplication pass); the cross-engine findings — a duplicate toggle registry, and the trace-flooding risk — are both fixed above.

Rebased onto current `main` (#1032, #1035, #1038, #1043, #1027, #1028).

Follow-ups filed: **#1053** (the real fix), #1048 (edge-only semantics), #1049 (the oversized `_w`-derived volume), #1051 (a proper collision budget for the trace).


tommy-xr:
not need anymore

PR #1047, opened 2026-08-19
feat(vr): arm-height dev param for on-device comfort tuning

tommy-xr:
Adds `arm_offset` ("Arm height (m)") to the dev-params registry (#1027) so the VR
arm height can be nudged live from the Developer screen (#1028) — no rebuild, no
redeploy. It moves **only the hands**, leaving the head where it is; that is the
deliberate complement to #1028's `eye_offset`, which moves the whole tracked stage
(head, hands and view together).

Range -0.5..+0.5 m, step 0.02, default **0.0** (bit-exactly inert until someone
opens the Developer screen).

## Where it is applied, and why that matters

At the top of `Game::update` — **not** `App::update`, which is where the first
draft put it.

#1027 rejected an eye-height knob wired into an accessor the Quest never reads:
it looked live in the debug runtime while being inert on device. This change hit
the *mirror image* of that trap, and the new e2e is what caught it: the desktop
and Quest runtimes drive a `shock2vr::App`, but **the debug runtime constructs a
`Game` directly**, so an offset applied at the `App` seam is live on device and
inert in the only harness that can verify it. `Game::update` is the seam all
three converge on, so one code path serves both.

It is also gated on `PresentationMode::Vr`: in flat the hands are synthesized
from the camera and never drawn, so a nonzero knob would silently move the
frob/aim ray origin with no visual feedback at all.

Every hand-pose consumer sits downstream of that one line — the interacting
hand, the rendered glove and sleeve, the forearm HUD panels, the frontend
pointer and teleport — so none of them needs its own wiring.

## Visuals

Sweeping `arm_offset` from -0.30 m to +0.48 m and back, live from the debug
runtime — medsci1 in `--vr`, hands held at a fixed tracked pose so the only
thing moving is the knob. Head, world and forearm-panel orientation stay put;
the hands, sleeves and forearm HUD rise and fall together.

arm height sweep

Before (0.0 m, left) / after (+0.3 m, right) — the same frame at the two ends of
the e2e's assertion:

before and after

## Verification

**Verified in the harness (debug runtime, `--vr`):**

- `tools/shock2-sdk/test/vr-arm-height.e2e.test.ts` — POSTs `arm_offset` and
  measures the objects the renderer was actually handed on the `player_hands`
  path: at 0.3 m they rise by 0.394 world units (`0.3 / METERS_PER_WORLD_UNIT`),
  and at -0.3 m they fall by the same. Negative-first: with the offset at the
  `App::update` seam the delta is 0.000 and the test fails.
- Four unit tests on the pure helper (`shock2vr/src/input_context.rs`): the
  metres→world-units conversion, the head not moving with the arms, only the
  vertical axis moving, and the default being bit-exactly inert.
- `cargo fmt --all --check`; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p
  desktop_runtime -p debug_runtime`; `cargo test -p shock2vr --lib` (850 passed);
  e2e: dev-params, dev-menu, all `vr-*`, pause-menu, and `missions` (23/23) green.

**NOT verified on device.** Nothing here has been worn. What a headset check
still owes us: that the knob's *feel* is right through the Quest's tracked pose
(the harness measures downstream of `stage_to_pawn`, never the tracked pose
itself), and that it composes with `eye_offset` as intended when both are
nonzero.

## Note on the symptom that prompted this

The knob was asked for as a comfort control, but the trigger was a report of
arms sitting much higher than the player's physical arms — which **went away on
restart**, so it is transient, not a persistent frame/pivot bug, and this PR
does not claim to fix it. The shape of "wrong until restart" points at a
one-shot sample: hand poses or the stage/reference-space mapping read once
before the OpenXR reference space is established and never re-sampled or
recentered. Nothing was hunted for it here and nothing obvious surfaced while
working in this code; filing it as a lead, not a diagnosis. If it recurs, this
knob at least makes it liveable in the moment.

## Not addressed (pre-existing, noted by review)

`virtual_hand.rs`'s `HAND_OFFSET` is a dead `vec3(0,0,0)` constant applied at a
strictly narrower seam (world space, interacting hand only — it never reaches
the rendered glove, forearm panels or teleport). It is now a second, inert,
less-correct expression of "translate both hands", and invites a future "just
bump `HAND_OFFSET`" fix in the wrong place. Left alone as out of scope.

### DIFF
--- a/shock2vr/src/dev_params.rs
+++ b/shock2vr/src/dev_params.rs
@@ -108,16 +108,9 @@ dev_params! {
     /// [`EYE_HEIGHT_OFFSET`]: that one moves the whole stage (head, hands and
     /// view together); this one moves *only* the hands.
     ///
-    /// Applied at the top of [`crate::Game::update`], the one place every
-    /// runtime's `InputContext` converges on, so it lands ahead of every
-    /// consumer - the interacting hand, the rendered glove and sleeve, the
-    /// forearm HUD panels, the frontend pointer and teleport - and lands
-    /// identically on the Quest and in the debug runtime. That placement is
-    /// the whole point, and it is `Game` rather than `App` on purpose: the
-    /// Quest and the desktop go through `App::update`, but the debug runtime
-    /// drives a `Game` directly, so applying it at the `App` seam would leave
-    /// the knob inert in the harness meant to verify it - the mirror image of
-    /// the accessor #1027 rejected for being inert on device.
+    /// Applied - VR only, and only there - at the top of
+    /// [`crate::Game::update`], which every runtime's `InputContext` converges
+    /// on; see that call site for why `Game` and not `App` is the seam.
     ARM_HEIGHT_OFFSET = float("arm_offset", "Arm height (m)", 0.0, -0.5, 0.5, 0.02),
 }
 
--- a/shock2vr/src/dev_params.rs
+++ b/shock2vr/src/dev_params.rs
@@ -102,6 +102,21 @@ dev_params! {
     /// deferral). Consequently the debug runtime cannot exercise this knob;
     /// it needs a worn check on device.
     EYE_HEIGHT_OFFSET = float("eye_offset", "Eye height (m)", 0.0, -0.5, 0.5, 0.02),
+    /// Vertical offset applied to BOTH tracked hands, in **meters**, relative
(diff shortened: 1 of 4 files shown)


tommy-xr:
Will hold off on this for now - the root issue was that the stage wasn't correctly set

PR #1056, opened 2026-08-19
fix(bench): accept negative coordinate arguments

### LINKED ISSUE #571: [tooling] cargo bn path show silently reports 'no path' for negative-leading coordinates
## Summary

`cargo bn path show` silently misparses coordinates that begin with a minus sign: clap reads e.g. `--to "-9.94,0,-30"` as a **flag**, not a value. The command then reports **no path** rather than erroring, producing a confident false negative.

## Severity: Low (tooling), but high-nuisance

This is an agent-verifiability trap. `cargo bn path show` is the recommended way to decide whether a location is reachable — a **silent** false "no path" from it can turn into a wrongly-filed engine blocker. It cost real time in an automated playtest session before being caught, and a nav-reachability conclusion drawn from it would have been wrong.

## Repro

```bash
# Silently reports no path (coordinate parsed as a flag):
cargo bn path show ops4.mis --from "53.5,-8.6,-27.6" --to "-9.94,-8.6,-30.0"

# Works correctly:
cargo bn path show ops4.mis --from="53.5,-8.6,-27.6" --to="-9.94,-8.6,-30.0"
```

Any coordinate whose first character is `-` triggers it. Positive-leading coordinates are unaffected, which is why it goes unnoticed — it only bites on one hemisphere of the map.

## Expected vs actual

- **Expected:** either the space-separated form parses correctly, or it fails loudly with a parse error.
- **Actual:** the argument is swallowed as a flag and the command reports "no path" — indistinguishable from a genuine unreachable result.

## Fix direction

In the `bench` CLI's clap setup for the `path` subcommands, either:
- mark the coordinate options with `allow_hyphen_values(true)`, or
- validate that the parsed value is a well-formed `x,y,z` triple and **error loudly** when it isn't.

Preferably both — the second guarantees a wrong invocation can never masquerade as a legitimate "no path" result again.

Also worth auditing any other `bn` subcommand taking coordinates (`path cell --at`, `path bench`) for the same trap.

Found by the `play-through` campaign (Ops deck, iteration 4).

tommy-xr:
## Summary

- accept space-separated negative-leading coordinates for every `cargo bn path` coordinate option
- parse and validate coordinate triples at clap's argument boundary
- cover `show --from/--to`, `cell --at`, `component --at`, equals syntax, and malformed triples

## Reproduction and result

On base `8977f23681174365daefa2ae1ec870140139d0ac`, the issue's space-separated command was rejected before pathfinding:

```text
$ cargo bn path show ops4.mis --from "53.5,-8.6,-27.6" --to "-9.94,-8.6,-30.0"
error: unexpected argument '-9' found
exit 2
```

Current main failed loudly rather than printing the originally reported silent `no path`, but still rejected the documented space-separated value. Its equals form succeeded with start cell 850, goal cell 305, and 45 waypoints.

With this change, both forms produce that same route:

```text
start cell: Some(850)  goal cell: Some(305)
45 waypoints, path length 88.28, straight-line 63.49, inflation 1.39, total turn 1150 deg
```

Malformed input now fails during CLI parsing, before any mission is loaded:

```text
error: invalid value '53.5,-8.6' for '--from <FROM>': expected 3 components, got 2
exit 2
```

## Root cause and fix

The coordinate options were unvalidated `String` arguments. Clap therefore interpreted a value beginning with `-` as another option, while coordinate parsing happened only after mission loading. The options now explicitly allow hyphen-leading values and use the existing `parse_vec3` function as their clap value parser, passing validated `Vector3<f32>` values to the diagnostics without changing pathfinding.

## Verification

- negative regression run on base: 2 failed, 1 passed (space-separated negative values and early malformed-value validation failed as expected)
- `RUSTFLAGS="-D warnings" cargo test -p bench` — 3 passed
- `RUSTFLAGS="-D warnings" cargo check -p bench` — passed
- `cargo fmt --all -- --check` — passed
- live `path show` with both space-separated and equals forms — identical cells and 45-waypoint result
- live `path cell --at "-9.94,-8.6,-30.0"` and `path component --at "-9.94,-8.6,-30.0"` — passed
- live malformed triple — clap value-validation error, exit 2

The full SDK e2e suite was not run because this is isolated to the standalone benchmark CLI's argv parsing and cannot affect a game runtime.

## Visual proof

`pr-visuals` gate: approved as genuinely non-renderable. This changes only shell-argument parsing and terminal diagnostics in the standalone `cargo bn` benchmark binary; it does not launch the game/debug runtime, alter mission simulation state, or enter a flatscreen/VR rendering path. A debug-runtime GIF or PNG cannot exercise whether clap accepts a particular argv token, so the matched before/after CLI transcript above and parser tests are the direct observable evidence.

Fixes #571

### DIFF
--- a/tools/bench/src/main.rs
+++ b/tools/bench/src/main.rs
@@ -10,7 +10,7 @@
 //!   cargo bn path stats medsci1.mis   # database stats + flag/okBits audit
 //!   cargo bn path bench medsci1.mis   # seeded A* benchmark
 //!   cargo bn path bench --all --json  # all missions, machine-readable
-//!   cargo bn path show medsci1.mis --from " -10,0,5" --to "20,0,30"
+//!   cargo bn path show medsci1.mis --from "-10,0,5" --to "20,0,30"
 //!   cargo bn path cell medsci1.mis --at "20,0,30"
 
 use std::sync::Arc;
@@ -76,26 +76,26 @@ enum PathCommand {
     Show {
         mission: String,
         /// Start position as "x,y,z" (world units, same as debug runtime)
-        #[arg(long)]
-        from: String,
+        #[arg(long, allow_hyphen_values = true, value_parser = parse_vec3)]
+        from: Vector3<f32>,
         /// Goal position as "x,y,z"
-        #[arg(long)]
-        to: String,
+        #[arg(long, allow_hyphen_values = true, value_parser = parse_vec3)]
+        to: Vector3<f32>,
     },
     /// List every cell containing a position, with flags and outgoing links
     Cell {
         mission: String,
         /// Position as "x,y,z"
-        #[arg(long)]
-        at: String,
+        #[arg(long, allow_hyphen_values = true, value_parser = parse_vec3)]
+        at: Vector3<f32>,
     },
     /// Show the walk component containing a position and its frontier links
     /// (how the component connects - or fails to connect - to neighbors)
     Component {
         mission: String,
         /// Position as "x,y,z"
-        #[arg(long)]
-        at: String,
+        #[arg(long, allow_hyphen_values = true, value_parser = parse_vec3)]
+        at: Vector3<f32>,
     },
 }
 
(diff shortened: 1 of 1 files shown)


tommy-xr:
Closing without merge after auditing the proposed change against issue #571's stated acceptance criterion.\n\nOn current origin/main at 8977f23681174365daefa2ae1ec870140139d0ac, the exact space-separated invocation exits 2 and prints: `error: unexpected argument '-9' found`. The equals form succeeds with start cell 850, goal cell 305, and 45 waypoints. Issue #571 defines success as either accepting the space-separated form OR failing loudly with a parse error, so current main already meets the required behavior and does not produce the reported silent false `no path`.\n\nThe coordinate clap definitions and clap 4.5.50 are unchanged from the issue-creation-time main SHA f819df9c159f6eef9e49d753255b25babd1b1ead; no responsible later parser fix is identifiable. The issue's `Preferably both` sentence is useful enhancement direction, but not an unmet acceptance requirement. This PR is therefore a convenience enhancement rather than a valid reproduction-driven fix for the sampled issue.\n\nThe approved pr-visuals assessment remains non-renderable: this behavior is solely standalone CLI argv parsing and terminal output, so matched command transcripts are the direct evidence; the game/debug renderer cannot exercise it.

PR #1091, opened 2026-08-22
feat(vr): prototype on-weapon ammo meter + glance-gated wrist health

tommy-xr:
**PROTOTYPE for design validation — not polish.** Explores the approved ambient-meter direction (ammo on the weapon, Alyx-style; health on the wrist but glance-gated, watch-style) to validate feel and architecture against the `virtual_arms` rendering path. Behind `--experimental ambient_meters`; without the flag (and everywhere in flat), behavior is unchanged.

## What it does (VR only)

- **On-weapon ammo meter** — the AMMOFULL readout (count / ammo icon / type, the exact `hud/ammo_panel.rs` canvas the right forearm wears today) is hung off the wielded weapon's live `RuntimePropTransform` (the same anchor muzzle flashes use), floating above the gun body and facing the shooter. The right-forearm ammo panel is retired while the flag is on.
- **Glance-gated wrist health** — the left-forearm BIOFULL panel renders only when the wrist panel's outward normal points at the eye (dot-product gate with hysteresis so it can't flicker at the boundary). Untracked zero-quaternion hand poses hide it (vr-ui-design rule 7).

Placement is decided once, in `shock2vr/src/hud/ambient_meters.rs` (§3 parity: the readout canvas is the shared `ammo_panel` layout; the meter module only supplies a root transform). A drive-by one-liner fixes `scenes/developer.rs`, which did not compile on main after #1082/#1084 crossed.

## Tuning knobs (all in `hud/ambient_meters.rs`)

| Knob | Value | Meaning |
| --- | --- | --- |
| `vr_config::WEAPON_METER_ANCHORS` | per weapon | weapon-local hang point (world units; +Y up out of the gun, +X back at the shooter) - see the update section below |
| `WEAPON_METER_TILT` | `-20°` | lean of the panel top away from the shooter |
| `WEAPON_METER_WIDTH` | `0.10` | panel width in world units (AMMOBACK 94:64 aspect preserved) |
| `GLANCE_SHOW_DOT` / `GLANCE_HIDE_DOT` | `0.55` / `0.30` | hysteresis thresholds on dot(panel normal, to-eye) |

## Demo

ambient meters demo

<sub>VR `debug_weapons`: two shots fired — the on-weapon meter counts 12 → 10 → 8 live — then the left wrist rolls toward the face and the health panel appears, and hides again on roll-away. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

| On-weapon meter + wrist glance (VR) | Flat: unchanged |
| --- | --- |
| vr still | flat unchanged |

## Verification

- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo test -p shock2vr` 902 passed (5 new unit tests: hysteresis, untracked-pose latch clear, zero-quaternion guard, meter transform offset + facing).
- VR runtime: meter appears when a gun is held, tracks the weapon, counts down on fire; wrist panel toggles at the glance threshold (verified via `/v1/scene` object counts: 7 ↔ 10) with no flicker.
- Flat (`cargo dbgr --mission debug_weapons`, no flag): HUD identical to main (screenshot above).

## Open questions for the owner

- **Readability at arm's length**: the wrist panel is small when glanced from a natural pose; may want a scale-up-on-glance or a larger authored panel.
- **Occlusion by hands/weapon**: the meter floats clear of the pistol, but long weapons (AR, shotgun) may intersect it — the offset likely wants to be per-model, like the vr_config grip table.
- **Glance threshold vs rest pose**: the dot-gate is position-dependent (it uses the true to-eye vector); at some rest poses the panel is already "glanced". Should the gate also require head gaze toward the wrist?
- **Reserve ammo**: the shared readout shows loaded clip + type only (as today's HUD does); if the meter should show loaded/reserve, the reserve count needs a new world query.
- Fade in/out on the glance transition (nice-to-have, skipped in the prototype).

## Review notes

Non-Critical review findings are tracked here rather than fixed (prototype): REVIEW_NOTES

https://claude.ai/code/session_01VNfTgRktqGXQ3C4xn6B7fK

## Update - condensed readout, anchored per weapon (owner feedback)

> *"we should we use the more 'condensed' UI (ie, the one that shows in flat when not in interface mode) - with just the single square for ammo (+ ammo type). I also wonder if we could bring it behind the weapon (we may need to customize attachment points for each weapon?)"*

- **Condensed readout.** The meter now draws the **AMMOBACK crop** of the AMMOFULL panel (94x64 - the single gauge square the flat HUD shows outside use mode): round count + ammo type, no icon, no wide backdrop. The crop's elements keep their AMMOFULL-authored rects, re-origined by one offset, so the compact and full readouts cannot drift apart (AGENTS.md §3).
- **Per-weapon attachment points.** Where the tag hangs is now a per-model anchor in `vr_config::weapon_meter_anchor_*`, beside the grip-offset table it mirrors. Anchors are derived from each 25AE view model's **authored bounding box** - back along the barrel by ~30% of its half-length, and clear of the top of its body - so the tag sits over the receiver like a rear-sight status plate instead of inside it. Unknown models (a classic install's world model) take a sensible default.
- **Scale fix.** `RuntimePropTransform` bakes `PropScale` in, so the meter root now strips scale - a scaled weapon no longer scales or displaces the tag.

compact on-weapon ammo tag

<sub>VR `debug_weapons`, AR15 wielded: the compact tag rides above the receiver and counts 12 → 11 → 10 → 9 as rounds go out. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

### Before → after

before/after

### Still

compact tag still

### Verification of the update

- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo fmt`; `cargo test -p shock2vr` **905 passed** (new: compact-anchor fallback, anchor keys pin to known models, scaled weapon does not scale the tag).
- VR runtime (`cargo dbgr --mission debug_weapons --vr --experimental ambient_meters`): cycled and wielded pistol / AR / shotgun / EMP / grenade launcher / stasis / fusion / worm launcher / viral proliferator, screenshotting each - the tag reads legibly over the body with no model clipping.

## Update — two owner-flagged leftovers fixed

**1. Ammo-type label overflow** — the ammo-type label shares its 94px `TYPE_LABEL` rect with the flat AMMOFULL/AMMOBACK readout, and an unbounded-length string could run past it (out of the panel entirely at extreme lengths). Fixed in 644a9aa5 by switching the ammo-type label to `canvas.text_native_fit(...)` (the existing shared ellipsize-at-layout facility) in both `emit` and `emit_compact` in `shock2vr/src/hud/ammo_panel.rs`.

Every real ammo-type string in the current weapon roster (`std`/`he`/`ap`/`pellet`/`rslug`/`lasershot`/`empshot`/`fusionshot`/`stasisshot`/`annrocket`/`viralshot`/the grenade-type family) measures comfortably under the 94px budget with `mainfont.fon` — none currently overflow, so there's no live repro with real ammo names. To make the fix visible, the before/after below temporarily forces a synthetic 25-character ammo-type string (`superlonghypotheticalammo`) through the same `emit`/`emit_compact` path, clearly overflowing the panel with the old `text_native` and cleanly ellipsizing (`SUPERLONGHYP...`) with the fix. This is a robustness/defensive fix for any string that does exceed the budget (a longer future ammo type, a mod, or localization) rather than a currently-visible defect with shipped content — confirmed via direct instrumentation of the shared `place_text` layout function against every real ammo-type string.

label overflow before/after

<sub>Flat `debug_weapons` HUD, ammo-type label forced to a synthetic 25-char string via a temporary capture-only override (reverted before landing). Before: `text_native` runs the label straight off the panel and off-screen. After: `text_native_fit` ellipsizes it to `SUPERLONGHYP...`, fully inside the 94px rect. Captured headlessly via the debug runtime.</sub>

**2. atek_h (pistol) tag anchor** — the arm-mesh-inflated bounding box floated the ammo tag about a panel-height above the pistol. Fixed earlier in e65146b6 by tightening `WEAPON_METER_ANCHORS["atek_h"]` from `(0.24, 0.27, 0.0)` to `(0.10, 0.15, 0.0)`; captured here (same framing, only the anchor constant changed) to refresh the PR's media.

atek_h anchor before/after

<sub>VR `debug_weapons`, pistol grabbed and viewed from a fixed free-camera framing (same weapon pose, only `WEAPON_METER_ANCHORS["atek_h"]` changed). Before: the tag floats well clear of the gun, disconnected from it. After: the tag hugs the rear of the slide like a rear-sight status plate. Captured headlessly via the debug runtime.</sub>

**9-weapon verification (VR, `--experimental ambient_meters`)**: grabbed all 9 debug-scene weapons in turn via the runtime's grab raycast (pistol, laser pistol, assault rifle, shotgun, EMP rifle, grenade launcher, stasis field generator, fusion cannon, worm launcher) and confirmed each swaps to its expected `_h` first-person model (`atek_h`, `lasehand`, `ar15_h`, `sg_h`, `empgun_h`, `gren_h`, `sfg_h`, `fsn_h`, `al_h`) with the compact ammo tag rendering legibly beside it — no overflow, no floating-tag regression on any weapon. One pre-existing, unrelated cosmetic artifact noted in passing: the pistol's rear iron sight renders as a flat solid-blue polygon (present in flat mode too, unaffected by this PR — likely a missing/placeholder texture on that one sight mesh).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -630,6 +630,16 @@ The project supports experimental flags for gating in-progress features during d
   default is the only reachable setting without a rebuild. Tell-tale log line:
   `high-detail (PMNM) meshes: true|false`.
 
+- **`ambient_meters`**: prototype VR ambient-meter model (flat HUD untouched).
+  The ammo readout moves onto the wielded weapon - a small AMMOFULL panel
+  anchored to the weapon's live transform, facing the shooter - and the
+  right-forearm ammo panel is retired while the flag is on. The left-forearm
+  health panel stays but renders only when the wrist is glanced at (panel
+  normal vs eye vector, with hysteresis; untracked poses hide it). Layout and
+  tuning knobs (weapon-local offset/tilt/size, glance thresholds) live in
+  `shock2vr/src/hud/ambient_meters.rs`. Without the flag, both forearm panels
+  render exactly as before.
+
 #### Adding New Experimental Features
 
 1. **Gate the feature in code**:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -631,14 +631,17 @@ The project supports experimental flags for gating in-progress features during d
   `high-detail (PMNM) meshes: true|false`.
 
 - **`ambient_meters`**: prototype VR ambient-meter model (flat HUD untouched).
-  The ammo readout moves onto the wielded weapon - a small AMMOFULL panel
-  anchored to the weapon's live transform, facing the shooter - and the
-  right-forearm ammo panel is retired while the flag is on. The left-forearm
-  health panel stays but renders only when the wrist is glanced at (panel
-  normal vs eye vector, with hysteresis; untracked poses hide it). Layout and
-  tuning knobs (weapon-local offset/tilt/size, glance thresholds) live in
-  `shock2vr/src/hud/ambient_meters.rs`. Without the flag, both forearm panels
-  render exactly as before.
+  The ammo readout moves onto the wielded weapon - the compact AMMOBACK square
+  (round count + ammo type, the same gauge the flat HUD shows outside use
+  mode), anchored to the weapon's live transform and facing the shooter like a
+  rear-sight status tag - and the right-forearm ammo panel is retired while the
+  flag is on. The left-forearm health panel stays but renders only when the
+  wrist is glanced at (panel normal vs eye vector, with hysteresis; untracked
+  poses hide it). Panel shape/tilt/size and the glance thresholds live in
+  `shock2vr/src/hud/ammo_panel.rs` + `shock2vr/src/hud/ambient_meters.rs`; the
+  per-weapon anchor (where the tag hangs in the weapon's own frame) lives in
+  `vr_config::weapon_meter_anchor_*`, beside the grip-offset table. Without the
+  flag, both forearm panels render exactly as before.
 
 #### Adding New Experimental Features
 
--- a/shock2vr/src/hud/ambient_meters.rs
+++ b/shock2vr/src/hud/ambient_meters.rs
@@ -16,6 +16,7 @@ use crate::vr_config::Handedness;
 
 use super::ammo_panel;
 use super::virtual_arms;
+use virtual_arms::OVERLAY_Z_OFFSET;
 
 /// The experimental-features name gating all of this.
 pub(crate) const FEATURE: &str = "ambient_meters";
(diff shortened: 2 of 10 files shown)


tommy-xr:
I like the direction, but we should use the more 'condensed' UI (ie, the one that shows in flat when not in interface mode) - with just the single square for ammo (+ ammo type) I also wonder if we could bring it behind the weapon (we may need to customize attachment points for each weapon?)

tommy-xr:
Done — both parts of the feedback are in `0209e563`.

**Condensed UI.** The meter now draws the **AMMOBACK crop** of AMMOFULL (94×64 — the single gauge square flat shows outside use mode): round count + ammo type, no icon, no wide backdrop. It is the same `ammo_panel` layout, cropped rather than re-authored — the elements keep their AMMOFULL rects, re-origined by one offset — so the compact and full readouts can't drift apart (§3).

**Behind the weapon, per weapon.** You were right that it wants per-weapon attachment points. The anchor now lives in `vr_config::weapon_meter_anchor_from_entity`, right beside the grip-offset table it mirrors, and is keyed on the model the VR hand actually wields. Rather than eyeballing each one, I pulled every 25AE view model's **authored bounding box** out of its `.bin` and derived the anchor from it — back along the barrel by ~30% of the model's half-length, and clear of the top of its body — so the tag sits on the receiver like a rear-sight status plate instead of inside the stock (which is exactly what the old single global offset did on the AR and shotgun). Unknown models fall back to a sensible default.

Verified in `debug_weapons --vr` across pistol / AR / shotgun / EMP / grenade launcher / stasis / fusion / worm launcher / viral proliferator — legible over the body, no clipping. Fresh GIF, before/after and still are in the PR body.

Two things I left alone and would take a steer on:
- **Ammo-type overflow.** Long type names ("LASERSHOT", "ANNELID") are wider than the 94px label rect, so they spill past the compact panel's edges. That rect is shared with the flat HUD, so fixing it here means changing the shared layout (ellipsize, or a narrower font) rather than a VR-only tweak — say the word and I'll do it in the shared canvas.
- **Height above the body.** The tag currently floats about one panel-height above the receiver, because the `_h` view models' bounding boxes include the arm mesh. Tightening it is a one-line-per-weapon nudge if you'd rather it hugged the gun.

tommy-xr:
Rebased onto current main (`370585fd`, well past the cyber-interface work and #1144/#1145) - one conflict, in `shock2vr/src/util.rs`'s `tracked_gaze()`: the incoming commit added an `is_tracked_rotation` early-return guard that main's already-current `tracked_rotation(head_rotation)?` call (right after) makes redundant, so the guard was dropped rather than duplicated. `is_tracked_rotation` itself stays live - `hud/ambient_meters.rs` still calls it directly.

Both flagged leftovers are now addressed:

- **Ammo-type label overflow.** `shock2vr/src/hud/ammo_panel.rs`'s `emit`/`emit_compact` now call `canvas.text_native_fit(...)` instead of `text_native(...)` for the ammo-type label - the existing shared ellipsize-at-layout-time facility (already used for save-name labels), so the fix lands once in the shared panel layout and both flat and VR inherit it (AGENTS.md §3). New negative-first unit test `a_long_ammo_type_name_fits_the_label_rect` confirms it (fails pre-fix, passes post-fix). Note: no ammo type currently shipping is actually long enough to overflow the 94px rect at today's roster (`stasisshot`/`fusionshot` top out around 68-69px), so this is a defensive fix verified with a synthetic long string, not a live repro - see the PR body's before/after for the synthetic case.
- **atek_h anchor** was already fixed on the branch (commit hugging the pistol) before this pass; re-verified with fresh before/after.

Verification: `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo test -p shock2vr --lib` 1135/1135 (added 1); SDK `npm test` 32/32; `missions.e2e.test.ts` 23/23; the ammo/weapon/reload e2e tests touched by this area are 13/14 (the one failure, `bio-ammo-full`'s "empty multi-ammo weapon" case, is a pre-existing baseline failure unrelated to this change). All 9 weapons re-verified in `debug_weapons --vr --experimental ambient_meters`: tag fits with no overflow, atek_h hugs the pistol like the others.

Pushed as `644a9aa5`; PR body updated with refreshed before/after media for both fixes.

Claude-Session: https://claude.ai/code/session_01CcZQxbiYUsaLJrJJaRAM72

tommy-xr:
superseded by other prs

PR #1097, opened 2026-08-22
fix(vr): collect keycards from every grab path

### LINKED ISSUE #583: Access-card pickup paths do not preserve both key access and the inventory item

tommy-xr:
## Summary

- route every `PropKeySrc` acquisition gesture through the derived `internal_keycard` Frob semantics
- cover corpse/container Take, VR panel squeeze, direct VR world squeeze, and held-card trigger recovery while leaving ordinary MOVE loot grabbable
- add default-VR Hydro2 regressions for authentic corpse 754 / Card B 942 and backpack acquisition, then verify reader 1120 opens HydroSectorB door 1127 without `hackfail`

## Play-through configuration

- scenario: `hydroponics`
- detail: `detailed`
- presentation: `VR`
- assets: `25th Anniversary`
- seed: `392917474`

## Verification

- `cargo test -p shock2vr` (904 passed)
- `cd tools/shock2-sdk && npm test` (310 tests; 26 passed, 284 e2e-gated)
- focused production VR E2E: 2 passed (`hydro2-vr-keycard.e2e.test.ts`)

## Visual verification

VR keycard acquisition demo

<sub>Same deterministic default-VR Hydro2 sequence on base and fix: open corpse 754's real world panel and squeeze Card B 942. Before, the physical card enters the hand and the reader answers `hackfail`; after, the credential is collected and door 1127 moves from y=0.1 to y=3.7. Captured headlessly through `/v1/step` + `/v1/screenshot`.</sub>

### Before → after

Before and after Card B acquisition

Fixes #583

### DIFF
--- a/shock2vr/src/scripts/gui/container.rs
+++ b/shock2vr/src/scripts/gui/container.rs
@@ -302,21 +302,11 @@ impl Gui<ContainerGuiState, ContainerGuiMsg> for ContainerGui {
         match msg {
             ContainerGuiMsg::GrabbedWithLeftHand(ent) => (
                 state.clone(),
-                Effect::GrabEntity {
-                    entity_id: *ent,
-                    hand: crate::vr_config::Handedness::Left,
-                    // TODO: Set this up to break link
-                    current_parent_id: None,
-                },
+                panel_grab_effect(world, *ent, crate::vr_config::Handedness::Left),
             ),
             ContainerGuiMsg::GrabbedWithRightHand(ent) => (
                 state.clone(),
-                Effect::GrabEntity {
-                    entity_id: *ent,
-                    hand: crate::vr_config::Handedness::Right,
-                    // TODO: Set this up to break link
-                    current_parent_id: None,
-                },
+                panel_grab_effect(world, *ent, crate::vr_config::Handedness::Right),
             ),
             // Wield-or-use (backpack click): a carried weapon - gun
             // (PropPlayerGun) or melee (PropLimbModel) - is wielded through
@@ -380,6 +370,21 @@ impl Gui<ContainerGuiState, ContainerGuiMsg> for ContainerGui {
                         (state.clone(), Effect::NoEffect)
                     };
                 }
+                // PropKeySrc gets an `internal_keycard` script even when its
+                // retail frob metadata is just MOVE. Scripted loot must run
+                // that Frob before any generic transfer, or the physical card
+                // reaches the backpack without its unlock credential (#583).
+                if crate::virtual_hand::uses_scripted_world_frob(world, *ent) {
+                    return (
+                        state.clone(),
+                        Effect::Send {
+                            msg: Message {
+                                payload: MessagePayload::Frob,
+                                to: *ent,
+                            },
+                        },
+                    );
+                }
                 let inventory_entity = world
                     .borrow::<shipyard::UniqueView<crate::mission::PlayerInfo>>()
                     .map(|player| player.inventory_entity_id)
@@ -399,13 +404,41 @@ impl Gui<ContainerGuiState, ContainerGuiMsg> for ContainerGui {
     }
 }
 
+/// A squeeze over a panel normally retrieves the icon into that hand. A
+/// keycard is the one semantic exception: it represents a collected access
+/// credential, so every panel (corpse loot and the backpack alike) must route
+/// the gesture through its derived keycard Frob instead of physically holding
+/// an unregistered object.
+fn panel_grab_effect(
+    world: &World,
+    entity_id: EntityId,
+    hand: crate::vr_config::Handedness,
+) -> Effect {
+    if crate::virtual_hand::is_key_source(world, entity_id) {
+        Effect::Send {
+            msg: Message {
+                payload: MessagePayload::Frob,
+                to: entity_id,
+            },
+        }
+    } else {
+        Effect::GrabEntity {
+            entity_id,
+            hand,
+            current_parent_id: None,
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::gui::GuiInputInfo;
     use crate::mission::PlayerInfo;
     use cgmath::{Quaternion, point2, vec3};
-    use dark::properties::{FrobFlag, Links, PropFrobInfo, PropHitPoints, ToLink, WrappedEntityId};
+    use dark::properties::{
+        FrobFlag, KeyCard, Links, PropFrobInfo, PropHitPoints, PropKeySrc, ToLink, WrappedEntityId,
+    };
 
     /// A world with a loot container holding one iconed, grabbable item,
     /// plus the player-info unique the Take path resolves the backpack
@@ -438,6 +471,17 @@ mod tests {
             inventory_entity_id: inventory,
         });
         (world, container, item, inventory)
+    }
+
+    fn mark_as_keycard(world: &mut World, item: EntityId) {
+        world.add_component(
+            item,
+            PropKeySrc(KeyCard {
+                is_master: false,
+                region_id: 128,
+                lock_id: 0,
+            }),
+        );
     }
 
     fn input_at(cursor: cgmath::Point2<f32>, pressed: bool) -> GuiInputInfo {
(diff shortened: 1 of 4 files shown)


tommy-xr:
Adversarial review (session): **merge with fixes**. The four-path routing is right and the effect ordering is verified safe (`internal_keycard` runs before `internal_frob_move`; the stale `DropEntityInfo` is absorbed by the aliveness guard; Hydro2 card 942 authors `PropHasRefs(false)` so the corpse's Contains link is cleaned; SwitchLinks/sibling scripts now fire on acquisition — more faithful than the old silent reparent). Findings to address before merge:

**H1** — `hydro2-vr-keycard.e2e.test.ts:171` triggers `MoveInventory`, which the merged cyber-interface stack (#1107) removed; the second test errors on first run against main, and its backpack-world-quad assumptions (`active_panel`, panel size/scale, "exactly one ui collider") no longer describe reality. Rewrite the leg against `ToggleUseMode` + `ui.strip` (see `inventory-strength.e2e.test.ts:150-160`). Until then the "VR backpack squeeze" claim is unverified on merged code.

**M1** — Take gates on `uses_scripted_world_frob` while the new `panel_grab_effect` gates on `is_key_source` only: a `MOVE|SCRIPT` item (eng1 circuit board / `FrobQB`) is Frobbed on click but grabbed on squeeze, and Take-on-click behavior for such items silently changed from backpack transfer to script-Frob. Use one predicate on both arms — `is_key_source` is the minimal fix for #583.

**M3** — the new world-squeeze Frob branch lacks the trigger path's `last_frobbed_entity` debounce; trigger+squeeze crossing 0.5 the same frame queues two Frobs, both drained before effects apply → duplicate `AcquireKeyCard` pushed into the save (`add_key_card` is a bare Vec::push). Set/gate `last_frobbed_entity` in the squeeze branch.

**M2 (defer to the always-collect generalization)** — the flat strip's cursor-drag lift is a fifth acquisition path (press lifts the item, never reaches `ContainerGui::handle_msg`; `FlatUiDragAction` has no use/Frob), so a keycard already in an old save's backpack can't be collected on flat. Fine to leave for the planned `is_always_collected` refactor, but noting so "every grab path" isn't over-claimed.

Also L1: the direct world-squeeze leg has no test; L2: the "same Frob as the trigger path" comment is inaccurate re: debounce (see M3).

tommy-xr:
Rebased onto main (`f220075a`, past #1119 "harden nanite collection") and addressed the review findings.

**Rebase note**: this PR's own `virtual_hand.rs` world-squeeze hunk (the `is_key_source` branch added directly in `handle_empty_hand_state`) was **dropped as subsumed**. Main's #1119 already restructured that squeeze branch around `uses_scripted_world_frob` with a per-entity `last_frobbed_entity` debounce, and `uses_scripted_world_frob` already covers `PropKeySrc` (keycards) alongside authored `SCRIPT` and nanite piles - so main now routes direct world-squeeze on a keycard through Frob with proper debounce on its own. Kept: the new `is_key_source` helper, `panel_grab_effect` (VR panel/strip squeeze), the Take-arm scripted-frob check, and the trigger-recovery path in `held_trigger_press_payload`.

**H1** - `hydro2-vr-keycard.e2e.test.ts` used the removed `MoveInventory` action and asserted the deleted backpack world-quad (`active_panel` size/scale, "exactly one ui collider"). Rewrote the second scenario against `ToggleUseMode` + `ui.strip`/`panel_pose` (the cyber-interface pointer-bridge idiom from `vr-cyber-interface-pointer.e2e.test.ts`). Since the fix now intercepts every squeeze path, a keycard can no longer come to rest in the backpack through gameplay, so the new scenario stages a card directly into the backpack via `spawnItem` (an old-save / deferred-M2-flat-drag proxy) and verifies squeezing it out of the strip still Frobs it - collecting the credential and leaving no card in the world or the strip - instead of lifting it into the hand. The corpse-loot-panel leg is unchanged (still the real VR world-panel collider, untouched by the cyber-interface rework). The separate held-card trigger-recovery case is left to its existing unit-level coverage (`virtual_hand::tests::held_keycard_trigger_collects_it_through_frob`) - there's no SDK path to place an item directly in a hand, and under the fix that state is no longer reachable through any live pickup path, only through a pre-fix save.

**M1** - Take's new scripted-frob check gated on the broad `uses_scripted_world_frob` (authored `SCRIPT` *or* keycard *or* nanite), diverging from `panel_grab_effect`'s narrower `is_key_source` and silently changing Take-on-click for `MOVE|SCRIPT` loot (e.g. eng1's circuit board, `FrobQB`) from backpack transfer to script-Frob. Narrowed the Take-arm check to `is_key_source`, matching `panel_grab_effect` exactly. `tools/shock2-sdk/test/world-frob-pickup.e2e.test.ts` ("eng1: frobbing MOVE items picks them up and preserves the scripted circuit board flow") already covers this scenario end-to-end and now passes (previously it would have hit the wrong new branch).

**M3** - `KeyCardScript::handle_message` unconditionally issued `AcquireKeyCard` + `DestroyEntity` on every `Frob`, and `QuestInfo::add_key_card` is a bare `Vec::push` (not idempotent), so two Frobs landing in the same message batch (e.g. both hands, or trigger+squeeze) would push a duplicate credential before the first `DestroyEntity` applies. Added an in-memory `collected` latch mirroring `InternalNanitesScript`, with a negative-first unit test (`a_second_frob_in_the_same_batch_acquires_nothing_more`) confirming the second Frob in a batch is now a no-op. Separately, the rebase's adoption of main's #1119 debounce structure already covers the direct-world-squeeze duplicate-Frob path with `last_frobbed_entity`.

**L2** - the inaccurate "same Frob as the trigger path" comment lived in the world-squeeze hunk that was dropped as subsumed during the rebase, so it's gone.

**M2** (flat cursor-drag lift, deferred to the `is_always_collected` generalization) - untouched, as instructed.

## Verification
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` - clean
- `cargo fmt --all -- --check` - clean
- `cargo test -p shock2vr --lib` - 946 passed (0 failed)
- `tools/shock2-sdk`: `npm install`, `npm run build` (tsc clean), `npm test` - 26 passed / 288 e2e-gated-skipped
- Foreground `SHOCK2_E2E=1` runs (each launched its own debug-runtime instance):
  - `hydro2-vr-keycard.e2e.test.ts` - 2/2 passed
  - `container-loot.e2e.test.ts` - 1/1 passed
  - `vr-cyber-interface.e2e.test.ts` - 2/2 passed
  - `inventory-strength.e2e.test.ts` - 3/3 passed
  - `world-frob-pickup.e2e.test.ts` (M1 regression coverage) - 1/1 passed
  - `missions.e2e.test.ts` - all 23 missions load

PR #1102, opened 2026-08-22
fix(vr): draw world loot panels over their host geometry

### LINKED ISSUE #1100: Hydro3 Desk #2 container rows are occluded by the desk collider in VR
## Campaign context

- Campaign: Hydroponics, detailed play-through, production VR, 25th Anniversary assets, seed `392917474`
- Mission: `hydro3.mis`
- Stable mission objects: `Desk #2` **2116** (template `-120`, model `desk_sd`) contains `Audio Log` **200** (template `-76`, deck 3 / log 16), plus Psi Booster 116 and Large Beaker 99
- This is an **optional, nonblocking** log/loot defect. Hydroponics progression can continue without these items.
- Reviewed campaign evidence: `/tmp/claude/playthrough/runs/hydroponics-iter0-replay1/action-transcript.jsonl`, especially labels `inspect-hydro3-desk354-panel-log200`, `verify-*-hydro3-log200-row`, `probe-hydro3-desk-ui-only-*`, and `scan-log200-upper-row-*`.

## Reproduction

1. Launch `hydro3.mis` in production VR and approach mission object 2116 at `(31.0, 1.098993, -1.8)` from the north-office foyer.
2. With an empty hand, aim at the desk and pulse that hand's trigger. The container opens normally.
3. Aim the same production VR controller ray at the Audio Log bottom row (`rect [15,221,35,34]`) and pulse trigger.
4. Repeat from the west, by reaching over the desk, and from the east/far-east side.

The panel reports all three rows, but none can be selected; the Audio Log row remains and deck 3/log 16 is not collected.

## Measured geometry

- Audio Log row world-space vertical span: **y `1.670993..1.806993`**.
- `desk_sd` host collider top: **y `1.809403`**. The whole row is therefore below the top; even its upper edge is buried by about `0.00241`.
- UI proxy plane: **x `30.995`**.
- Desk collider spans **x `30.39751..31.60249`**, so the proxy lies inside the desk from either side.
- UI-only probes at y `1.739` hit the proxy at x `30.995` for z `-2.05`, `-1.8`, and `-1.55`, confirming the panel exists.
- Production/all-group probes hit the desk instead:
  - west cardinal ray hits x `30.39751`;
  - far-east ray hits x `31.60249`;
  - starts placed over/inside the row report distance-zero hits on the desk.
- Upper-row scans at y `1.79`, `1.8`, `1.805`, and `1.807` all still hit the desk at x `30.39751`.
- Two production empty-hand trigger attempts left the row present and the player's collected-log list unchanged.

## Expected

Opening a diegetic container positions its shared panel/proxy outside the host's selectable collider, so every rendered row is reachable by the same VR ray that supplies hover and click. Flat presentation should retain the identical 188×296 canvas layout.

## Actual

The panel is rendered through/inside the desk volume. The selectable desk collider wins the shared VR ray before it can reach any buried row, making the items inaccessible with no physical approach workaround.

## Distinct from #835

This is **not #835**. That audit concerns the zero local offset used by one model-bounds collider branch for potentially asymmetric models. Mission object 2116 authors an oriented-box physics representation and its `desk_sd` host collider is correctly centered/symmetric around the desk. The failure is the panel/proxy placement relative to that valid host collider, not a displaced selectable collider.

tommy-xr:
Fixes #1100

## Summary

Reworked per owner direction (see PR comments): the placement approach in
the original version of this PR is superseded by #1115, which already fixed
hit-testing/reachability by letting a panel bypass its own host collider.
The remaining problem was purely visual — a loot panel whose plane sits
inside its host mesh (Hydro3 Desk #2) rendered buried/z-fighting: clickable,
but not visible.

- **Reverted** the OBB front-face placement change in `ContainerGui` (and its
  unit test) — no longer needed now that #1115 covers reachability.
- **Draw world loot panels over everything**, the way other VR system panels
  already present: `GuiManager::render_filtered` (the single shared draw path
  for every `GuiScript` world panel — container/loot, keypads, computers,
  replicators, etc.) now tags its panel quads `RenderLayer::SceneUi`, the same
  "draw over the world" depth-bypass layer the per-eye HUD/viewmodel are
  bumped into. Each `RenderLayer` clears depth at its own group boundary
  (`engine/src/gl_engine.rs`), so a panel on this layer can never be occluded
  by earlier `World`-layer geometry (the host mesh), while still sitting
  *before* `SystemOverlay` — so the pause menu and the cyber interface still
  draw over an open panel. This fixes the whole class of "panel plane buried
  in its host" bugs (any host, not just Desk #2) with one shared-path change
  instead of a per-object placement heuristic.
- Flat presentation is untouched: the flat MFD-docked panel renders through
  `FlatUiHost`/`flat_ui_host.rs`, a completely separate screen-space path that
  never touches `GuiManager`'s world-space quads.
- Kept the PR's Hydro3 Desk #2 regression, but rewrote its pivotal assertion:
  dropped the raw-raycast nearest-hit check (it guarded raw physics geometry,
  which is no longer meaningful post-#1115 since the hand ray bypasses the
  host collider regardless of panel placement). It now asserts the
  player-observable outcomes instead: the buried Audio Log row is still
  clickable end-to-end (aim + trigger collects it), and — for the visual claim
  — the panel's rendered quads report `render_layer: "scene_ui"` (not
  `"world"`) via `GET /v1/scene`, with `depth_write: true` on every quad.

## Visual evidence

Hydro3 Desk #2 loot panel drawn over the desk

<sub>Hydro3 Desk #2: the Audio Log row now renders fully visible, drawn over the desk's own geometry instead of buried behind it. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

### Before → after

Before and after: the Audio Log disc is no longer clipped by the desk's top surface

<sub>Before: the disc icon is sliced by the desk's top-surface edge, its lower ~40% invisible. After: the same disc renders fully in front of the desk, unclipped.</sub>

## Testing

- `cargo test -p shock2vr --lib`
- `npm run build && npm test` in `tools/shock2-sdk`
- `SHOCK2_E2E=1 node --test --test-concurrency=1 dist/test/hydro3-vr-desk-loot.e2e.test.js` (rewritten)
- `SHOCK2_E2E=1 node --test --test-concurrency=1 dist/test/vr-container-world-panel.e2e.test.js`
- `SHOCK2_E2E=1 node --test --test-concurrency=1 dist/test/container-loot.e2e.test.js` (flat MFD path, unaffected)
- `SHOCK2_E2E=1 node --test --test-concurrency=1 dist/test/vr-cyber-interface.e2e.test.js` (SystemOverlay still draws over the panel's SceneUi layer)
- `SHOCK2_E2E=1 node --test --test-concurrency=1 dist/test/missions.e2e.test.js` (23/23)
- Full SDK e2e suite (`npm run test:e2e`)

### DIFF
--- a/shock2vr/src/gui/gui_manager.rs
+++ b/shock2vr/src/gui/gui_manager.rs
@@ -2,7 +2,10 @@ use std::collections::HashMap;
 
 use cgmath::{Deg, EuclideanSpace, InnerSpace, Matrix4, Vector2, Vector3, vec3};
 
-use engine::{assets::asset_cache::AssetCache, scene::SceneObject};
+use engine::{
+    assets::asset_cache::AssetCache,
+    scene::{RenderLayer, SceneObject},
+};
 use rapier3d::prelude::RigidBodyHandle;
 use shipyard::{Component, EntitiesView, EntityId, Get, UniqueView, View, World};
 
@@ -443,7 +446,20 @@ impl GuiManager {
             let canvas = UiCanvas::from_elements(size_px, elements);
             let z_step =
                 COMPONENT_Z_STEP.min(PANEL_DEPTH_BUDGET / canvas.element_count().max(1) as f32);
-            ret.extend(canvas.render_world_space(asset_cache, root_transform, None, None, z_step));
+            let mut panel_objects =
+                canvas.render_world_space(asset_cache, root_transform, None, None, z_step);
+            // Draw over everything the panel's own host mesh might otherwise
+            // occlude it with (issue #1100: a wide host like Hydro3 Desk #2
+            // encloses the panel plane, burying it behind the desk's own
+            // geometry). `SceneUi` is the same "draw over the world" layer the
+            // per-eye HUD/viewmodel are bumped into (see `lib.rs`); it clears
+            // depth at its own group boundary and is drawn before
+            // `SystemOverlay`, so the pause menu and cyber interface still
+            // draw over an open panel.
+            for object in &mut panel_objects {
+                object.set_render_layer(RenderLayer::SceneUi);
+            }
+            ret.extend(panel_objects);
 
             // gui_obj.set_local_transform(
             //     Matrix4::from_translation(info.offset)
--- a/shock2vr/src/scripts/gui/container.rs
+++ b/shock2vr/src/scripts/gui/container.rs
@@ -1,7 +1,6 @@
 use cgmath::{Vector2, Vector3, vec2};
 use dark::properties::{
-    FrobFlag, Link, PhysicsModelType, PropFrobInfo, PropInventoryDimensions, PropObjIcon,
-    PropPhysDimensions, PropPhysType, ReceptronEffect,
+    FrobFlag, Link, PropFrobInfo, PropInventoryDimensions, PropObjIcon, ReceptronEffect,
 };
 
 use shipyard::{EntityId, Get, View, World};
@@ -85,12 +84,6 @@ const BACKPACK_GRID_ORIGIN: Vector2<f32> = Vector2::new(4.0, 17.0);
 /// `res/iface/BLOCK.PCX` is authored to cover one cell's 34x32 interior,
 /// leaving the separators in `INVBACK.PCX` visible around it.
 const BLOCK_SIZE: Vector2<f32> = Vector2::new(34.0, 32.0);
-
-/// Keep the thin world-panel collider wholly outside an authored host OBB.
-/// Rapier clamps its zero authored depth to 0.01 world units; this leaves a
-/// further half-thickness of air so the host can never win the hand ray at
-/// coincident surfaces.
-const LOOT_PANEL_HOST_CLEARANCE: f32 = 0.01;
 
 impl ContainerGui {
     pub fn loot_container() -> ContainerGui {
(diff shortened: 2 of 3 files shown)


tommy-xr:
sueprseded by #1107

tommy-xr:
Rationalized against the merged cyber-interface stack + #1115 (session review):

- The **gameplay symptom is now subsumed by #1115**: the desk is the panel's own `gui_host`, so the hand ray bypasses it and the Audio Log row is clickable on plain main. The remaining real problem is **visual** — the panel plane still renders buried in the desk mesh (occlusion/z-fighting: clickable but not visible).
- Owner direction: rather than moving the panel in front of the authored OBB, **draw loot panels over everything** (depth-bypass overlay layer, the way the pause-menu/frontend world panels present). That covers the whole class (hosts with no usable OBB too) without per-host placement heuristics.
- The regression test's pivotal assertion (`raycast` nearest-hit `=== panel.entity_id`) guards raw physics geometry, not player-observable behavior — post-#1115 it would fail without this placement change even though the player is unblocked. It should be re-framed around visibility/screenshot or interaction evidence.

Suggest closing this placement approach and replacing it with the render-over change (can reuse this PR's e2e scenario minus the raw-raycast assertion).

tommy-xr:
Reworked per the discussion above.

**Reverted** the OBB front-face placement approach (`ContainerGui::get_config_for` + its unit test) - no longer needed now that #1115 fixed hit-testing by letting a panel bypass its own host collider.

**Implemented instead**: draw every `GuiScript` world panel over its host geometry, the way the pause menu and cyber interface already present. `GuiManager::render_filtered` (`shock2vr/src/gui/gui_manager.rs`) - the single shared draw path for every world panel (container/loot, keypads, computers, replicators, ...) - now tags its panel quads `RenderLayer::SceneUi`, the same depth-bypass "draw over the world" layer the per-eye HUD/viewmodel are bumped into. Each `RenderLayer` clears depth at its own group boundary (`engine/src/gl_engine.rs`), so:
- `SceneUi` panel quads can never be occluded by earlier `World`-layer geometry (the host mesh) - fixes the whole class of "panel plane buried in its host" bugs, not just Desk #2.
- `SystemOverlay` (pause menu, cyber interface) still comes after `SceneUi`, so they still draw over an open panel - verified with `vr-cyber-interface.e2e.test.ts`.
- Flat presentation is untouched - it renders through the separate `FlatUiHost` screen-space path, which never touches `GuiManager`'s world-space quads.

Kept the Hydro3 Desk #2 e2e scenario but rewrote its pivotal assertion: dropped the raw nearest-hit raycast against `collision_groups: ["all"]` (it guarded raw physics geometry, invalid post-#1115 since the hand ray bypasses the host collider regardless of panel placement). It now asserts player-observable outcomes: the log row is clickable end-to-end (aim + trigger collects deck-3/log-16), and the panel's rendered quads report `render_layer: "scene_ui"` with `depth_write: true` via `GET /v1/scene` - the render-state signature of "drawn over the world" rather than "buried in it."

Visual evidence (before/after PNG + GIF) is in the PR body above. Tests: `cargo test -p shock2vr --lib`, and the e2e suite - reworked `hydro3-vr-desk-loot`, `vr-container-world-panel`, `container-loot`, `vr-cyber-interface`, and `missions` (23/23) - all green.

PR #1142, opened 2026-08-24
spike(vr): hold a gun physically too, behind physical_held_items

tommy-xr:
Spike answering: *"could we have the same movement properties for held ranged weapons (i.e. just don't pass through solid stuff)?"*

Short answer: **yes, and the machinery generalises almost for free** — but two things had to be answered honestly first, and one of them (the collider) needed real work. It is behind `--experimental physical_held_items`, default OFF, VR only, so nothing shipped changes until you try it in a headset.

## 1. Does the melee drive generalise?

Yes. Nothing in `set_held_melee` / `drive_held_melee` reads a melee property — the drive is about *held physicality*: sweep the body from where it is toward the tracked hand each step, stop at the first WORLD surface, take the hand's rotation exactly. Commit 1 is a pure rename (`held_melee*` -> `held_item*`, no behavior change) so the wider gate reads honestly. The genuinely melee-specific pieces keep their names: `CollisionGroup::held_melee` and `is_vr_melee_weapon`.

## 2. Colliders — this was the real work

`posed_weapon_bounds` covered only **LGMM skinned** meshes (the melee `_h` rigs). The 25AE gun `_h` models are **LGMD object meshes**, so `weapon_geometry` returned `None` for every gun and a held gun would have kept its loose-prop box: the size of the **world** model and centred on the **grip**, i.e. wrong in both directions.

Extended it instead: an object mesh's sub-object hierarchy *is* a skeleton (the same one the renderer poses it with), and the weapon/arm split is by **material** (`ND-arm*`) rather than by terminal joint, because a gun's arm is not a joint — `split_at_terminal_joint` is a melee-rig heuristic and does not apply. New: `obj_weapon_geometry` + `ss2_bin_obj_loader::{to_vertices_by_material, build_skeleton}`.

Measured on the shipped pistol at wield time:

```
wield 'atek_h': rendered weapon 46x30x8 cm, center (-0.365, -0.005, 0.020)
```

(46 cm is a *pistol*: the view models are ~2x life size, the same finding the melee scale work landed on.)

## 3. Does contact damage leak?

No, and now it cannot. `TriggeredMeleeWeapon` is attached only to `PropLimbModel` entities (`entity_creator`), and guns author no `PropLimbModel` (`cargo dq templates 17` — Pistol has `PropPlayerGun` + `WeaponScript`, no limb model). But rather than rely on that, a held gun takes **`CollisionGroup::held_inert()`** — empty memberships, empty filter — so it generates no contacts at all.

That is not just belt-and-braces; it removes two hazards that would otherwise have been blockers:

- **The gun would be shot by its own bullets.** `internal_fast_projectile` starts its ray `0.25 * SCALE_FACTOR` = 0.625 units *behind* the muzzle — inside the barrel — and a solid raycast reports a shape containing the origin at distance 0. A gun with an ordinary `ENTITY` membership is in that ray's mask.
- **Anything spawned at the muzzle would be shoved by it** — a grenade, say.

The sweep that stops the weapon runs its **own** query and does not consult the body's groups, so an inert body is still blocked by the level while touching nothing. There is a test for exactly that claim.

## 4. What breaks?

The one that matters: **a blocked weapon renders where its body is** (the render transform comes from physics), and a VR shot leaves the **rendered** muzzle vhot (`create_projectile`). So a gun held back by a wall also fires from the held-back position. That is self-consistent — you shoot from where you can see the barrel — but it *is* a change to how firing into cover behaves, and it is the thing to judge in a headset.

Everything else checked out: `store` and `drop` remove the body through the same `make_un_physical` the melee path uses (the gate now covers guns at all five sites, including the save/load restore); the two-handed/backpack paths route through `GrabEntity`, which is gated; reload does not touch the body.

## 5. Recommendation

**Worth trying.** The win (no gun clipping through walls, the classic VR immersion break) is real, the drive was already built and proven, and the inert group makes the gameplay risk structural rather than a matter of care. The residual risk is feel — a gun that visibly detaches from your hand near geometry, and the muzzle displacement above — which is exactly what a flag is for.

## Verification

Measured headlessly on the debug runtime (`--mission medsci1.mis --vr`), same deterministic sequence both times: wield the pistol, face a wall 1.42 units ahead, push the hand 1.9 units forward (0.83 past the wall).

| hand (pawn z) | flag OFF | flag ON |
| --- | --- | --- |
| 0.55 — free | weapon at x = -34.409 | -34.409 (identical) |
| 1.9 — into the wall at x = -33.9 | **-33.072** (0.83 units *through* the wall) | **-33.976** (stopped at the face) |

- Held gun with the flag: `kinematic`, `collision_groups: []` — inert, as designed.
- Held gun without the flag: **no body at all** — shipped behavior preserved.
- Held wrench, flag either way: `kinematic`, `collision_groups: ["entity"]` — melee unchanged.
- Fired blocked and unblocked: both shots hit level geometry; **no `Damage` message to the pistol itself**.

Tests:
- `an_inert_held_item_is_still_stopped_by_world_geometry` — negative-first: with the sweep stubbed out it fails with `x=1` (straight through the wall).
- `an_inert_held_item_touches_nothing_it_sweeps_through` — the counterpart to the existing melee-contact test.
- Four gate tests (`held_item_physics_tests`): gun unphysical without the flag, inert with it, melee keeps its contact body either way, nothing physical in flat.
- `vertices_come_back_grouped_by_their_material` in the obj loader.

Also `SHOCK2_E2E=1 test/missions.e2e.test.ts` — **all 23 missions load** (this change adds a world unique at level load, which is exactly that test's blast radius). The full e2e suite was not run to completion: it parks on the `earth.mis` hacking scenario, which is one of the three tests that already fail on `main` itself, so its verdict would not have been about this change either way.

```
cargo fmt --all -- --check                                              clean
RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime   clean
RUSTFLAGS="-D warnings" cargo test -p shock2vr --lib                    993 passed
RUSTFLAGS="-D warnings" cargo test -p dark --lib                        118 passed
```

## No GIF — and why

The visible effect is *where the weapon is*, and the numbers below pin it exactly. I could not get a frame that shows it honestly: the debug runtime's VR eye sits about a metre behind the hand looking `+Z`, and every reachable WORLD wall in medsci1 is off to the side of that view — with the flag OFF the gun ends up **behind** the wall and is depth-culled out of frame entirely, so the "before" shot is a picture of a wall. A headset is the right instrument for this one, which is what the flag is for. What a still *can* show is that the VR `_h` wield the fit measures is real and rendering — this is the pistol pressed against that wall, flag on, and it is the same 46 cm view model the log line reports:

VR wielded pistol, held against a wall

## Review

`/xreview` (Claude + Codex + a de-dup pass) found **no correctness defect in the drive or the lifecycle** — both engines independently confirmed the melee path is unchanged and that every un-hold path (drop, store, destroy, level change) removes the body and its drive target. It did catch one real hole, now fixed in its own commit:

- **A classic install would have got a body with no fit.** The gate keyed on `PropPlayerGun` alone, but on a classic install VR keeps the *world* model, so no `_h` wield fires and the fit never runs — the gun would have swept its inherited loose-pickup box. The gate now asks the same question `InternalSwitchHeldModelScript` asks before swapping, so "no fit" means "no body", and the gate test asserts that both ways round.

Plus small simplicity fixes (redundant predicate, an unreachable `let-else` arm, last-wins material lookup, `is_inert` also checking solver groups, one duplicated test folded into a group-parameterized scenario) and two knowingly-inherited limits now documented rather than papered over:

- **Rotation is never swept**, only translation — so a long barrel can be *turned* into a wall it cannot be *pushed* into. That is the drive's existing, deliberate tradeoff (a rotationally-swept weapon whose angle is not the hand's angle was judged worse); a gun just makes it more visible.
- **`melee_max_speed` also clamps a gun's tracking**, at 60 units/s. It never binds for a real hand, so it is left shared rather than forked into a second dev param.

Claude-Session: https://claude.ai/code/session_017cm626D5M4tAz7yLPWPMFb

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -570,6 +570,20 @@ The project supports experimental flags for gating in-progress features during d
   default is the only reachable setting without a rebuild. Tell-tale log line:
   `high-detail (PMNM) meshes: true|false`.
 
+- **`physical_held_items`**: give a VR-held **gun** the same movement
+  properties a held melee weapon has had since #1121 - a swept kinematic body
+  that stops at world geometry instead of sliding through it. The collider is
+  fitted to the rendered `_h` view model (arm geometry excluded, by material),
+  and the held body is **inert**: no memberships, no filter, so it generates no
+  contacts, is invisible to the projectile raycast that starts inside its own
+  barrel, and cannot shove anything spawned at the muzzle. Only the level holds
+  it back. Melee weapons are unaffected either way (their body is their damage
+  volume and keeps `CollisionGroup::held_melee`), and the flat presentation is
+  untouched. Note a blocked weapon *renders* where its body is, so the shot -
+  which leaves the rendered muzzle vhot - leaves from the held-back position
+  too. VR only; without the flag a held gun has no body at all, exactly as
+  before. Tell-tale log line: `wield '<model>': rendered weapon WxHxD cm`.
+
 #### Adding New Experimental Features
 
 1. **Gate the feature in code**:
--- a/dark/src/importers/model_importer.rs
+++ b/dark/src/importers/model_importer.rs
@@ -261,13 +261,21 @@ fn obj_weapon_geometry(obj: &SystemShock2ObjectMesh) -> Option<VrHeldWeaponGeome
 }
 
 fn weapon_geometry(mesh: &SystemShockContentModel) -> Option<VrHeldWeaponGeometry> {
-    let SystemShockContentModel::Mesh(ai_mesh, skeleton, pmnm) = mesh else {
-        let SystemShockContentModel::Obj(obj) = mesh else {
-            return None;
-        };
-        return obj_weapon_geometry(obj);
-    };
+    match mesh {
+        SystemShockContentModel::Obj(obj) => obj_weapon_geometry(obj),
+        SystemShockContentModel::Mesh(ai_mesh, skeleton, pmnm) => {
+            skinned_weapon_geometry(ai_mesh, skeleton, pmnm)
+        }
+    }
+}
 
+/// The weapon-only geometry of an LGMM skinned first-person model - the melee
+/// `_h` rigs, with or without a 25AE high-detail `PMNM` chunk.
+fn skinned_weapon_geometry(
+    ai_mesh: &SystemShock2AIMesh,
+    skeleton: &Rc<Skeleton>,
+    pmnm: &Option<crate::ss2_bin_pmnm::PmnmMesh>,
+) -> Option<VrHeldWeaponGeometry> {
     if let Some(pmnm) = pmnm {
         let runs = pmnm.to_skinned_vertices();
         let has_named_arm = runs
--- a/dark/src/importers/model_importer.rs
+++ b/dark/src/importers/model_importer.rs
@@ -225,9 +225,40 @@ fn split_at_terminal_joint(
         .partition(|vertex| rigid_joint(vertex) == Some(terminal_joint))
 }
 
+/// The weapon-only geometry of an LGMD first-person model (the 25AE gun `_h`
+/// meshes, and any classic world model wielded as-is).
+///
+/// These carry no skeleton of their own: the "joints" are the sub-object
+/// hierarchy, which is how a slide or a magazine moves, and the renderer poses
+/// them through exactly the skeleton [`ss2_bin_obj_loader::build_skeleton`]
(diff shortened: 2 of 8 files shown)


tommy-xr:
Replaced with #1472

tommy-xr:
Superseded: physically held guns landed via #1472 (constrain held guns against level geometry).

PR #1153, opened 2026-08-24
feat(vr): let a held gun be heard hitting the level

tommy-xr:
Stacked on **#1142** — review that first.

## The gap

#1139 gave a held **melee** weapon an impact sound on every contact, so a
wrench on a wall thuds. It works because melee is held with a live collision
group, so Rapier's narrow phase reports the touch and `TriggeredMeleeWeapon`
turns it into a sound.

A held **gun** under #1142 gets none of that, on purpose: it is `held_inert`
(no memberships, no filter) so the projectile ray cannot hit the barrel it
starts inside, and nothing spawned at the muzzle can be shoved by it. No
contacts, no sound. Push a gun into a wall and it stops — silently.

## Why not just give it a collision group

Both hazards above are about the **raycast** and the **solver**, not about
contact generation. Re-enabling contacts to get a sound would take on real
gameplay risk (a gun shot by its own bullets) to solve an audio problem.

**The sweep already knows what stopped the gun.** `held_item_sweep_fraction`
called `cast_shape` and matched `if let Some((_, hit))` — discarding the
collider handle and keeping only the time of impact. That handle *is* the
collision, and the hit carries the witness point and normal a
`CollisionContact` needs.

So the sweep now reports its blocking hit as a `CollisionEvent::CollisionStarted`,
and everything downstream already worked: `mission_core` dispatches it as
`Collided`, and the material lookup, schema query and `NO_COLLISION_SOUND` opt-out
are #1139's, unchanged.

## The four decisions

**Double-firing.** A melee weapon gets a real narrow-phase contact *and* would
get a sweep event — every consumer, damage included, fires twice for one hit.
The sweep event is scoped to `is_held_inert` bodies, which is also what keeps
the new script silent for a gun lying on the floor, thrown, or wielded flat.

**Who handles it.** A gun has no `TriggeredMeleeWeapon` (that is attached only
to `PropLimbModel` weapons), so a new `HeldItemImpactSound` script owns the
gun's side. The rate-limit guard is *shared* with melee rather than
reimplemented — `scripts/impact_sound.rs` now holds both, and `melee_weapon.rs`
gets 123 lines lighter.

**The speed guard survives the kinematic body** — the open question going in.
It reads velocity, and a held item is `KinematicPositionBased`; that is sound
because Rapier derives such a body's velocity each step from how far it
actually moved. Measured: an item arriving at a wall reports its real closing
speed on the frame it is stopped, and **0.000** on every frame it rests there.
Exactly the distinction the guard wants, so the sweep's own numbers were not
needed.

**Cooldown granularity, stated rather than glossed.** The key is an *entity*,
and a mission's entire static geometry is one collider owning one entity — so
it is one thud per 0.15 s for the whole level, plus one per prop. Right
granularity for an impact sound, and deliberately coarser than per-surface.

## Verification

- `cargo test -p shock2vr --lib`: **1040 passed**, 1 ignored.
- New: `a_held_gun_blocked_by_the_level_makes_one_impact_sound`,
  `a_repeat_block_within_the_cooldown_is_silent`,
  `a_gun_that_is_not_held_inert_stays_silent` — the last pinning the narrow
  scope above. #1139's five melee tests still pass unchanged, which is the
  no-double-fire and no-regression evidence.
- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo fmt --all -- --check` clean.
- e2e: `held-gun-impact-sound.e2e.test.ts`.

https://claude.ai/code/session_017cm626D5M4tAz7yLPWPMFb

### DIFF
--- a/shock2vr/src/mission/entity_creator.rs
+++ b/shock2vr/src/mission/entity_creator.rs
@@ -23,9 +23,10 @@ use dark::{
         FrobFlag, InternalPropOriginalModelName, Link, Links, PhysicsModelType, PoseType, PropAI,
         PropClassTag, PropCollisionType, PropCreature, PropCreaturePose, PropFrobInfo,
         PropHUDSelect, PropHasRefs, PropHitPoints, PropImmobile, PropKeySrc, PropLimbModel,
-        PropModelName, PropPhysAttr, PropPhysDimensions, PropPhysState, PropPhysType, PropPosition,
-        PropRenderType, PropScale, PropSymName, PropTemplateId, PropTranslatingDoor, PropTripFlags,
-        RenderType, StimPropagator, StimSourceOptions, TemplateLinks, WrappedEntityId,
+        PropModelName, PropPhysAttr, PropPhysDimensions, PropPhysState, PropPhysType,
+        PropPlayerGun, PropPosition, PropRenderType, PropScale, PropSymName, PropTemplateId,
+        PropTranslatingDoor, PropTripFlags, RenderType, StimPropagator, StimSourceOptions,
+        TemplateLinks, WrappedEntityId,
     },
     ss2_entity_info,
 };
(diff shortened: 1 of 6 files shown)


tommy-xr:
Fixed by #1142

tommy-xr:
Superseded by #1474 (material impacts for physically held guns).

PR #1156, opened 2026-08-24
feat(audio): play authored creature footsteps (+ assessment of the player half)

tommy-xr:
The owner asked for footstep sounds — player and creatures — and "what would be the lift?". So this is a **scoped assessment plus the half that turned out to be small**.

## The pivotal question first: does the shipped data have footstep audio?

**Yes, comprehensively.** Verifiable with `cargo dq sound`:

```
$ cargo dq sound +event:footstep +creaturetype:player +material:metal
Result: ftmet3 (volume -1000 millibels)
$ cargo dq sound +event:footstep +creaturetype:oncegrunt +material:fleshtarget +material2:metal
Result: ft_ogm1 (volume -1200 millibels)
$ cargo dq sound +event:footstep +creaturetype:monkey
Result: ft_monk1 (volume -1200 millibels)
```

The schema keys `event=footstep` on `creaturetype`, then refines: the player branches on the surface underfoot (`plasticrete`/`metal`/`glass`/`fabric`/`flesh`/`earth`, each with a heavier `landing=true` variant) and on water immersion (`medialevel` → wading / swimming / stroking); creatures mostly resolve on type alone to a four-sample set (`ft_monk1..4`, `ft_dro1..4`, `ft_rumb1..4`, `ft_pro1..4`, `ft_mw1..4`, `ft_spid1..4`, `ft_ninj1..4`), with hybrids alone branching further on the ground material. Full vocabulary in the assessment doc.

## What this PR does

**Creatures — implemented** (~30 lines of production code). The two halves already existed and nothing joined them:

- The motion format's per-frame flags are parsed into `AnimationClip::motion_flags`, `AnimationPlayer::update` returns the flags crossed this frame, and `mission_core` already dispatches them as `MessagePayload::AnimationFlagTriggered`.
- `script_util` already resolves schema queries and emits `Effect::PlayEnvironmentalSound`.

So: a new `script_util::play_footstep_sound`, and one more arm in `AnimatedMonsterAI::handle_message` beside the existing `FIRE` and `MELEE_CONTACT_START` arms — the same pattern `weapon_script` uses for `TRIGGER1`.

**The player — written up, not implemented.** The player has no creature animation and therefore no foot flag; a footstep has to be derived from locomotion state. `projects/footstep-sounds.md` has the design, the honest sizing (~a day), and the traps.

## Two things worth knowing

**No rate limiting is needed for creatures, and that is not luck.** The shipped locomotion clips are *per-half-step*: `ogpwlklt` and `ogpwlkrt` are separate clips, each carrying exactly one foot-plant flag. A walk cycle is exactly two footsteps, paced by the animation at whatever speed the creature is actually moving. No accumulator, no cooldown, no speed threshold.

**It is deliberately not gated on "is this a locomotion clip".** Scanning the 589 clips referenced by the shipped motion schemas, 219 carry foot-plant flags — including idles (`ogpidle2`), turns (`ogpturn`), staggers (`ogprecwd`) and death collapses (`ogpdie2`). Those are authored, not accidental: a hybrid shifting its weight while standing *does* plant a foot. So this arm fires on the flag wherever it appears and, unlike the `FIRE` and melee arms, deliberately does not consult `is_dead`.

## Known gap, shared by both halves

The port has no per-texture material lookup for world geometry (`RayCastResult` carries no surface id), so the ground material is the default bulkhead metal. For hybrids that resolves `ft_ogm*` rather than `ft_og*` — right for most of the ship, wrong on Hydroponics soil. Fixing it would upgrade footsteps, bullet impacts and melee impacts at once, so it belongs in its own change; §4 of the doc.

## Verification

- `cargo fmt --all -- --check` clean
- `cargo test -p shock2vr --lib` — 1074 passed (1071 on main, +3 new)
- Negative test done first: with the production arm removed, both foot-plant tests fail with `one plant, one footstep: []` (0 queries vs 1 expected); the non-foot-flag control test passes either way.

No visuals — this change is audible only; runtime audio evidence is in a comment below.

https://claude.ai/code/session_017cm626D5M4tAz7yLPWPMFb

### DIFF
--- a/projects/footstep-sounds.md
+++ b/projects/footstep-sounds.md
@@ -1,0 +1,196 @@
+# Footstep Sounds
+
+Assessment of what it takes to give the player and creatures footstep audio,
+plus the design for the player half (the creature half is implemented).
+
+## 1. The data supports this fully
+
+The shipped sound schema authors a complete footstep vocabulary. Resolve it
+with `cargo dq sound`:
+
+```
+$ cargo dq sound +event:footstep +creaturetype:player +material:metal
+Result: ftmet3 (volume -1000 millibels)
+$ cargo dq sound +event:footstep +creaturetype:oncegrunt +material:fleshtarget +material2:metal
+Result: ft_ogm1 (volume -1200 millibels)
+$ cargo dq sound +event:footstep +creaturetype:monkey
+Result: ft_monk1 (volume -1200 millibels)
+```
+
+The tree is keyed `event=footstep` first, then `creaturetype`, then per-type
+refinements. Two distinct shapes:
+
+**The player** branches on the surface underfoot and on how the foot arrived:
+
+| key | value | samples |
+| --- | --- | --- |
+| `material` | `plasticrete` | `ftpoly1..4` |
+| `material` | `metal` / `metalbig` / `metaldebris` / `metaltarget` | `ftmet1..4` |
+| `material` | `glass` | `fttil1..4` |
+| `material` | `glassbits` | `ftsnow1..4` |
+| `material` | `fabric` | `ftcar1..4` |
+| `material` | `flesh` / `fleshtarget` | `ftfle1..4` |
+| `material` | `earth` | `ftear1..4` |
+| `medialevel` | `foot` | `ftwat1..4` (wading) |
+| `medialevel` | `body` | `swimtop1..3` |
+| `medialevel` | `head` | `stroke1..3` (submerged swim stroke) |
+
+plus a `landing=true` sub-key under every material giving the heavier
+one-shot landing thud (`ftmetj`, `ftpolyj`, `fttilj`, `ftearj`, `ftsnowj`,
+`ftflej`). A player query with no `material` and no `medialevel` resolves to
+**nothing** — the player's branch requires one of them.
+
+**Creatures** mostly resolve on `creaturetype` alone to a four-sample set:
+
+| `creaturetype` | samples |
+| --- | --- |
+| `oncegrunt` (hybrid) | `ft_og1..4` (soft floors) / `ft_ogm1..4` (metal) |
+| `monkey` | `ft_monk1..4` |
+| `droid` | `ft_dro1..4` |
+| `protobot` | `ft_pro1..4` |
+| `rumbler` | `ft_rumb1..4` |
+| `midwife` | `ft_mw1..4` |
+| `spider` | `ft_spid1..4` |
+| `assassin` (ninja) | `ft_ninj1..4` |
+
+Hybrids are the only type that branches further: `material` = the creature's
+*own* material (its feet — `fleshtarget`), `material2` = the surface underfoot.
+No footsteps are authored for swarms, apparitions, SHODAN, servbots or the
+overlord; those resolve to nothing and are silent, which is correct.
+
+There is also an `event=climbstep` tag in the vocabulary, but nothing in the
+tree resolves under it for the player — ladder-step audio would need new
+sourcing and is out of scope.
+
+## 2. Creatures — implemented
+
+Cost: about an hour. Two small pieces:
+
+- `script_util::play_footstep_sound(world, entity_id)` builds the schema query
+  (`event=footstep` + class tags + `material`/`material2`) and emits
+  `Effect::PlayEnvironmentalSound` at the creature's transform, reusing the
+  existing `play_environmental_sound` helper.
+- `AnimatedMonsterAI::handle_message` handles `AnimationFlagTriggered` with
+  `LEFT_FOOT_STEP | RIGHT_FOOT_STEP`, beside the existing `FIRE` and
+  `MELEE_CONTACT_START` arms.
+
+Everything else already existed: the motion format's per-frame flags are parsed
+into `AnimationClip::motion_flags`, `AnimationPlayer::update` returns the flags
+crossed this frame, and `mission_core` already dispatches them as
+`AnimationFlagTriggered`.
+
+**Why no rate limiting is needed.** The shipped locomotion clips are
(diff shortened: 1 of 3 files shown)


tommy-xr:
## Runtime audio evidence (medsci1)

Headless debug runtime on `medsci1.mis`, player teleported next to a Blue Monkey to alert it; several hybrids and a droid were already patrolling. Sounds read from `/v1/audio/recent`.

**Right family per creature type:**

| creature | samples heard | resolved query |
| --- | --- | --- |
| hybrid | `ft_ogm1..4` (153 plays: 42/46/34/31) | `event=footstep, creaturetype=oncegrunt, material=fleshtarget, material2=metal` |
| droid | `ft_dro1`, `ft_dro2`, `ft_dro4` | `event=footstep, creaturetype=droid, material=metaltarget, material2=metal` |
| monkey | `ft_monk3` | `event=footstep, creaturetype=monkey, ...` |

The four samples in each pool are being drawn from properly rather than one repeating, and the monkey's footstep positions tracked its walk (z = -26.89 → -24.49 → -23.31) as it approached.

**Not flooding.** Isolating one hybrid's trail by position (a single patrol lane, x ≈ -7.1..-8.1, y = -4.70) and reading `sim_time`:

```
sim  5.633  ft_ogm1          sim  7.900  ft_ogm2   dt 0.367
sim  5.833  ft_ogm1  dt 0.200   sim  8.100  ft_ogm4   dt 0.200
sim  6.200  ft_ogm4  dt 0.367   sim  8.467  ft_ogm3   dt 0.367
sim  6.400  ft_ogm1  dt 0.200   sim  8.667  ft_ogm1   dt 0.200
sim  6.967  ft_ogm1  dt 0.567   sim  9.033  ft_ogm2   dt 0.367
sim  7.333  ft_ogm1  dt 0.367   sim  9.233  ft_ogm1   dt 0.200
sim  7.533  ft_ogm4  dt 0.200   sim  9.600  ft_ogm3   dt 0.367
                                sim  9.800  ft_ogm1   dt 0.200
                                sim 10.167  ft_ogm4   dt 0.367
```

16 footsteps in 4.53 s of sim time — **~3.5/s** for an alerted, jogging hybrid, against 60/s if this were firing per frame.

The alternating `0.200 / 0.367` spacing is not a coincidence, and it is the clearest confirmation that the pacing comes from the animation rather than from anything in this code:

```
$ cargo dq motion-info oggwlkl
  duration:     0.367
  frame   10:   LEFT_FOOT_STEP | RIGHT_FOOT_UP
```

The clip is 0.367 s long with its single foot plant at frame 10 (0.200 s in). The observed footstep spacing *is* the clip's own plant-to-plant timing, played back to back. A creature that slows down steps more slowly for free; there is no timer to retune.

**Idle control:** before any AI woke, zero `ft_*` events (the buffer held only camera speech clips). Footsteps began exactly when `AIBehavior` flipped `Idle` → `Wander`.

No anomalies attributable to this change. (The player died to environmental damage near the cryo area mid-session — unrelated, and did not affect the capture.)

tommy-xr:
## Review pass

Cross-engine review; the Codex half was unavailable (out of usage credits, and its first run silently returned findings for an unrelated branch — worth knowing that failure mode is not always loud). The Claude reviewer found four real problems, all now fixed in a740152e:

**The test fixture keyed hybrids on a creature type the schema does not have.** The shared AI test fixture uses `"creaturetype hybrid"`; the shipped value is `oncegrunt`. This mattered more than a wrong string, because `EnvSoundQuery::to_tag_query` silently `continue`s past an unknown tag *value* rather than failing the query — so the test asserted `creaturetype=hybrid` was present in a query that, against real data, would have dropped `creaturetype` entirely. Fixed, and verified against the gamesys.

**The part most able to fail quietly was the part not asserted.** Hybrids are the one type needing both `material` and `material2`, and missing either resolves to *silence*, not to a wrong-but-audible sound:

```
+creaturetype:oncegrunt                                   -> No matching sounds found
+creaturetype:oncegrunt +material:metal +material2:metal   -> No matching sounds found
+creaturetype:oncegrunt +material:flesh +material2:metal   -> No matching sounds found
```

So if a hybrid archetype ever lost its `PropMaterial`, `get_impact_material`'s `"metal"` fallback would make it *mute* rather than mis-pitched. The fixture now carries the shipped `PropMaterial("Material FleshTarget")` and the test pins both tags.

**Flags do not always arrive alone — a genuine bug.** `AnimationPlayer::update` *unions* every flag frame crossed in a tick, so on a hitch one message can carry `FIRE | LEFT_FOOT_STEP`. As an `else if` arm the footstep was silently dropped; worse, on a dying creature the `FIRE` arm's early `return` suppressed it via a guard that has nothing to say about where a creature's feet are. The footstep is now resolved independently of the attack chain and combined. New test `a_foot_plant_sharing_a_tick_with_an_attack_flag_still_sounds` fails under the old chain (isolated: that one test and only that one) and passes now.

**No panic on a missing transform.** `play_environmental_sound` unwrapped `RuntimePropTransform`. Animation players and physics bodies live in separate maps and ownership changes hands around death — which is exactly the state the footstep arm deliberately runs in. Missing transform is now a silent no-sound.

### One finding accepted as behaviour, recorded rather than changed

`ogpidle2` is 5.53 s with 5 foot plants (~0.9/s), about the same cadence as walking — so a standing creature keeps ticking and standing is not audibly distinguishable from walking. That is what the data authors, and in the medsci1 run no idle creature produced a single footstep (they don't animate until alerted). Recorded in §"Open question: idle cadence" of the doc; if it grates in play the fix is a per-creature cooldown, not a locomotion gate.

Re-validated after the fixes: `cargo fmt --all -- --check` clean, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean, `cargo test -p shock2vr --lib` 1075 passed / 0 failed.

Issue #1170, opened 2026-08-25
test(e2e): four suites fail on main (baseline)


tommy-xr:
Known e2e failures on unmodified main as of 2026-08-25, each confirmed pre-existing by at least one agent building clean main in isolation during the interaction campaign:

- `creature-loot.e2e.test.ts`
- `earth-hacking.e2e.test.ts` (nanite-pickup assertion)
- `log-reader.e2e.test.ts` — hydro3 Korenchkin transcript case
- `bio-ammo-full.e2e.test.ts` — "loaded magazine must not expose a conversion control" (empty-multi-ammo-weapon case)

These poison full-suite runs as a merge gate (compounded by the runner's exit-0 bug, filed separately). Fix or quarantine with a documented skip reason.

tommy-xr:
Baseline has grown: on clean main (d3c63cfb, 2026-08-29) three more suites fail, confirmed by an isolated re-run of just these files:

- `baby-arachnid.e2e.test.ts` — Hydro3 Baby Arachnids pushed aside / hit by normal weapons
- `medsci-saved-vr-melee.e2e.test.ts` — canonical 9-damage VR melee across decks and saves (likely related to the #1160/#1173 melee billing changes)
- `trap-unlock.e2e.test.ts` — eng1 locked grav-lift button unlock trap

That makes 7 known-failing on main including the original four. (The verification run also reproduced the exit-0-despite-failures bug — 3 fails, exit code 0.)

tommy-xr:
Superseded by #1262 (same e2e baseline on newer main, with the four suites here plus four more).

PR #1181, opened 2026-08-29
fix(render): depth-bias flat decal models so floor signs stop z-fighting

tommy-xr:
Signs authored flat in the horizontal plane (station.mis floor signs: `block`, `Clear1`, `caution`, `C_Shut`, `mispost`, ...) sit exactly coplanar with the level floor and z-fight it — the decal is patchily eaten by the floor in both flat and VR presentations.

The existing decal fix only handles models paper-thin along local **X**, whose outward normal is known (local -X) and can be lifted off the surface. For Y/Z-thin flats the face direction cannot be read off the bounding box, so instead give them a **polygon-offset depth bias** at draw time (view-relative, direction-free), applied on the **opaque pass only**.

- `decal_needs_depth_bias` (entity_creator): Y/Z-thin, both in-plane extents must dwarf the thin axis — thin rods and small solid props stay out (unit-tested).
- `SceneObject::draw_geometry` brackets the draw with `glPolygonOffset(-1, -2)`; state restored unconditionally.
- `/v1/scene` now reports `depth_bias` per object, so the heuristic is headlessly auditable: on station.mis the biased set is signs/decals only (`caution`, `C_Shut`, `mispost`, `Clear1`, `block`, strp strips, letter tiles, ...).

z-fight before/after demo

<sub>Camera sliding over station.mis's "Do Not Block" floor sign: before, the floor eats most of the decal and the patches crawl with the viewpoint; after, it renders whole. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

## Before → after (flat and VR, identical framing)

before/after

Known gap (pre-existing): decals drawn on the transparent pass get neither the lift nor the bias.


tommy-xr:
e2e triage: full suite on this branch showed 7 failures — the 4 documented in #1170 plus baby-arachnid, medsci-saved-vr-melee, and trap-unlock. All 3 extras were re-run in isolation on clean main (d3c63cfb) and fail there identically, so none are caused by this change (noted on #1170). No new failures attributable to this PR.

PR #1224, opened 2026-08-31
feat(debug): developer cheat pad (X+Y empty-hand chord, weapon/nanite rain)

tommy-xr:
Adds a small developer cheat pad: a two-button overlay that rains useful items around the player, so a test situation can be set up without playing the game up to it.

## What

- **Gate**: a new `cheats` bool dev param (`shock2vr/src/dev_params.rs`). With it off the open action is never read, so neither a stray controller chord nor an HTTP injection can conjure free weapons into an ordinary run. Turning it off while the pad is up closes it.
- **Open**: `InputAction::ToggleCheatPad` - `Alt+K` on desktop, left `X`+`Y` chord on the Quest, gated in VR on that hand being **empty** (`GameScene::hand_is_empty`). Same chord, or the pad's `Done` button, dismisses it.
- **UI**: a `Game`-owned overlay modelled on `pause_menu` (so every pausable scene gets it, the world keeps rendering behind it, only `update` is skipped). The page is a **scrolling list** built on `ui::list_scroll` - the same paging, gutter rocker, inert-end rule and arrow art the Developer screen's parameter page and debug-scene launcher use - at the launcher's 19px row pitch and font, on the Developer page's authored widget rects (`GAMELODR.BIN` over `GAMELOD.PCX`). Placement is decided once in canvas pixels, so flat and VR draw the identical canvas (AGENTS.md section 3).
- **Adding a cheat is one entry in `CHEATS`.** The list pages and grows a scroll rocker on its own once the entries outrun the pane; rows carry the index of the entry *scrolled into* them, never their own slot.
- **Effect**: one new `Effect::RainItems { template_ids }`, applied in `mission_core` as a deterministic ring 2.5 units above the player.

Templates (from `cargo dq`): weapons rain `-928` Wrench, `-17` Pistol, `-19` Shotgun, `-18` Assault Rifle, `-1358` x2 / `-1360` Small Standard/AP Clip, `-42` Pellet Shot Box. Modules rain `-938` EXP Cookies x4 and `-89` 20 Nanites x4.

## Verification

- `cargo test -p shock2vr --lib`: 1175 pass, including new unit tests for the hit test, the chord binding, and the press-edge rule (the chord that opens the pad must not click a button on the first frame).
- New `tools/shock2-sdk/test/cheat-pad.e2e.test.ts`: gate off -> the action and a click on the button are inert; gate on -> the pad opens, rain-weapons adds a wrench that owns a dynamic body and settles, rain-modules adds exactly 4 module stacks + 4 nanite piles. **Negative-verified**: removing both gate checks makes it fail, restoring them makes it pass.
- Rendered and photographed in **both** presentations (below): labels land in the same place relative to their widgets.
- `cargo apk check` passes for `oculus_runtime` (the chord wiring lives there).
- `/xreview` (Claude + Codex + a de-dup pass) run and its Critical/High findings fixed:
  - the pause menu could open *behind* an already-open cheat pad, so the pad's opaque backdrop hid the menu that was actually taking the clicks (and "Quit to Main Menu" sits under the pad's own rows) &mdash; one overlay is up at a time now, asserted in the e2e test;
  - `hand_is_empty` was implemented on `MissionCore` only, but `Game` holds the `Mission` / `DebugScene` wrappers, which kept the trait default `true` and left the empty-hand gate inert everywhere it actually runs;
  - opening the pad now closes use mode, as the pause menu does &mdash; on the Quest the first half of the `X`+`Y` chord fires `X`'s own action first;
  - `is_paused()` reports every overlay that suspends the scene, so automation cannot read a frozen sim as running;
  - the *close* branch of the toggle is ungated, so a pad that is up is always closable by the chord that opened it.
- Capturing the GIF also caught a real bug the tests had not: the ring spawned 6.25 ft above the player's body origin, high enough to land **on top of corridor ceiling geometry** instead of reaching the player. It is now 3.5 ft over the origin (just above the standing capsule's 3 ft crown), written in SS2 feet over `SCALE_FACTOR` like the spawn handlers beside it.

## Revision: the list UI

Per maintainer feedback the pad was reworked from two fixed-rect buttons into a scrollable list, so it scales as cheats are added. Gating, chord, `RainItems` and the two entries are unchanged.

**What was reused:** `shock2vr/src/ui/list_scroll.rs` — already the shared home for the Developer screen's two lists. Its *primitives* (`rows_per_page` / `max_scroll` / `visible_rows` / `rocker` / `hit` / `apply` / the arrow art) dropped straight in.

**The friction, and what it turned into:** review caught that reusing the primitives still left me copying the *composition* around them — `rows_per_page`/`max_scroll`/`visible_rows`/`rocker`/`row_rect`/`text_rect` were character-for-character the debug-scene launcher's `scene_*` helpers, which are themselves a near-copy at a third pitch inside `dev_params_panel`. A third copy would have made "the lists scroll the same way" a convention rather than a fact, which is the opposite of the ask. So `list_scroll::ListGeometry` now owns that composition — including the slot-to-item mapping and the rule that a row clears the gutter *exactly* when the rocker exists — and all three lists go through it. Net: the launcher and the parameter page each lost their copy.

Two more review fixes rode along: `list_scroll::apply` now clamps a stranded offset before it moves (an over-scroll previously needed dead `Up` clicks to unstick), and the pad's `hit`/`build_canvas` take the cheat entries rather than a length, so a row cannot be hit-tested without a cheat behind it.

**One player-visible change** beyond the layout: rows are the launcher's 19px rather than the old 40px buttons, so the VR ray target is smaller. That is the pitch the launcher already ships with in VR, and the VR still below is a fresh capture of the list — but it is a real tradeoff, noted rather than hidden.

## Gaps

- The VR *click* was not exercised headlessly: the offscreen `--vr` runtime's controller ray could not be steered onto a row from HTTP (patched hand rotations stick but the rendered ray does not follow). The panel and its aim ray render correctly, and the hit test is the same shared function the flat path clicks through, but a real headset press is unverified.
- Raining twice onto the same spot piles items on each other and the solver flings them apart; a single press settles cleanly.
- The empty-hand gate itself has no automated test &mdash; filling a VR hand headlessly is expensive. Its wiring is the piece review caught, so it is worth a follow-up.
- `dev_params_panel` keeps its own four-column `row_rects` (a parameter row is a label plus `<`/value/`>`, not one target), so only its paging moved to `ListGeometry`.
- Codex hit its API usage limit part-way through the revision's `/xreview`, so that delta pass is single-engine (the Claude reviewer) rather than cross-engine.
- Deferred de-dup findings, all deliberate to keep this diff small: `cheat_pad` repeats `pause_menu`'s overlay shell (open / close / held-press latch / render), and `Game` repeats its update-and-close pair &mdash; extracting a shared `FrontendOverlay<T>` would cut ~150 lines and make the "two overlays open at once" bug structurally impossible, but it is a refactor of shipped UI code rather than part of this feature. Same for the two near-identical chorded-pair blocks in `oculus_runtime`, and the `click` helper this e2e test shares verbatim with `dev-menu.e2e.test.ts`.

## Visuals

cheat pad demo

<sub>Flat, captured headlessly on the debug runtime: the pad (now a list) opens, "Rain weapons" is pressed, "Done" closes it, and the spread drops and settles at the player's feet.</sub>

flat cheat pad open

<sub>Flat &mdash; the pad</sub>

the rained weapons

<sub>Flat &mdash; one press of "Rain weapons", settled</sub>

VR cheat pad open

<sub>VR &mdash; the same list canvas on a world-space panel, with the controller aim ray and its hit dot. Rows land in the same place relative to the pane as they do flat.</sub>

## Full e2e suite

`SHOCK2_E2E=1 npm run test:e2e` on this branch (`ea2ba14`): **126 pass / 4 fail**.

- `Earth hackable keypad ...` and `creature loot: ... Watts yields the 12451 disc` &mdash; already known to fail on `main`.
- `Hydro3 Baby Arachnids can be pushed aside and hit by normal weapons` and `use mode exposes AMMOFULL ammo-cycle for an empty multi-ammo weapon` &mdash; **verified pre-existing**, not caused by this branch. Re-ran both against `origin/main` (`9c4b3f66`, fresh `cargo build -p debug_runtime`) in isolation and they fail identically:
  - baby-arachnid: `an aimed wrench swing should damage the Baby Arachnid` &mdash; `expected: true, actual: false`
  - bio-ammo-full: `loaded magazine must not expose a conversion control` &mdash; `expected: true, actual: false`

All 4 failures are pre-existing on `main`; none are caused by this PR.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -128,6 +128,7 @@ Debug bindings take `Alt` (`Option` on macOS) to keep them clear of gameplay key
 | `Alt+G` | `DebugForceChase` | every monster hunts the player, pinned |
 | `Alt+C` | `DebugCalmAll` | clears the pin |
 | `Alt+V` | `ToggleFreeCamera` | requires the **Free camera** developer option; Quest: right `A`+`B` together |
+| `Alt+K` | `ToggleCheatPad` | requires the **Cheats** developer option; Quest: left `X`+`Y` together, on an empty left hand |
 
 Every action is also triggerable without a keyboard - over HTTP on the debug
 runtime (`POST /v1/input/action`) or through the SDK (`game.input.trigger(...)`).
@@ -150,6 +151,14 @@ shows exactly what the *player's* viewpoint decided to draw. Turn on **Cull
 from cam** when you would rather just look at things. The developer option is
 the gate: while it is off the `Alt+V` / `A`+`B` toggle does nothing, and turning
 it off while detached re-attaches the camera.
+
+**Cheats** arms the developer cheat pad: a two-button panel (`Alt+K`, or left
+`X`+`Y` on the Quest with nothing in that hand) that rains a spread of weapons
+and ammo, or of cyber modules and nanites, down around the player so a test
+situation can be set up without playing the game up to it. Like the free
+camera, the developer option is the gate - with it off the toggle does nothing
+from any input path, HTTP injection included, and turning it off while the pad
+is up closes it.
 
 Flying uses the ordinary locomotion controls, so they are the ones your hands
 already know - and while the camera has them, the player stands still rather
--- a/runtimes/desktop_runtime/src/input_mapper.rs
+++ b/runtimes/desktop_runtime/src/input_mapper.rs
@@ -189,6 +189,13 @@ impl DesktopInputMapper {
                 modifier: Modifier::Alt,
                 action: InputAction::ToggleFreeCamera,
             },
+            // Open/close the developer cheat pad. Alt-modified and gated on
+            // the `cheats` dev param, exactly like the free camera above.
+            Binding {
+                key: Key::K,
+                modifier: Modifier::Alt,
+                action: InputAction::ToggleCheatPad,
+            },
         ];
 
         Self {
--- a/runtimes/oculus_runtime/src/lib.rs
+++ b/runtimes/oculus_runtime/src/lib.rs
@@ -821,16 +821,35 @@ fn main() {
             }
             crouch_button_was_pressed = crouch_state.current_state;
         }
-        action_state.sync_discrete_button(
-            shock2vr::input::InputAction::ToggleUseMode,
-            use_mode_state.is_active,
-            use_mode_state.changed_since_last_sync,
+        // Left X and Y carry two meanings, on the same terms right A and B do
+        // below: alone they toggle the cyber interface and play the last audio
+        // log, together they toggle the developer cheat pad. The sharing is
+        // dead unless the `cheats` developer option is on, and while it is on
+        // the button whose partner is ALREADY held is suppressed so completing
+        // the chord cannot also open the interface.
+        let cheats_gate = shock2vr::dev_params::get_bool(shock2vr::dev_params::CHEATS);
+        let suppress_use_mode = cheats_gate && audio_log_state.current_state;
+        let suppress_audio_log = cheats_gate && use_mode_state.current_state;
+        if !suppress_use_mode {
+            action_state.sync_discrete_button(
+                shock2vr::input::InputAction::ToggleUseMode,
+                use_mode_state.is_active,
+                use_mode_state.changed_since_last_sync,
+                use_mode_state.current_state,
+            );
+        }
+        if !suppress_audio_log {
+            action_state.sync_discrete_button(
+                shock2vr::input::InputAction::ReadLastUnreadLog,
+                audio_log_state.is_active,
+                audio_log_state.changed_since_last_sync,
+                audio_log_state.current_state,
+            );
+        }
+        action_state.sync_chord(
+            shock2vr::input::InputAction::ToggleCheatPad,
+            use_mode_state.is_active && audio_log_state.is_active,
             use_mode_state.current_state,
-        );
-        action_state.sync_discrete_button(
-            shock2vr::input::InputAction::ReadLastUnreadLog,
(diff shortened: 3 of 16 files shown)


tommy-xr:
Superseded by #1508, which delivers the same two cheats and the same `Effect::RainItems` through the Developer screen the game already has, rather than a separate overlay.

Per maintainer direction: the main menu's Developer screen has a **Scenes** category behind its upper framed button, and the in-game Developer page leaves that same slot empty — so the in-game screen gets a **Cheats** category there instead. A cheat needs a running mission; a scene launcher needs there not to be one.

That drops everything this PR built around the pad: `cheat_pad.rs`, `InputAction::ToggleCheatPad`, the `Alt+K` binding, the Quest `X`+`Y` empty-hand chord (so `oculus_runtime` is untouched), `GameScene::hand_is_empty`, and the `cheats` dev-param gate — the Developer screen is already the gate. It also makes the "two overlays open at once" bug this PR had to fix structurally impossible, since there is only ever one overlay.

Carried over unchanged: `Effect::RainItems` and its `mission_core` ring handler, including the 3.5 ft spawn height (at 6.25 ft the rain landed on top of corridor ceilings).

The one genuinely reusable piece of this branch — the `list_scroll::ListGeometry` extraction from `ea2ba14`, which de-duplicates the scene launcher's and `dev_params_panel`'s list composition — is split out on its own as #1507, with #1508 stacked on it.

Closing in favor of #1507 + #1508.

PR #1227, opened 2026-08-31
feat(dark_explorer): preview .fon fonts with sample strings

tommy-xr:
Selecting a `.FON` in `cargo dx` showed a hex dump. It now shows what the font *is*: the header metadata (format variant, glyph count and code range, height, row width) and a pangram rasterized from the font's own glyphs at 1x/2x/3x — the fonts are bitmaps, so whole multiples of the native height are the only meaningful "sizes".

The preview reuses the shipping parser rather than adding a second one. The glyph pixel decode moves out of `Font::read` into a new no-GL `FontBitmap` in `dark/src/font.rs` (metrics + bitmap strip, `glyph_alpha`, `render_string`); `Font::read` now builds its texture atlas from that, so the in-game text path and the preview decode pixels through one function.

All three shipped variants render: packed format 0 (`MAINFONT.FON`), format 1 antialias-16 (`METAFONT.FON` — the variant that has bitten before), and 0xCCCC 8-bit alpha (`MAINAA.FON`). A format the parser does not specially handle is labeled "unknown (treated as 8-bit alpha)" rather than silently mislabeled. Reads past a truncated bitmap strip now yield transparent instead of panicking.

## Visuals

`MAINFONT.FON` — format 0, packed 1bpp:

mainfont

`METAFONT.FON` — format 1, antialias-16:

metafont

`MAINAA.FON` — format 0xCCCC, 8-bit alpha (dimmer because the shared decoder's `> 205 -> 0` clamp is faithful to what the game draws):

mainaa

No GIF: the preview is static.

## Verification

- `cargo test -p dark --lib font` — 8 tests green, including three new ones for the rasterizer (packed 1bpp unpacking, format-1 coverage scaling + glyph spacing, truncated-strip safety).
- `RUSTFLAGS="-D warnings" cargo check -p dark -p dark_explorer -p shock2vr`.
- Screenshots above captured headlessly: `cargo dx ui --select iface/fonts/mainfont.fon --screenshot <out>.png` (likewise `intrface/metafont.fon`, `iface/fonts/mainaa.fon`).

- In-game text path re-verified after the `Font::read` refactor: the main menu (METAFONT) still renders correctly through the debug runtime, and the SDK e2e suite was run.

## Review

`/xreview` (Claude + Codex + a de-dup pass): **no Critical/High**. Applied: the sample now lays out on glyph advances (the UI's `measure_text_width` adds nothing between glyphs — only the dark_viewer-only `Font::get_mesh` uses `SPACING`), the file is read once instead of twice, and a truncated bitmap strip logs a warning.

Both engines flagged that reads past a truncated strip now decode as transparent where `Font::read` previously panicked. Kept deliberately: the explorer must tolerate whatever file the user clicked, and the leniency never fires on real data (every shipped font's strip is exactly `row_width * height`) — the new `warn!` makes a bad asset visible in the log instead of silently blank.

Continues the dark_explorer campaign (stack #1193, now merged); branches from `main`.

### DIFF
--- a/dark/src/font.rs
+++ b/dark/src/font.rs
@@ -16,7 +16,7 @@ use engine::{
     texture_atlas::{TexturePackResult, TexturePacker},
 };
 use image::ImageBuffer;
-use tracing::info;
+use tracing::{info, warn};
 
 use crate::ss2_common::{read_bytes, read_i16, read_u8, read_u16, read_u32};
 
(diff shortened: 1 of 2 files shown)


tommy-xr:
Full-suite run on this branch was lost to a session interrupt; re-ran the relevant safety net for the `dark/` font refactor: `missions.e2e.test.ts` on this branch — **all 23 missions load, 23/23 pass**.

PR #1242, opened 2026-09-02
feat(tools): AI reachability evaluation harness

tommy-xr:
Step 1 of the pathfinding evaluation plan: measure whether AIs actually get where they are going, so every later steering/graph fix has before/after numbers.

## What it does

`npm run ai-reachability -- --mission medsci2.mis --out /tmp/aire` runs two passes per mission, each on a freshly launched runtime:

- **idle** — no player input; the AIs run their authored patrols (3600 frames).
- **chase** — the player stands at a fixed spot, `DebugForceChase` pins everyone onto them (1800 frames).

Every AI is discovered by its AI runtime properties (never by a hardcoded runtime id) and sampled every 30 frames: position, yaw, `AIBehavior`/`AIAlertness`, its `/v1/ai/paths` entry (outcome, live waypoint, live path length, stall seconds, live target) and its distance to the player. At the end of a pass each AI is classified `arrived` / `progressing` / `wedged` / `no_route` / `expected_unreachable` / `idle_by_design`.

Outputs per mission: `<out>/<mission>.json` (every sample + classifications), `<out>/summary.md` plus the same table on stdout, and a top-down trail map per pass.

The expected-unreachable filter is data-driven, no hand lists: an AI that never travels (ScriptedSequence/Noop/Dead/SelfDestruct), one whose alertness cannot be lifted off `Lowest` by a forced chase (the mission's alert cap), and one with no walk route from its start cell to the player (only counted when both endpoints resolved to a nav cell, so an off-mesh sample is never mistaken for "by design"). A wedge outranks all of them, so an AI that is physically stuck is still reported even where it could never have reached the player.

## Supporting changes

- `GET /v1/pathfinding/route?from=x,y,z&to=x,y,z` — does a walk route exist between two positions? This is what separates "the AI failed to route" from "nothing walkable connects it to the player".
- `cargo bn path dump <mission>` — every nav cell as JSON (polygon, center, flags, walk component), the geometry the trail map draws over; `cargo bn path missions` lists the missions the engine can actually open (they live inside an archive on a 25AE install).
- `scripts/render-trail-map.py` — nav cells faint, one colored polyline per AI, dot at the start, square at the end, red ring at each wedge, magenta cross at the player.

## Results (this branch, no experimental flags)

| mission | pass | AIs | arrived | progressing | wedged | no_route | expected_unreachable | idle_by_design |
|---|---|---|---|---|---|---|---|---|
[...]

PR #1248, opened 2026-09-02
fix(ai): never steer at a goal with no route, and whisker past geometry the mesh misses

tommy-xr:
Fixes the user-reported medsci2 railing wedge: a patrolling hybrid pins itself in the balcony's south corner, sweeps its yaw for a few seconds, then stands there for the rest of the mission.

**Root cause.** OG-Pipe template 668 patrols six markers; marker 649 sits on the floor *below* the balcony, in a disconnected walk component (no link chain even under a stressed search - a genuine data dead end). A* therefore answers with no route at all. With no route, the AI kept the heading it had - straight at the marker - and pressed into the converging corner (45 degree wall plus a thin `Railing End` collider). Static-geometry whisker avoidance could not help: it only ran when *no* path existed at all, and its single body-centre ray sails just over a railing whose top sits a few inches below an upright creature's centre. After 3 stall skips the patrol watchdog retired the route and the AI went Idle permanently.

## What changed

1. **A patrol skips a point it has no route to** (`patrol_behavior.rs`, `path_follow_steering_strategy.rs`). Steering now reports an unreachable goal - A* answered with a *partial* route whose last waypoint stops more than 6 Dark feet short, in ground distance or in height - and the patrol advances to the next point at once instead of steering at it on an unchecked heading. One bad marker never retires a patrol; the route ends only if it comes back around to a point it already gave up on (so a wholly unreachable loop still stops, rather than cycling and re-querying forever).
2. **Whiskers while path-following** (new `steering/whisker_avoidance.rs`). Static-geometry whiskers now bias the path follower's aim point - capped at 3 Dark feet, exactly like the existing crowd-separation bias, so the route stays authoritative - and are sampled every 200 ms per AI rather than every frame. Probes go out at **two heights** (knee and chest, relative to the body origin): a railing is a thin slab that a single centre ray misses, and requiring *both* heights to be blocked also keeps stair risers and kerbs from bending routes sideways.
3. **Whiskers steer *around*, not just away.** A surface square across the route answers with a normal pointing straight back, which only stops the body in front of it; the blocked centre whisker now also pushes toward whichever side has room, and the aim point keeps only the sideways part of a backward-pointing bias.

The chase chain order is unchanged, and the stall/nudge logic and `pathfinding/mod.rs` are untouched (separate PRs).

## Before / after at the corner

Free camera on the balcony (medsci2, no player input, deterministic stepping). Before: the hybrid is pinned in the corner and only its idle animation moves - same spot at every timestamp. After: the corridor is clear while it is away on its route, and it walks back through.

contact sheet

| before (origin/main) | after |
|---|---|
| before | after |

## Measurements

**Repeated medsci2 idle observation of template 668** (60 s, no player input; longest window the body stayed within 1 world unit, and whether it was still patrolling at the end). This is the direct measure of the reported bug. AI path queries run on a worker thread, so runs vary - hence the repetition.

| build | runs | still Patrol at 60 s | longest hold (s) | ground covered (world units) |
|---|---|---|---|---|
| origin/main | 8 | **0/8** | 38-57 | 8.6, 9.0, 9.1, 9.1, 9.3, 37.2, 37.5, 72.8 |
| this PR | 8 | **4/8** | 4-48 | 21.1, 21.2, 57.1, 92.9, 127.5, 127.5, 128.0, 162.2 |

Every baseline run ends with the patrol retired (Idle) and the body parked in the corner; half the runs here walk the route for the full minute, and the median ground covered goes from 9 to ~110 units.

**AI reachability harness** (#1242's `ai-reachability`, idle 3600 frames + force-chase 1800 frames per mission, one run each):

before (origin/main)

| mission | pass | AIs | arrived | progressing | wedged | no_route | expected_unreachable | idle_by_design |
|---|---|---|---|---|---|---|---|---|
| medsci2.mis | idle | 18 | 0 | 6 | 5 | 0 | 0 | 7 |
| medsci2.mis | chase | 18 | 1 | 9 | 0 | 0 | 8 | 0 |
| medsci1.mis | idle | 26 | 0 | 13 | 1 | 0 | 0 | 12 |
| medsci1.mis | chase | 26 | 3 | 0 | 6 | 0 | 17 | 0 |

after

| mission | pass | AIs | arrived | progressing | wedged | no_route | expected_unreachable | idle_by_design |
|---|---|---|---|---|---|---|---|---|
| medsci2.mis | idle | 18 | 0 | **9** | **2** | 0 | 0 | 7 |
| medsci2.mis | chase | 18 | 2 | 8 | 0 | 0 | 8 | 0 |
| medsci1.mis | idle | 26 | 0 | 13 | 1 | 0 | 0 | 12 |
| medsci1.mis | chase | 26 | 3 | 0 | 7 | 0 | 16 | 0 |

medsci2's idle pass goes from 5 wedged / 6 progressing to 2 wedged / 9 progressing; medsci1 is unchanged, which is what a change that only bends aim points should do. (Template 668 itself happened to draw a bad run here - 47 s wedged - which is why the repeated table above, not a single harness pass, is the measure for that one creature.)

medsci2 idle trail maps (AI trails over the nav cells):

| before | after |
|---|---|
| trail before | trail after |

## Tests

- New unit tests: the patrol skip and its two termination cases (stubbed steering outcomes), route-reaches-goal including the stacked-floor case, the whisker bias math (sideways escape, cap, ignored floors), and the aim-bias projection (sideways-only, never onto the body).
- New e2e `tools/shock2-sdk/test/medsci2-patrol-railing.e2e.test.ts`: loads medsci2, no input, samples template 668 for 60 s and asserts it never holds one spot for 6 s and is still patrolling at the end. **Fails on origin/main** (`held 6s ... 48s at 49.00,0.10,-98.45`, behavior flips to Idle), passes here.
- `cargo test -p shock2vr` 1220 pass; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo fmt`.
- e2e: `patrol` (4/4), `force-chase`, `chase-last-seen`, `missions` (23/23); the new railing test passes 1 run in 4 (see the follow-up section above).

## Follow-up in this branch: a wedge must not retire the patrol

Tracing the e2e failures (`shock2vr::scripts::ai=debug`, medsci2, 3600 stepped frames) showed the residual failure was **not** an unreachable-point sweep. In every failing run the body wedged in the corner and the *stall* watchdog then retired the whole route:

```
05:51:07 patrol: stalled on point EId(376.0), skip #1
05:51:12 patrol: stalled on point EId(702.0), skip #2
05:51:12 patrol: no route to point EId(152.0), skipping
05:51:18 patrol: stalled on point EId(612.0), skip #3   -> route retired, AI Idle for the rest of the mission
```

Two defects behind that:

- **Three stall skips retired the patrol permanently.** A stall is a fact about the spot the body is standing in, not about the route - the AI usually walks out of it on the next leg. Both give-up paths (no route, and stalled) now share one ledger, cleared on arrival: the route ends only when *every* point on it has been given up on with no arrival in between.
- **A patrol adopted another goal's route.** Path queries are keyed by entity, so an answer outlives the strategy that asked for it: a patrol that gave up on a point built a fresh follower while the old query was in flight, then adopted that answer as its own. Because a fixed-point target never re-paths while it holds a route, it then steered along the abandoned point's route forever, carrying that goal's unreachable verdict onto a point nothing had asked about. A response is now only adopted if it answers the goal the follower actually wants.

New unit tests: a wedged patroller gets the whole loop's worth of tries before retiring (fails before the change - the old code retired after three), and an answer for an abandoned point is not adopted.

**Railing e2e after this commit: 1/4** (was 1-2/4). The retire-to-Idle failure mode is gone - the failing runs now stay in Patrol - but the run still fails on the 6-second hold rule, because the body is *physically* wedged and skipping to another point does not free it. That wedge is the stall/blacklist work in the stacked #1264, where it is also 1/4 today.

## Known remaining

About half the runs still end in a *different* wedge on the same balcony: the AI has a **full** route and simply presses against geometry the navigation mesh doesn't model, which the whisker bias (capped, never a veto) does not always clear. That is the stall/blacklist half of the ledger step - deliberately not touched here.

## Review

Ran the two-engine adversarial review plus a dedicated duplication pass. Fixed: unreachable skips are now bounded (a route that comes back around to a point it already gave up on retires, so nothing cycles or re-queries forever) but no longer spend the stall budget; only a *partial* route can declare a goal unreachable, and the check looks at height too (the balcony case is a goal almost directly below); the verdict is cleared whenever a fresh query goes out; whiskers no longer see living bodies or the player (that is crowd separation's job, and it repelled a chasing AI from its target); probe heights follow the creature's own height (fixed feet put both probes underground for a monkey); separation and whiskers are capped once, together; the e2e asserts a single matching patroller, Patrol throughout, and real ground covered; the stills were replaced with the timestamped contact sheet above.

Deliberately **not** done: merging the whisker probe with `collision_avoidance_steering_strategy.rs`'s. The two responses differ (a capped aim-point bias vs a heading override), the chain short-circuits so they never both run, and refactoring the existing no-route fallback is outside this fix - flagged as a follow-up rather than folded in.

### DIFF
--- a/shock2vr/src/scripts/ai/behavior/patrol_behavior.rs
+++ b/shock2vr/src/scripts/ai/behavior/patrol_behavior.rs
@@ -111,8 +111,10 @@ impl PatrolBehavior {
 
     fn steering_to(goal: Vector3<f32>) -> Box<dyn SteeringStrategy> {
         steering::chained(vec![
-            // Path steering leads; whisker avoidance only covers the no-route
-            // case (see chase_behavior for the deadlock this prevents)
+            // Path steering leads - it carries its own whiskers as a bias on
+            // its aim point; this one is the no-route fallback, where a
+            // heading override is safe (see chase_behavior for the deadlock
+            // that overriding an ACTIVE route causes)
             Box::new(PathFollowSteeringStrategy::to_point(goal)),
             Box::new(CollisionAvoidanceSteeringStrategy::conservative()),
         ])
@@ -163,6 +165,22 @@ impl PatrolBehavior {
                 Effect::combine(effects)
             }
         }
+    }
+
+    /// Give up on the current point and try the next one. Enough of those in
+    /// a row - without reaching one in between - and the whole route is out
+    /// of reach, so stop patrolling and hand back to idle rather than
+    /// shuffling (and re-querying A*) between points forever.
+    fn skip_point(&mut self, world: &World, entity_id: EntityId) -> Effect {
+        self.skipped_points += 1;
+        if self.skipped_points >= PATROL_MAX_SKIPPED_POINTS {
+            self.finished = true;
+            return Effect::SetAICurrentPatrol {
+                entity_id,
+                target: None,
+            };
+        }
+        self.advance(world, entity_id, false)
     }
 
     /// Whether the AI has failed to cover ground for PATROL_STALL_SECONDS.
@@ -217,26 +235,19 @@ impl Behavior for PatrolBehavior {
                 // the floor below). Move on at once rather than steering at it
                 // on a heading nothing checked, grinding into the geometry in
                 // between until the stall watchdog retires the whole route.
+                // Counted like a stalled skip, so a route with no reachable
+                // point left ends instead of cycling (and re-querying) forever.
                 tracing::debug!(
                     "patrol {entity_id:?}: no route to point {:?}, skipping to the next one",
                     self.target_point
                 );
-                patrol_effects.push(self.advance(world, entity_id, false));
+                patrol_effects.push(self.skip_point(world, entity_id));
             } else if self.stalled(position, time) {
                 // Going nowhere: give up on this point and try the next one.
                 // Enough of those in a row and the whole route is out of
                 // reach, so stop patrolling and hand back to idle via
                 // next_behavior rather than walking in place forever.
-                self.skipped_points += 1;
-                if self.skipped_points >= PATROL_MAX_SKIPPED_POINTS {
-                    self.finished = true;
-                    patrol_effects.push(Effect::SetAICurrentPatrol {
-                        entity_id,
-                        target: None,
-                    });
-                } else {
-                    patrol_effects.push(self.advance(world, entity_id, false));
-                }
+                patrol_effects.push(self.skip_point(world, entity_id));
             }
         }
 
@@ -415,6 +426,9 @@ mod tests {
 
     /// Steering that reports no route to whatever goal it was given.
     struct UnreachableSteering;
+    /// ...and its counterpart, which has a route.
+    struct ReachableSteering;
+    impl SteeringStrategy for ReachableSteering {}
     impl SteeringStrategy for UnreachableSteering {
         fn goal_unreachable(&self) -> bool {
             true
@@ -439,18 +453,9 @@ mod tests {
             }
         }
 
-        assert!(
-            !patrol.finished,
-            "skipping unreachable points must keep the route running"
-        );
-        assert!(
-            !emitted
-                .iter()
-                .any(|effect| matches!(effect, Effect::SetAICurrentPatrol { target: None, .. })),
-            "the patrol must not retire itself over unreachable points"
-        );
-        // The three-point loop keeps advancing: every tick targets a
-        // different point than the one before it.
+        // The route moves on at once - one tick per point, no waiting out the
+        // stall watchdog - and, since NO point on this loop is reachable, it
+        // ends at the skip cap rather than cycling (and re-querying) forever.
         let targets: Vec<EntityId> = emitted
             .iter()
             .filter_map(|effect| match effect {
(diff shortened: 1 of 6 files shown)


tommy-xr:
## Full e2e

Ran the full SDK e2e suite on this worktree (chunked, 11 `not ok`). Triaged the ones not already known-pre-existing (creature loot 12451 disc, Earth hackable keypad, hydro3 Korenchkin log), by re-running each isolated test file on this branch and, where plausibly steering-related, against `main`.

| Test | Verdict | Evidence |
|---|---|---|
| use mode exposes AMMOFULL ammo-cycle for an empty multi-ammo weapon | pre-existing | Fails identically on branch and on `main` (`loaded magazine must not expose a conversion control`) |
| MedSci saved mission Wrench keeps canonical 9-damage VR melee across decks and saves | pre-existing | Fails identically on branch and on `main` (`fresh Wrench should emit one Damage message`, 0 !== 1) |
| the medsci2 balcony patroller never wedges on the railing | branch (known, tracked) | This branch's new test. 2/4 reruns passed, 2/4 wedged at ~(49.0, 0.10, -96.0) on the same railing spot, matching the PR body's documented 4/8-ish flake rate. Stacked stall-reroute PR is intended to close this. |
| eng1: a locked grav-lift button only works once its unlock trap fires | pre-existing | Fails identically on branch and on `main` (`a locked button should refuse`, false !== true) |
| VR authored melee damages only during a trigger-gated physical contact window | pre-existing | Fails identically on branch and on `main` (`idle held-Wrench contact must remain harmless`); no creature/AI involved (breakable pane) |
| hydro2: Floor Pod gibs launch, settle, expire, and stay non-solid to characters | flake | Passed clean on rerun (`gib 1529 should settle and sleep`) |
| the flat Developer screen launches a debug scene and can back out of the list | pre-existing | Fails identically on branch and on `main`; wrong scene launched (`cs1.avi` instead of `debug_minimal`) |

Also found (not in the original triage list, same root cause as the row above): **"the VR Developer screen launches a debug scene through the controller ray"** also fails identically on branch and `main` (same `cs1.avi` vs `debug_minimal` mismatch) — pre-existing.

Net: no new failures attributable to this branch beyond the balcony test it itself added, which is the known/tracked flake. Not marking ready — that's for the stall-reroute follow-up to close out.

tommy-xr:
Ready as the parent of #1264. Note: `medsci2-patrol-railing.e2e` is 1-2/4 on this PR alone (the body is pinned by a physics recovery bug, not steering) and 4/4 with #1264 stacked on top — merge the pair together.

tommy-xr:
Rebased the whole stack (#1248 → #1264 → #1271 → #1313 → #1318 → #1337) onto current `main`, which now contains #1245's agent-radius clearance — the stack previously branched from a pre-merge `main`, so every earlier validation ran without it.

This branch: `ed4e174e` → `fbc92a3c`, no conflicts, all 7 commits carried over unchanged (`git range-diff`).

Stack-tip validation (`feat/ai-turn-clips`): `cargo fmt --check` clean; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo test -p shock2vr -p dark` 1417/149/11 pass, 0 fail; `cargo bn path bench medsci2.mis --seed 424242` unchanged vs `main` (200/0 found, inflate 1.67, turn 944°, 45.8 waypoints).

PR #1272, opened 2026-09-03
feat(debug): debug_psi pen and full stats

tommy-xr:
## What

`debug_psi` gets an enemy pen and a full character sheet so psi powers can be measured against a live creature:

- Two pipe hybrids (-397, the actor `debug_melee`/`debug_hitbox` use) stand in a closed, knee-high pen 9 and 11 units ahead. The 1.2 wall was measured, not guessed: an alerted hybrid piles up at x≈-8.2 and never crosses it, while a level cast from eye height (~2.3) clears it onto the near creature's torso. The far one is offset so the near one doesn't screen it.
- The psi pool starts full (50/50 instead of the template's 40) and every stat, skill and psi tier is at cap via the same path as `POST /v1/player/stats`. Nothing on the cast path reads the sheet yet (debug scenes already know every power; the amp's effective PSI is a constant), so that half is parity with `debug_weapons`, stated as such in the code.

The box geometry idiom, `cube_object`, and the max-stats block were each on their third copy across debug scenes; they move to `debug_common` (`boxes_to_geometry`, `cube_object`, `max_player_stats`) and `debug_melee`/`debug_weapons` use them. Net −60 lines.

Groundwork for a per-power psi audit (next PRs: registry dump over HTTP, an SDK audit script, e2e lock-in of the implemented powers).

## Verification

- `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p debug_runtime` clean.
- Runtime: two straight cryo casts take the near hybrid 12→10 HP and psi 50→48; damage-alerted hybrids stay penned after 15 s.
- e2e: every suite that launches `debug_psi`, `debug_melee` or `debug_weapons` (29 files) — see the results comment below.
- `/xreview` (Claude reviewers; Codex was out of quota): the duplication and target-placement findings are addressed in this commit.

## Media

debug_psi pen demo

<sub>Two Projected Cryokinesis casts into the pen from the player's eye. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

impact still

### DIFF
--- a/shock2vr/src/scenes/debug_common.rs
+++ b/shock2vr/src/scenes/debug_common.rs
@@ -15,12 +15,12 @@ use engine::{
         light::SpotLight,
     },
 };
-use rapier3d::prelude::{Collider, ColliderBuilder};
+use rapier3d::prelude::{Collider, ColliderBuilder, Isometry, SharedShape};
 use shipyard::EntityId;
 
 use crate::{
     GameOptions,
-    game_scene::GameScene,
+    game_scene::{DebugPlayerStatsRequest, DebugSkillLevelsRequest, DebuggableScene, GameScene},
     input_context::InputContext,
     mission::{
         AbstractMission, AlwaysVisible, GlobalContext, SpawnLocation,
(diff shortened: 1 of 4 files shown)


tommy-xr:
e2e: all 29 suites that launch `debug_psi`/`debug_melee`/`debug_weapons` — 58/58 pass on c96b717e (`SHOCK2_E2E=1 node --test` over the listed files).

PR #1320, opened 2026-09-03
feat(traits): Speedy, Lethal Weapon, Power Psi, Strong Metabolism, Spatially Aware

tommy-xr:
Five O/S upgrades had no consumer, so the trait machine refused them outright rather than waste its one-shot pick. Each now has one, and Pharmo-Friendly's existing 20% item bonus extends to the psi hypo.

| id | Trait | Effect | Call site |
| --- | --- | --- | --- |
| 4 | Speedy | stick locomotion x1.15 | `mission/mission_core.rs` - `PLAYER_MOVE_SPEED` at the locomotion vector |
| 9 | Lethal Weapon | player melee damage x1.35 | `scripts/weapon_script.rs` `flat_melee_hit` + `scripts/melee_weapon.rs` `authored_contact_damage` (held weapon only, so a loose one a creature blunders into is unaffected) |
| 14 | Power Psi | psi burnout damage -> 0 | `scripts/psi_amp_script.rs` `burnout` (the cast still fails and the points are still spent) |
| 1 | Strong Metabolism | radiation damage x0.75 | `scripts/radiation.rs` `ActiveRadiation::advance` / `resist` |
| 16 | Spatially Aware | automap fully revealed | `mission_core.rs` at load, and on acquire via `apply_os_trait_world_grants` |
| 2 | Pharmo-Friendly | psi hypo 20 -> 24 | `scripts/psi_kit.rs`, through the same truncating x6/5 `pharmo_friendly_amount` the med items use |

Each effect is a pure helper beside the trait ids in `scripts/gui/traits.rs` (`speedy_move_speed`, `lethal_weapon_damage`, `strong_metabolism_damage`, `power_psi_burnout_damage`, `pharmo_friendly_amount`), read at the call site through the character sheet, so it re-derives after save/load and a level change the way the existing Tank and Pack-Rat bonuses do. The GUI stops refusing these six ids; 5, 7, 10, 11, 12 and 15 still refuse.

**Radiation resistance banks its fraction.** The shipped sources plateau where a 6-second tick bills one hit point, so multiplying each pulse by 0.75 and truncating would turn the trait into either a no-op or total immunity depending on the level. `ActiveRadiation` carries the sub-point remainder instead, so a quarter off is a whole point saved every fourth pulse. Unresisted behavior is byte-identical (the scale is 1.0 and the remainder never accrues).

**Toxin caveat:** Strong Metabolism resists toxins as well in the original. This port has no toxin system, so only the radiation half exists to scale.

**Spatially Aware in VR:** the reveal is recorded on the character sheet in both presentations, but this port's automap panel is flat-only, so in VR the upgrade has nothing to draw until a VR map exists. It is not refused - the data is right and a save moved to flat shows it.

Two supporting bits: Tank's acquire-time hit-point grant moved into a shared `apply_os_trait_world_grants` (so the trait machine and the debug provisioning path can't drift), and `POST /v1/player/stats` accepts `os_traits` so automation can grant traits without a one-shot machine. `/v1/info` now also reports `map_location_count`.

## Tests

Unit, negative-first per trait:

- `scripts/gui/traits.rs`: `the_round_one_traits_are_selectable_now_that_they_have_consumers`, `traits_without_an_effect_are_still_refused` (drives the panel and asserts the refusal), `trait_multipliers_are_identity_without_the_trait`
- `scripts/radiation.rs`: `strong_metabolism_cuts_a_quarter_off_the_damage_actually_dealt` (four 1-point pulses cost 4 unresisted, 3 resisted)
- `scripts/psi_kit.rs`: `pharmo_friendly_boosts_the_hypo_by_a_fifth`
- `scripts/melee_weapon.rs`: `lethal_weapon_scales_a_held_swing_only`

`cargo fmt`; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`; `cargo test -p shock2vr --lib` -> 1267 passed, 0 failed.

New e2e `tools/shock2-sdk/test/os-upgrades.e2e.test.ts` grants traits through `POST /v1/player/stats` (`os_traits`) and measures the untraited behavior first in the same run:

```
ok 1 - Speedy: stick locomotion covers 15% more ground
ok 2 - Lethal Weapon: a flat wrench swing hits 35% harder
ok 3 - Power Psi removes burnout damage; Pharmo-Friendly boosts the psi hypo
ok 4 - Spatially Aware reveals the whole automap, on acquire and on every load
# pass 4
# fail 0
```

Strong Metabolism has no e2e case: the only reachable shipped radiation source (medsci2's Rad Barrel) plateaus the stored level near 6.5, where the difference is one hit point over the better part of a minute, under the noise of the barrel's own blast. The arithmetic is unit-tested instead.

`os-traits.e2e.test.ts` now clicks Sharpshooter (still stubbed) for its "upgrade unavailable" case, since Speedy is purchasable now.

Regression sweep over the touched systems - `os-traits`, `cyber-modules`, `save-load`, `melee-hitbox`, `vr-melee-contact`, `flat-melee-animation-event`, `psi`, `psi-overload`, `psi-sustained`, `psi-trained`, `rad-patch`, `map`, `healing-items`, `inventory-strength`, `provisioning`, `trainer`, `missions`, plus the new file: **56 pass, 0 fail**. (`melee-hitbox`'s limb case, listed as flaky, passed here.)

### DIFF
--- a/runtimes/debug_runtime/src/commands.rs
+++ b/runtimes/debug_runtime/src/commands.rs
@@ -904,6 +904,10 @@ pub struct PlayerInfo {
     /// Climb state: whether the player is on a ladder/hand hold at all, and
     /// (VR) which hands hold what. See `shock2vr::vr_climb`.
     pub climb: shock2vr::game_scene::DebugClimbState,
+    /// How many automap locations the current mission defines (locations run
+    /// `0..count`), so a caller can tell a partly-explored map from a fully
+    /// revealed one. 0 when the mission has no automap.
+    pub map_location_count: usize,
 }
 
 /// Input state snapshot
--- a/runtimes/debug_runtime/src/main.rs
+++ b/runtimes/debug_runtime/src/main.rs
@@ -2424,6 +2424,10 @@ fn capture_frame_snapshot(
                     .debug_scene()
                     .and_then(|scene| scene.player_climb())
                     .unwrap_or_default(),
+                map_location_count: state
+                    .as_ref()
+                    .map(|s| s.map_location_count)
+                    .unwrap_or_default(),
             }
         },
         entity_count,
--- a/shock2vr/src/game_scene.rs
+++ b/shock2vr/src/game_scene.rs
@@ -565,6 +565,9 @@ pub struct DebugPlayerStatsRequest {
     pub psi_tier: Option<i32>,
     /// Target cyber-module balance (the upgrade currency).
     pub cyber_modules: Option<i32>,
+    /// O/S upgrade trait ids to grant (1..=16), as if picked at a machine.
+    /// Already-owned ids are ignored; the slot cap still applies.
+    pub os_traits: Option<Vec<u8>>,
 }
 
 /// Target skill levels, mirroring `player.stats.skills`. Named fields (rather
--- a/shock2vr/src/lib.rs
+++ b/shock2vr/src/lib.rs
@@ -721,6 +721,10 @@ pub struct PlayerStateSnapshot {
     /// Empty when the scene has no `QuestInfo` or no automap. Persisted per
     /// mission in `QuestInfo`. See `projects/flat-ui-panels.md` §5.
     pub explored_map_locations: Vec<i32>,
+    /// How many automap locations the current mission defines (locations run
+    /// `0..count`), so a caller can tell a partly-explored map from a fully
+    /// revealed one. 0 when the scene has no automap.
+    pub map_location_count: usize,
 }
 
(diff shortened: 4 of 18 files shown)


tommy-xr:
Replaced by #1619 #1626 #1618 #1624

tommy-xr:
Superseded: each trait landed separately on main — Speedy (#1619), Lethal Weapon (#1626), Power Psi (#1618), Spatially Aware (#1624); Strong Metabolism is already implemented on main (`radiation.rs`/`traits.rs`).

PR #1325, opened 2026-09-04
feat(traits): Sharpshooter scales the player's ranged damage

tommy-xr:
**Sharpshooter (O/S trait 5): the player's ranged weapon damage x1.15.**

Every shot fired from a gun in the player's own hands lands 15% harder - bullets
and energy shots alike, since they all leave through the same gun fire path.
Melee is Lethal Weapon's domain and is untouched.

Two pre-existing damage-model gaps, both unchanged here and both noted at
`shooter_damage_scale`:

- A projectile's *radius blast* (a grenade, the fusion cannon) is billed by
  `internal_explosion` straight from the authored StimSource intensity and
  honours no shot modifiers at all - so today it ignores the fire setting's own
  `stim_modifier` exactly as it ignores Sharpshooter.
- The slow (physics) projectile impact path still bills a placeholder damage of
  `1.0 * stim` (`internal_collision_type.rs`, its own TODO), so 1.15 rounds back
  to 1 there: the bonus is only *observable* on the fast (raycast) path until
  that damage is resolved from authored data.

Both belong with #1251's modifier / the damage TODO rather than with this trait.

### One damage-scale path

The scale rides the projectile's fire-time `RuntimePropShotModifiers.stim` (#1251)
rather than being re-applied at the impact. The projectile already carries one
damage multiplier that the two impact paths honour, so folding Sharpshooter into
it at fire time bills the upgrade exactly once, and no impact code has to know who
fired.

`shot_modifiers` now takes the shooter's own damage scale and multiplies the fire
setting's authored modifier by it, so the EMP rifle's 3x overcharge becomes 3.45x
rather than being replaced.

Only a gun `held_in_hand` by the player earns it: AI and turret shots are created
with `CreateEntityOptions::default()` and never come down this path at all, and a
loose gun going off is nobody's aim.

`sharp_mult` is not read from the gamesys - `dark/` does not parse the TRAITPARAM
chunk, so the value is hardcoded 1.15 the way #1320 hardcodes Lethal Weapon's
1.35.

Fidelity note: 15% is the figure the trait's own description states, and it is
what this port implements - retail play measures a larger bonus than its own
description promises, and that divergence is deliberate rather than matched.

Trait 5 is un-refused in `traits.rs`, so the trait machine vends it.

### Tests

Unit (`cargo test -p shock2vr --lib`, 1268 passed / 0 failed):

- `trait_multipliers_are_identity_without_the_trait` - `sharpshooter_damage_scale`
  is 1.0 without the trait, 1.15 with it.
- `only_the_players_own_gun_earns_sharpshooter` - held + trait is 1.15; held
  without it, and traited-but-unheld, are both 1.0.
- `the_shooters_damage_scale_compounds_with_the_settings` - 3.0 stim modifier x
  1.15 = 3.45, launch speed untouched.
- `the_round_one_traits_are_selectable_now_that_they_have_consumers` /
  `traits_without_an_effect_are_still_refused` - 5 moves from the refused list to
  the selectable one.

e2e (`os-upgrades.e2e.test.ts`, new case "Sharpshooter: a pistol shot lands 15%
harder"): on earth.mis, a spawned Pistol is aimed at the authored Training Droid's
torso and one round fired, measured on the droid's own hit points. Untraited the
shot costs it 6; with the trait granted through the debug provisioning path it
costs 7 (6 * 1.15 = 6.9 on an integer hit-point pool).

Watched fail first: with the scale forced to 1.0 the case fails
`Sharpshooter must raise ranged damage; 6 -> 6`.

The turret/AI non-scaling half is covered by the unit test rather than a
`debug_turret` e2e: a turret's hit rate over a fixed window is noisy enough that a
15% damage assertion would be flaky, and the AI fire path provably never builds a
`RuntimePropShotModifiers` at all.

`os-traits.e2e.test.ts` used Sharpshooter as its "not yet implemented" example;
that moves to Cybernetically Enhanced (7), as do the equivalent unit tests.

Verified: `cargo fmt`; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p
desktop_runtime -p debug_runtime`; `cargo test -p shock2vr --lib`; e2e
`os-upgrades`, `os-traits`, `weapon-fire-mode-stats`, `weapon-fire-mode`,
`weapon-burst-fire`, `blood-spang`, `ranged-hitbox`, `energy-weapon-recharge`
(18/18 after the os-traits update), and `save-load` + `missions` (27/27).

### DIFF
--- a/shock2vr/src/scripts/gui/traits.rs
+++ b/shock2vr/src/scripts/gui/traits.rs
@@ -52,6 +52,7 @@ pub const TRAIT_STRONG_METABOLISM: u8 = 1;
 pub const TRAIT_PHARMO_FRIENDLY: u8 = 2;
 pub const TRAIT_PACK_RAT: u8 = 3;
 pub const TRAIT_SPEEDY: u8 = 4;
+pub const TRAIT_SHARPSHOOTER: u8 = 5;
 pub const TRAIT_NATURALLY_ABLE: u8 = 6;
 pub const TRAIT_TANK: u8 = 8;
 pub const TRAIT_LETHAL_WEAPON: u8 = 9;
@@ -77,6 +78,11 @@ pub const SPEEDY_MOVE_SCALE: f32 = 1.15;
 /// melee only - AI melee resolves through its own attack path.
 pub const LETHAL_WEAPON_DAMAGE_SCALE: f32 = 1.35;
 
+/// Sharpshooter: ranged weapons hit 15% harder (Trait5). The player's own
+/// shots only. 15% is the figure the trait's own description states; retail
+/// play measures a larger bonus, and this port implements the stated 15%.
+pub const SHARPSHOOTER_DAMAGE_SCALE: f32 = 1.15;
+
 /// Strong Metabolism: radiation damage is cut by a quarter (Trait1). The
 /// retail trait also resists toxins; this port has no toxin system yet, so
 /// only the radiation half has anything to scale.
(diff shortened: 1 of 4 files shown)


tommy-xr:
Replaced with #1625

tommy-xr:
Superseded by #1625 (classic Sharpshooter ranged damage).

PR #1332, opened 2026-09-04
feat(vr): an off-panel trigger pull throws the cursor item into the world

tommy-xr:
The cyber interface's cursor-is-the-item drag could lift, place, swap and wield in VR, but had no way to commit an item to the world: flat's bare-view click had no VR reading (cyber-interface plan, open thread "drop-into-world from the panel").

**Now:** with an item on the cursor, a trigger pull whose controller ray is **off the panel** throws it along that hand's ray (the pointer pass's own ray, mapped through the pawn with `hand_world_position`). Empty panel pixels still keep the item, as before. An off-panel pull with an empty cursor stays that hand's ordinary world press.

- `FlatUiDragAction::Throw` carries the throwing hand; `FrontendPointerPass` publishes `claiming_hand` (held trigger wins, ties fall to the right hand) so the off-panel hand is decided once, in the pass.
- While an item rides the cursor, both triggers are masked from the world in VR (`trigger_safe_mask`), so the throw cannot double as a frob. The mask reads the cursor one frame late; the latch covers both edges of the pull.
- Flat-facing side effect: the off-canvas throw now also clears the double-click mark, like the in-canvas throw already did (a click on the thrown item's old slot within 20 frames could previously read as a wield).

**Known limits:** the pass reports one aggregate `pressed` edge, so a fresh off-panel pull while the *other* trigger is already held does not throw (nothing misfires; release and pull again). The spawn point is `capsule radius + 0.2` along the hand ray, sized for the flat eye ray; a controller held against the body and aimed back through it can spawn the item inside the capsule and get it ejected.

**Tests:** host unit tests (off-panel throw by hand, off-panel pointer hand, empty-cursor no-op), `trigger_safe_mask` with a loaded cursor, and a VR e2e that throws once with each hand in opposite directions and asserts the launch velocities are anti-parallel (a throw aimed by the wrong hand or the head would fly the same way twice).

VR panel throw

<sub>Lift the wrench onto the cursor, swing the ray below the panel, pull: the icon leaves the cursor and the wrench lands on the floor ahead (its name tag flashes as the world hand crosses it). Brightened 1.8x - the comfort dim makes the world dark. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

| lifted | thrown |
| --- | --- |
| lifted | thrown |

### DIFF
--- a/shock2vr/src/mission/flat_ui_host.rs
+++ b/shock2vr/src/mission/flat_ui_host.rs
@@ -179,20 +179,12 @@ pub fn vr_canvas_pointer(
     pass: &crate::ui::FrontendPointerPass,
     input_context: &crate::input_context::InputContext,
 ) -> CanvasPointer {
-    // Off the panel there is no owning ray, so the hand is whichever one is
-    // pulling the trigger - the throw of a cursor item aims along its ray.
-    let held =
-        |hand: &crate::input_context::Hand| hand.trigger_value > crate::ui::VR_TRIGGER_THRESHOLD;
+    // Off the panel no ray owns the point, so the hand is the one the pass
+    // says is claiming the panel - the throw of a cursor item aims along it.
     let hand = pass
         .active_ray()
         .map(|ray| ray.handedness)
-        .unwrap_or_else(|| {
-            if held(&input_context.left_hand) && !held(&input_context.right_hand) {
-                Handedness::Left
-            } else {
-                Handedness::Right
-            }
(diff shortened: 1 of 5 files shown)


tommy-xr:
i don't think this makes sense - dragging an item holds it, and triggering while held should _use_ the item - we can have an actual 'throw' throw an item

PR #1337, opened 2026-09-05
feat(ai): play the authored turn clip when pivoting

tommy-xr:
A pivot past a right angle already stops the body (#1313). It now performs the creature's own authored turn clip instead of sliding the heading around underneath a run cycle.

The clips were there all along, filed beside the idle stand clip — human `humstand`/`humtul`/`humtur`/`humtu180`, hybrid `ogptu90l`/`ogptu90r`/`ogstu180` — and the only thing that tells a turn apart from an idle is the facing change it is authored to end on. So a pivot picks the clip whose authored angle is nearest the heading it needs and that it can afford to stand still for; plain standing picks from the clips that don't turn at all.

**The pose does the turning.** Measured across the stock clips, the hip joint's net yaw over a turn clip *equals* that clip's signed `end_direction`: `humtur` +102.58° authored, +102.3° in the pose; `ogptu90l` +91.83° / +91.8°. The body is turned by the animation, so the entity must hold still underneath it — rotating as well would turn the creature twice — and take the authored facing change afterwards, spread over the blend that swings the pose back to neutral. That is what keeps the visible facing continuous across the handoff.

The AI script stays pure: it emits `Effect::PlayTurnClip { delta, max_seconds }`, the effect applier resolves the creature's schema and reports back with `MessagePayload::TurnClipStarted { turn, duration, blend }`, and the pivot runs as a small `Requested → Playing → Settling` state machine. While it runs, the behavior is **not steered at all** — a body that is deliberately standing still would otherwise read as no progress, and the path follower would blacklist the cell it is turning on while the patrol route gave up on the point it is turning toward.

Two numbers, both load-bearing:

- `TURN_CLIP_MAX_SECONDS = 4.0` — the stock turn clips run 2.4–5.6 s. A pivot long enough to read as a wedge (the reachability harness calls holding one spot for 6 s stuck, and it is right to) is worse than steering the turn, so clips longer than this are simply not picked. `humtu180` (4.8 s) and `ogptu90l` (5.6 s) don't qualify; the rest do.
- `TURN_CLIP_COOLDOWN = 3.0` — two pivots back to back hold the creature still for longer than either of them, which is exactly what a wedge looks like.

Ledger step 11 (`~/notes/projects/shock2quest/pathfinding-eval-plan.md`), and it closes out the "turn clips deferred" note on #1313: the `+facing` branch really is empty, but the turn clips are reachable under `+stand` with the creature's own actor tags.

**Bonus bug, same PR:** because the turn clips sit in the `+stand` schema, plain idle selection was picking them at random — idlers swung their bodies through a 100° turn with no heading change behind it. A `stand` query now excludes anything that turns.

## Before → after

before/after

turn clip demo

<sub>After: a chasing hybrid reverses in place through `ogstu180` — its yaw is held at 207.3° for the whole clip (frames 1→53) and then takes the authored −180° across the blend. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep), free camera locked to the creature.</sub>

before

<sub>Before, on `feat/ai-door-wait`: the same scenario (medsci2, `DebugForceChase`, same capture script) — the reversal is 95° of yaw in 0.7 s with the run cycle still playing. Clip selection is random per query, so the two runs are the same request sequence rather than the same creature at the same spot.</sub>

## Tests

- Unit, `dark`: `an_about_face_covers_a_pivot_across_the_half_turn`, `a_turn_the_creature_cannot_afford_is_not_picked`, `a_pivot_takes_the_clip_nearest_its_heading_change`, `a_small_pivot_has_no_clip_worth_playing`, `only_the_pivots_count_as_turn_clips` (the idle-selection bug), `signed_end_direction`.
- Unit, `shock2vr`: `a_reversal_asks_for_an_authored_turn_clip`, `a_shallow_pivot_is_steered_as_before`, `the_body_holds_its_heading_and_stride_while_the_clip_plays`, `the_authored_turn_lands_on_the_entity_once_the_clip_ends`, `a_later_clip_completing_does_not_restart_the_blend`, `a_completion_that_arrives_early_is_not_the_turn_clip`, `a_second_pivot_waits_for_the_cooldown`, `a_pivot_cannot_hold_long_enough_to_look_stuck`, `an_unanswered_pivot_falls_back_to_steering`, `a_stalled_creature_steers_its_turn_instead_of_performing_it`, `a_creature_is_measured_in_the_same_units_as_the_fallback`.
- `cargo test -p shock2vr`: 1421 passed.
- e2e on the fix: `medsci2-doorjamb-chase` 5/5 (five separate runs), `patrol` 4/4, `force-chase` 1/1, `medsci2-patrol-railing` 2/2, `missions` 23/23.
- e2e re-run after the review restructure, reduced to the pivot-heavy scenarios: `medsci2-doorjamb-chase` 3/3, `medsci2-patrol-railing` 1/1.
- Pathfinding is untouched (no file under `shock2vr/src/pathfinding/` or `dark/src/mission/` is in the diff), so `cargo bn path bench` is unaffected by construction.

## Two regressions found and fixed while building this

Both were caught by the e2e set, not by unit tests, and both have a test now:

1. **The pivot never ended.** The behavior's own clip is re-queued the moment a turn clip completes, and a run cycle is a third of a second long — so its completion landed inside the settle and restarted it. Patrollers turned on the spot forever and never took another step (`patrol` e2e: `traveled 0.0`). Fixed by the explicit state machine: only a `Playing` pivot settles.
2. **A long pivot looked like a wedge.** `ogptu90l` runs 5.6 s; held one spot for 6–11 s, the railing e2e failed. Fixed by the affordability budget plus the cooldown above.

## The pile-up e2e failure, and two more fixes

`medsci2-doorjamb-chase`'s pile-up test was failing, and a per-frame trace found two causes - neither of them the pivot state machine. It was already failing on this stack's **parent**, so it is not purely a turn-clip regression.

- **A pivot performed while scrambling out of a wedge.** The path follower's stall recovery re-paths into a sidestep, that reads as a >90 degree pivot, and the creature stands through a 2.7 s turn clip plus its blend before taking it - turning a recovery into a longer wedge. A stall now bars pivoting, for long enough to cover the recovery's re-path (which lands after the stall clock has already been reset).
- **`creature_height` divided by `SCALE_FACTOR` twice** (pre-existing, and the dominant cause). The authored bounding box is already in world units - as its doc comment, the no-definition fallback and the physics capsule all say - so a hybrid measured 2.6 feet instead of 6.5. `door_is_passable` then called a leaf risen 2.7 feet clear, and the AI walked under a still-rising door that carried it off the floor and dropped it. The same bug had collapsed the whisker knee and chest probes to nearly one height.

Six runs per variant through a harness mirroring the test's exact request sequence:

| build | pile-up test |
| --- | --- |
| parent `ab8a51de` (pre-turn-clip) | 5/6 fail |
| `feat/ai-turn-clips` before these fixes | 4/6 fail |
| + pivot barred after a stall | 2/6 fail |
| + creature height in world units | 0/6 fail |

0/6 again after the review restructure.

## Review

`/xreview` — Claude adversarial reviewer plus the dedicated de-dup reviewer. **The Codex engine is unavailable until Sep 6, so this ran single-engine**; nothing here carries cross-engine agreement.

Fixed:

- **High** — `nearest_turn_clip` compared an *unwrapped* residual, so an about-face authored just past 180° (midwife `humuntu180` = 182.97° → −177.03°) was rejected for a pivot just short of it (+175°) as if it were 352° away. Wrapped, with a test.
- **High** — any completion settled the pivot, including one belonging to a clip that had *preempted* the turn (a wound reaction, an alertness change). The AI would then swing through the full authored turn with a neutral pose. Only a completion arriving at the end of the turn clip's own runtime settles now; anything earlier abandons the pivot.
- **Medium** — the frustration gesture could fire mid-pivot (it was suppressed only on the frame the pivot was requested), preempting the clip and reporting frustration at a standstill the AI chose.
- **Medium** — a single unanswered request latched the creature out of pivoting for the rest of its life. Removed; the cooldown spaces out the asking instead, and a pivot with no affordable clip is no longer reported as a failed animation.
- **Medium-Low** — the settle window was hardcoded at 500 ms, but the monkey turns author a 100 ms blend. It is now the clip's own `blend_length`.
- **Low** — the wedge-budget assertion didn't model the overrun margin.

Noted, not changed:

- **`mission_core`'s `DirectionChanged` applier is wrong** — it applies `end_rotation * -0.5`, i.e. half the *unsigned* on-disk heading, negated. It has no effect on any AI creature (the script re-asserts `SetRotation` every frame, after `update_animations`), which is presumably why it survived; this PR relies on that. Auditing what it does to non-AI entities, and making it apply the signed authored angle, is a follow-up.
- De-dup reviewer suggested merging `signed_end_direction` with `ai_util::clamp_to_minimal_delta_angle`. Declined: they live in different crates (`dark` cannot call `shock2vr`), and they disagree at exactly ±180°, so the merge is a behavior change to twenty-odd existing call sites for no gain here.
- Two turn clips per human schema now go unused (`humtu180`, `ogptu90l`) because they exceed the standstill budget. Playing a clip's turn portion only, or time-scaling it, is the obvious follow-up if reversals should always be performed.

### DIFF
--- a/dark/src/motion/mod.rs
+++ b/dark/src/motion/mod.rs
@@ -5,6 +5,7 @@ mod motion_clip;
 mod motion_info;
 pub mod motion_query;
 mod motion_schema;
+mod turn_clips;
 
 pub use animation_clip::*;
 pub use animation_player::*;
@@ -14,6 +15,7 @@ pub use motion_info::*;
 pub use motion_query::*;
 pub use motion_schema::*;
 use rand::{Rng, thread_rng};
+pub use turn_clips::*;
 
 use crate::{
     NameMap, SCALE_FACTOR, TagDatabase, ss2_chunk_file_reader,
--- a/dark/src/motion/turn_clips.rs
+++ b/dark/src/motion/turn_clips.rs
@@ -30,15 +30,22 @@ pub fn is_turn_clip(end_direction: Deg<f32>) -> bool {
 }
 
 /// Index of the turn clip that best covers `delta` degrees of heading change,
-/// or `None` when standing still leaves the creature closer to `delta` than
-/// any authored turn would (small pivots have no clip: the shortest stock
-/// human turn is ~102 degrees).
-pub fn nearest_turn_clip(delta: Deg<f32>, end_directions: &[Deg<f32>]) -> Option<usize> {
-    end_directions
+/// given each clip's authored facing change and how long it runs.
+///
+/// `None` when no clip fits: a turn longer than `max_duration` keeps the
+/// creature standing still for longer than it can afford, and standing still
+/// leaves it closer to `delta` than a badly-matched turn would (small pivots
+/// have no clip at all - the shortest stock human turn is ~102 degrees).
+pub fn nearest_turn_clip(
+    delta: Deg<f32>,
+    clips: &[(Deg<f32>, f32)],
+    max_duration: f32,
+) -> Option<usize> {
+    clips
         .iter()
         .enumerate()
-        .filter(|(_, end)| is_turn_clip(**end))
-        .map(|(index, end)| (index, (signed_end_direction(*end).0 - delta.0).abs()))
+        .filter(|(_, (end, duration))| is_turn_clip(*end) && *duration <= max_duration)
+        .map(|(index, (end, _))| (index, (signed_end_direction(*end).0 - delta.0).abs()))
         .filter(|(_, residual)| *residual < delta.0.abs())
         .min_by(|a, b| a.1.total_cmp(&b.1))
         .map(|(index, _)| index)
@@ -48,8 +55,15 @@ pub fn nearest_turn_clip(delta: Deg<f32>, end_directions: &[Deg<f32>]) -> Option
 mod tests {
     use super::*;
 
-    /// The stock human `+stand` schema: idle, left, right, about-face.
-    const HUMAN_STAND: [Deg<f32>; 4] = [Deg(0.0), Deg(258.15), Deg(102.58), Deg(191.95)];
+    /// The stock human `+stand` schema: idle, left, right, about-face, with
+    /// the seconds each one runs for.
+    const HUMAN_STAND: [(Deg<f32>, f32); 4] = [
+        (Deg(0.0), 5.03),
+        (Deg(258.15), 3.60),
+        (Deg(102.58), 3.83),
+        (Deg(191.95), 4.80),
+    ];
+    const NO_BUDGET_LIMIT: f32 = 1000.0;
 
     #[test]
     fn an_unsigned_heading_reads_as_the_minimal_signed_turn() {
@@ -70,20 +84,46 @@ mod tests {
 
     #[test]
     fn a_pivot_takes_the_clip_nearest_its_heading_change() {
-        assert_eq!(nearest_turn_clip(Deg(95.0), &HUMAN_STAND), Some(2));
-        assert_eq!(nearest_turn_clip(Deg(-95.0), &HUMAN_STAND), Some(1));
-        assert_eq!(nearest_turn_clip(Deg(-170.0), &HUMAN_STAND), Some(3));
+        assert_eq!(
+            nearest_turn_clip(Deg(95.0), &HUMAN_STAND, NO_BUDGET_LIMIT),
+            Some(2)
+        );
+        assert_eq!(
+            nearest_turn_clip(Deg(-95.0), &HUMAN_STAND, NO_BUDGET_LIMIT),
+            Some(1)
+        );
+        assert_eq!(
+            nearest_turn_clip(Deg(-170.0), &HUMAN_STAND, NO_BUDGET_LIMIT),
+            Some(3)
+        );
     }
 
     #[test]
     fn a_small_pivot_has_no_clip_worth_playing() {
-        assert_eq!(nearest_turn_clip(Deg(30.0), &HUMAN_STAND), None);
-        assert_eq!(nearest_turn_clip(Deg(-30.0), &HUMAN_STAND), None);
+        assert_eq!(
+            nearest_turn_clip(Deg(30.0), &HUMAN_STAND, NO_BUDGET_LIMIT),
+            None
+        );
+        assert_eq!(
+            nearest_turn_clip(Deg(-30.0), &HUMAN_STAND, NO_BUDGET_LIMIT),
+            None
+        );
     }
 
     #[test]
     fn a_schema_without_turn_clips_never_pivots() {
-        assert_eq!(nearest_turn_clip(Deg(150.0), &[Deg(0.0)]), None);
-        assert_eq!(nearest_turn_clip(Deg(150.0), &[]), None);
+        assert_eq!(
+            nearest_turn_clip(Deg(150.0), &[(Deg(0.0), 1.0)], NO_BUDGET_LIMIT),
+            None
+        );
+        assert_eq!(nearest_turn_clip(Deg(150.0), &[], NO_BUDGET_LIMIT), None);
+    }
+
+    /// A creature can only stand still for so long: an about-face it cannot
+    /// afford is steered rather than performed.
+    #[test]
+    fn a_turn_the_creature_cannot_afford_is_not_picked() {
+        assert_eq!(nearest_turn_clip(Deg(-170.0), &HUMAN_STAND, 4.0), Some(1));
+        assert_eq!(nearest_turn_clip(Deg(-170.0), &HUMAN_STAND, 3.0), None);
     }
 }
--- a/dark/src/motion/turn_clips.rs
+++ b/dark/src/motion/turn_clips.rs
@@ -1,0 +1,89 @@
+//! Picking the authored turn clip for a pivot.
+//!
+//! Creature schemas file their turn-in-place motions beside the idle stand
+//! clip (human: humstand, humtul, humtur, humtu180); the only thing that
+//! distinguishes them is the authored facing change each one ends on. So a
+//! pivot picks the clip whose authored angle is nearest the heading it needs,
+//! and plain standing picks from the clips that don't turn at all.
+
+use cgmath::Deg;
+
+/// Below this, a clip's authored facing change is incidental drift rather
+/// than a turn (stock idle clips end a fraction of a degree off).
+const TURN_CLIP_MIN_ANGLE: f32 = 20.0;
+
+/// A clip's authored facing change as a signed minimal angle, in (-180, 180].
+/// On disk it is an unsigned 0..360 heading.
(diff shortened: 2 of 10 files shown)


tommy-xr:
Stack rebased onto current `main` (now contains #1245's agent-radius clearance; the stack previously branched from a pre-merge `main`, so all earlier validation ran without it). This branch: `09f65486` → `59572bc1`, no conflicts, all 7 commits carried over unchanged (`git range-diff`).

Conflicts elsewhere in the stack, both additive and resolved keeping both sides:
- `shock2vr/src/pathfinding/mod.rs` — #1245's `pinch_and_detour_db` helper + clearance tests concatenated with the stack's `detour_db` / stall-blacklist tests.
- `tools/shock2-sdk/test/medsci2-doorjamb-chase.e2e.test.ts` — the stack's generalised pile-up test kept, #1245's jamb-specific test appended as a second test in the file.
- `shock2vr/src/scripts/script_util.rs` — test import list, union of both sides.

Validation at this tip: `cargo fmt --check` clean; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo test -p shock2vr -p dark` 1417/149/11 pass, 0 fail; `cargo bn path bench medsci2.mis --seed 424242` unchanged vs `main` (200/0, inflate 1.67, turn 944°, 45.8 waypoints). E2E: `medsci2-patrol-railing` 2/2, `patrol` 4/4, `force-chase` 1/1, `door-frob` 1/1, `missions` 23/23.

**One regression to chase before merge**: `medsci2-doorjamb-chase`'s pile-up test ("chasing medsci2 hybrids never grind in one spot") now fails **4 of 5 runs** at this tip — one hybrid holds for exactly 7 s (limit 6 s) around `x ≈ 39.7, z ≈ -44…-46`, a different spot from the jamb. The jamb-specific test passes every run. Bisecting the rebased stack: `fix/ai-stall-reroute` and `feat/ai-door-wait` tips pass, and the *pre-rebase* `feat/ai-turn-clips` tip passed 3/3 — so it is this layer's turn clips interacting with the new base. The hold is always exactly 7 s and never longer, which smells like the bounded pivot budget (4 s clip + 3 s cooldown) crossing the test's 6 s window rather than a true wedge, but it needs confirming (and either a fix or a justified threshold change) before this lands.

tommy-xr:
## `medsci2-doorjamb-chase` pile-up failure: two causes, neither of them the pivot state machine

Traced per-frame with the debug runtime (`DebugForceChase` on medsci2, one line per frame per locomoting AI: turn state, clip, heading, desired heading, stall clock, position).

**It is not a chained pivot, a Settling that never ends, or a lost `end_direction`.** The 7 s is one hold made of two things:

```
t=7.2   hybrid reaches 38.50,-36.93 - delta ~0, locomotion scale 1.0, heading not held.
        It is asking to walk forward and the body does not move: physically wedged.
        It is even carried up to y=1.54 and dropped back.
t=10.5  the path stall clock passes 2.5; at 3.0 the follower recovers and re-paths.
t=11.67 the recovery's new leg points 94 deg right -> that reads as a pivot, and the
        creature stands through ogptu90r (2.7 s) + settle (0.5 s).
t=14.9  steered again, want=100.3 (the same escape heading), walks free immediately.
```

### 1. A pivot performed while scrambling out of a wedge (this PR's layer)

The turn clip is doing exactly what it was built to do - the escape heading is real, and the pivot ends with the creature free. But spending 3.2 s standing still is the wrong answer for a creature that has just failed to make progress: it turns a recovery into a longer wedge. A stall now bars pivoting, for long enough to cover the recovery's re-path (which lands *after* the stall clock has been reset). Test: `a_stalled_creature_steers_its_turn_instead_of_performing_it`.

### 2. `creature_height` divided by `SCALE_FACTOR` twice (pre-existing, the dominant cause)

The wedge in the trace above is a *door* carrying the AI off the floor. `ai_util::creature_height` divided the authored bounding box by `SCALE_FACTOR` a second time - the box is already in world units, as its own doc comment, the no-definition fallback (`6.5 / SCALE_FACTOR`) and the physics capsule all say. A hybrid therefore measured **2.6 feet instead of 6.5**, so `door_is_passable`'s `rise >= actor_height` called a leaf that had risen 2.7 feet clear, the AI walked under a still-rising door, rode it up and got dropped. (Same bug collapsed the whisker knee and chest probes to nearly one height.) Test: `a_creature_is_measured_in_the_same_units_as_the_fallback`.

### Numbers

The pile-up test was failing on this stack's **parent** too, so it is not purely a turn-clip regression. Six runs per variant through an identical harness that mirrors the test's exact request sequence:

| build | pile-up test |
| --- | --- |
| parent `ab8a51de` (pre-turn-clip) | **5/6 fail** |
| `feat/ai-turn-clips` before these fixes | **4/6 fail** |
| \+ pivot barred after a stall | **2/6 fail** |
| \+ creature height in world units | **0/6 fail** |

Real e2e after both: `medsci2-doorjamb-chase` **5/5**, `medsci2-patrol-railing` 2/2, `patrol` 4/4, `force-chase` 1/1, `missions` 23/23, `cargo test -p shock2vr` 1419 passed. `cargo fmt` and `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean.

tommy-xr:
Follow-up: `/xreview` on the two fixes above. Codex found nothing actionable; the Claude reviewer found one real hole and the de-dup reviewer one real redundancy, both now fixed in `a824d720`.

- **High, stale stall reading.** `PathFollowSteeringStrategy` publishes its stall clock only while it steers, and `record_ai_steering` is never aged - so the last value it published outlives it. A creature that stalled and then went to melee or idle would have re-armed the bar off a frozen reading. The bar now reads the clock only under `may_turn`, and `is_locomotion` is exactly the set of behaviors that follow a route, so the value it sees was published by that same frame's steer.
- **Redundancy.** The bar was a second countdown beside `turn_cooldown` with the same length and the same meaning ("seconds before this AI may pivot"). It arms that cooldown instead; the field, the constant and the plumbed-through gate are gone. `path_stall_seconds` is also read once per frame now rather than twice.
- Two unguarded invariants the review named are pinned: the bar outlasts the follower's recovery, and a hybrid's whiskers probe its knee and chest (the height bug had bunched both rays at chest level).

Re-validated at reduced scope, on the pivot-heavy scenarios only: `cargo test -p shock2vr` 1421 passed, `medsci2-doorjamb-chase` 3/3, `medsci2-patrol-railing` 1/1, plus the 6-run harness above 0/6 fail. `cargo fmt` and `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean. (`patrol`, `force-chase` and `missions` were run green on the previous commit and not repeated.)

PR #1353, opened 2026-09-05
fix(ai): a patrol given up under a temporary exclusion resumes when it expires

tommy-xr:
A patrolling AI gives up on a route point when the path query says the goal has no route, and once it has come back around to a point it already gave up on - the whole lap tried, nothing reached - the patrol retires and hands back to Idle for the rest of the mission.

That verdict does not distinguish *why* the query failed. The stall system blacklists crossings an AI physically could not traverse (`report_blocked_link`, 30 s TTL) and penalizes cells an AI is wedged in, so a route can be unroutable purely because something is in the way *right now*. A patrol unlucky enough to run its lap inside that window retired permanently over a blockage that cleared 30 seconds later.

So a query that fell short is asked *why*: `PathfindingService::exclusion_expiry_for_unreached_goal` re-runs the same search on the raw mesh, with the exclusions lifted. Only if the goal is reachable that way was the shortfall caused by them, and the deadline is then the expiry of the excluded crossings **on that recovered route** - not the level's blacklist at large, so a stall on the other side of the deck is not mistaken for this patrol's problem. A goal unreachable on the raw mesh too - a point on a disconnected part of the navigation graph - reports no deadline and retires exactly as before.

The worker attaches that deadline to any query that fell short, and `PathFollowSteeringStrategy` exposes it beside `goal_unreachable` as `goal_unreachable_until`. A patrol whose route ran out with a deadline in hand waits it out and runs the route again instead of retiring.

The re-run costs one more A* on a query that already failed, and only while some crossing is excluded at all (the common case skips it outright). The deadline is an upper bound: it measures the cheapest raw route, so a pricier route that reopens sooner is not noticed.

Stalled *cells* are deliberately not part of the signal: they are a cost penalty, not an exclusion, so they can never be why a goal came back unreachable.

The wait is bounded (`PATROL_TEMPORARY_RETRIES`, 3 waits with no point reached in between), so an AI on a deck where something is always blocked somewhere eventually hands back to idle rather than standing there for the mission. Reaching a point resets the count, as it already does the give-up list.

While waiting the AI stands - no steering at a goal nothing can route to - and animates idle rather than locomotion. Alertness still replaces the behavior normally, so a waiting patroller is not deaf to the player.

Ledger item P1 (`~/notes/projects/shock2quest/pathfinding-eval-plan.md`).

## Tests

- Pathfinding, negative-first (both fail against the previous level-wide signal): `an_unreachable_goal_is_not_made_temporary_by_an_unrelated_exclusion` (a goal on a linkless island, with an unrelated crossing excluded elsewhere: was `Some(30)`, now `None`) and `a_goal_cut_off_only_by_exclusions_reports_its_own_route_s_deadline` (two exclusions on the route it reopens by, a third later one on the detour it will not take, plus a stalled cell: `Some(32)` - so a level-wide `max` (35) and a level-wide `min` (30) both fail).
- Propagation with a **real worker** (not a stub): `worker_reports_a_deadline_only_for_an_exclusion_caused_shortfall` submits two queries through `AsyncPathfinding` against one service and asserts `PathQueryResponse::exclusion_expires_at` is `None` for the island goal and the route's own deadline for the cut-off one. The last hop - worker response to `goal_unreachable_until` - stays covered by the steering unit tests; driving a real AI into a real stall from outside the game is not deterministic (see below).
- Retry policy, unchanged from the first round: `a_patrol_given_up_under_a_temporary_exclusion_resumes_when_it_expires`, `an_always_blocked_route_stops_waiting_and_retires`, `a_topologically_unreachable_route_still_retires`, `a_stalled_cell_alone_is_never_something_to_wait_for`.
- `cargo test -p shock2vr`: 1428 passed.
- e2e: `patrol` 4/4 (again after the fix). Earlier rounds: `medsci2-patrol-railing` 1/1 (x2), `force-chase` 1/1, `missions` 23/23.
- `cargo bn path bench medsci2.mis --seed 424242` unchanged: 200/0 found, inflation 1.67, turn 944 deg.

**No e2e for the fix itself.** Nothing outside the game can put a crossing on the blacklist - exclusions are only ever produced by an AI actually stalling mid-route, and there is no debug endpoint for `report_blocked_link`. A test that waits for the right AI to wedge itself in the right place would not be deterministic, so the coverage here is the unit + worker tests plus the existing patrol e2e as a no-regression check.

No media: nothing visible changes except that an AI which used to stand idle forever eventually walks again.

## Review

`/xreview` - all three reviewers ran (Claude adversarial, Codex, and the de-dup pass). Both engines independently raised the same central finding. Fixed:

- **High [AGREED]** - the temporary/topological signal was the *whole level's* blacklist, so any creature stalling anywhere made an unrelated patrol's genuinely disconnected route look temporary, and a busy deck could keep it retrying indefinitely. Now it reads blocked *crossings* only (cells cannot cause unreachability), it is computed only for a query that fell short, and the wait is capped at 3 - so the retry is bounded by construction rather than by hoping the blacklist empties.
- **Medium** - the stall give-up inherited a `temporary_until` left behind by an earlier point's routing verdict, so a lap closed by a physical wedge waited on a routing exclusion. The verdict is now passed per give-up (`None` from the stall branch), which also deleted the accumulator field and its two clear sites.
- **Medium (Codex)** - an open patrol chain that simply ran out of points never reached the pause check and retired regardless. Running out of chain is now treated as the same empty route as a completed lap.
- **Medium (Codex)** - a deadline that had already lapsed by the time the response was adopted fell through to retirement; it now pauses and resumes on the next frame.
- **Low** - `ChainedSteeringStrategy::goal_unreachable_until` took the first `Some` regardless of which strategy actually reported the goal unreachable; it now takes the latest deadline among the strategies that did.

Noted, not changed: the de-dup reviewer asked whether the existing `REPATH_FAILURE_BACKOFF_SECONDS` re-query already covers this. It does not - it makes steering ask again, but the patrol has already retired by then; what is missing is precisely the *verdict* (is waiting worth anything at all), which the backoff cannot express. Otherwise: no actionable duplication.

### Second round (re-review of the claim)

The temporary/topological distinction was still only *bounded*, not *made*: the deadline was the latest excluded crossing anywhere in the level, so an unrelated stall made a disconnected patrol point look temporary and the 3-wait cap merely limited the damage. Fixed by proving the verdict against the raw mesh (above), with the deadline scoped to the recovered route. A second `/xreview` over that commit:

- **Medium [AGREED]** - the exclusion set the search used and the expiry map were read at two different moments, so a crossing reported in between could invent or extend a deadline (or a pruned one erase it). The deadline is now scoped to the query's own avoidance snapshot, with a crossing since pruned counting as already lapsed.
- **Medium** - neither test distinguished "max over the recovered route" from "min over the level" (the numbers coincided). Both now carry two exclusions with different expiries on the recovered route.
- **Low, both engines** - the stressed second pass was copy-pasted into the re-run, and `blocked_links` repeated the retain-under-lock. Both searches now share `cell_path_avoiding`, and `blocked_links` derives from the expiry map.
- Noted, not changed: the re-run costs an extra failing A* on the worker in exactly the disconnected case, which a per-component index could answer in O(1) - not worth a bits-aware component index for a search that runs only when something is excluded and only off the frame thread. The extra pass counts in `PathfindingStats` (it is a real search).

### DIFF
--- a/shock2vr/src/pathfinding/async_queries.rs
+++ b/shock2vr/src/pathfinding/async_queries.rs
@@ -44,9 +44,10 @@ pub struct PathQueryResponse {
     pub outcome: AiPathOutcome,
     /// Route waypoints; empty when `outcome` is `Failed`
     pub waypoints: Vec<Vector3<f32>>,
-    /// When the query ran against live steering-reported exclusions, the
-    /// mission time the last of them expires at. A failure under one may be
-    /// temporary - the consumer can retry rather than write the goal off.
+    /// For a query that did not reach its goal, the mission time the last
+    /// excluded crossing comes back into play at (if any is live). The
+    /// failure may be temporary - the consumer can retry rather than write
+    /// the goal off.
     pub exclusion_expires_at: Option<f32>,
 }
 
@@ -84,7 +85,6 @@ impl AsyncPathfinding {
                     // including fresh arrivals' - route around the obstacle
                     // instead of piling into it
                     let avoid = service.avoidance(request.now_seconds);
-                    let exclusion_expires_at = avoid.expires_at;
                     let path = service
                         .find_path_avoiding(
                             request.start,
@@ -108,6 +108,10 @@ impl AsyncPathfinding {
                             Vec::new()
                         }
                     };
+                    // Only a query that fell short can be waiting on one
+                    let exclusion_expires_at = (outcome != AiPathOutcome::Full)
+                        .then(|| service.blocked_link_expiry(request.now_seconds))
+                        .flatten();
                     service.record_ai_path(
                         request.entity,
                         AiPathRecord {
--- a/shock2vr/src/pathfinding/async_queries.rs
+++ b/shock2vr/src/pathfinding/async_queries.rs
@@ -118,6 +118,7 @@ impl AsyncPathfinding {
                                 request.start,
                                 request.goal,
                                 request.movement_bits,
+                                &avoid,
                                 request.now_seconds,
                             )
                         })
@@ -306,8 +307,8 @@ mod tests {
         let service = Arc::new(PathfindingService::new(Arc::new(
             crate::pathfinding::tests::island_db(),
         )));
-        service.report_blocked_link(0, 1, 0.0);
         service.report_blocked_link(1, 2, 0.0);
+        service.report_blocked_link(0, 1, 2.0);
         service.report_blocked_link(3, 2, 5.0);
         let async_pf = AsyncPathfinding::spawn(service.clone());
 
@@ -336,9 +337,9 @@ mod tests {
         assert_ne!(response.outcome, AiPathOutcome::Full);
         assert_eq!(
             response.exclusion_expires_at,
-            Some(crate::pathfinding::BLOCKED_LINK_TTL_SECONDS),
-            "wait for the crossings on the route it becomes reachable by, not \
-             for the level's latest blockage (5 s later, on the detour)"
+            Some(2.0 + crate::pathfinding::BLOCKED_LINK_TTL_SECONDS),
+            "wait for the LAST crossing on the route it becomes reachable by, \
+             neither the level's earliest nor its latest (the detour's)"
         );
     }
 
--- a/shock2vr/src/pathfinding/async_queries.rs
+++ b/shock2vr/src/pathfinding/async_queries.rs
@@ -44,6 +44,10 @@ pub struct PathQueryResponse {
     pub outcome: AiPathOutcome,
     /// Route waypoints; empty when `outcome` is `Failed`
     pub waypoints: Vec<Vector3<f32>>,
+    /// When the query ran against live steering-reported exclusions, the
+    /// mission time the last of them expires at. A failure under one may be
+    /// temporary - the consumer can retry rather than write the goal off.
+    pub exclusion_expires_at: Option<f32>,
 }
 
 struct SharedState {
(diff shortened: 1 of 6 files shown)


tommy-xr:
Linearized the phase-2 chain on top of #1337's current tip (`a824d720`), so there is one chain rather than two siblings off #1354:

`feat/ai-turn-clips` → #1353 → #1354 → #1355 → #1357 → #1356 → #1362

This PR was branched from the older #1337 tip and now sits on `a824d720` (no conflicts; both commits replay identically).

The one real conflict was in #1355, where `a824d720`'s `bar_pivot_while_stalled` used `FRUSTRATION_STALL_SECONDS` and `path_stall_seconds`, both deleted there. The bar is kept — a stall still arms the pivot cooldown, and its test with it — but folded into the hold/cooldown design: it now reads the follower's published stall clock and trips on the follower's own `STALL_SECONDS` mark instead of the deleted frustration constant.

Verified at the top of the chain: `cargo fmt --check`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`, `cargo test -p shock2vr -p dark` (1447 + 151 + 11 pass), and e2e `medsci2-doorjamb-chase` (x2) + `medsci2-patrol-railing` all green.

tommy-xr:
Re-review finding addressed: the temporary verdict is now proved against the raw mesh (search re-run with exclusions lifted) and the deadline is scoped to the excluded crossings on the recovered route, so an unrelated stall elsewhere no longer makes a disconnected patrol point look temporary. Adds a real-worker propagation test through `AsyncPathfinding`. `patrol` e2e 4/4, `cargo bn path bench medsci2.mis --seed 424242` unchanged (200/0, 1.67, 944).

PR #1354, opened 2026-09-05
fix(ai): a head-on neighbour makes the AI sidestep instead of pushing through

tommy-xr:
Crowd steering had a blind spot exactly where it exists: a neighbour **squarely ahead**. The push is away from the neighbour, so head-on it points straight back down the route — and `aim_with_bias` drops whatever part of a bias points backwards, because turning to a point behind you is not steering. The whole push was therefore discarded, and two AIs walking into each other pushed capsule-to-capsule until one of them wedged.

Inside a 30 degree cone of anti-parallel the crowd bias is now rotated onto the perpendicular at the same strength: the AI **steps around** instead of leaning back. The side is always the same one relative to travel, which is what lets two AIs pass without coordinating — facing opposite ways, their own sides are opposite sides of the corridor. The conversion eases in across the cone (reusing `repel_ramp`) rather than switching at its edge, and fades with the melee approach like the repel it mostly comes from, so an AI in reach of its target is never stepped off the blow.

Two smaller pieces of the same finding:

- **Geometry outranks the crowd — on both axes that matter.** The whisker bias and the crowd bias were summed, so two opposite pushes of similar size cancelled to nothing — no avoidance *and* no yielding. Whatever part of a crowd push opposes the whiskers is now dropped, and the whiskers also pick which way the AI yields: if its own side is a wall, the other one is open by construction, and both AIs of a pair read the same wall the same way. That rule works along the whiskers' own direction, which turned out not to be enough (see **Wall priority in the space that survives**, below).
- **Separation and repel were already one response** — one neighbour walk, one summed bias in `crowd_bias` — so there was nothing to merge there.

Phase 2 item P4 (`~/notes/projects/shock2quest/pathfinding-eval-plan.md`).

## Wall priority in the space that survives

Dropping the crowd's opposing part fixed the *collinear* case and missed the perpendicular one. `aim_with_bias` keeps only what points across the direction of travel, so a bias's lateral component is the only part that reaches the aim point at all — and two vectors can be perpendicular to each other (so neither "opposes" the other, and the rule above passes them both through) while their *lateral* parts are exactly opposite.

Travelling `+z`, a diagonal wall hit `(1,0,-1)` and a neighbour outside the head-on cone at `(-1,0,-1)`: the sum is `(0,0,-2)`, purely backward, and the projection erases it whole. The AI got no wall avoidance at all — precisely when a neighbour was also present, which is when it needs it.

The same priority is now applied a second time along the lateral axis, where it survives. It is the same `drop_opposing` call, aimed along the whiskers' lateral direction instead of the whiskers themselves, so the two rules are one idea applied on two axes rather than two mechanisms. The head-on cone logic and the deterministic side choice are untouched; the perpendicular that `yield_sideways` yields onto and the one the priority is applied along are now literally the same function (`ai_util::lateral_axis`) — the rule is only correct because both read the same handedness.

Unit, negative-first on the full composition (`yield_sideways` → `blend_biases` → `aim_with_bias`) with exactly those vectors:

- `a_wall_survives_the_projection_when_a_crowd_is_present` — pre-fix the aim point came back unbent (`(0,0,10)`, the raw waypoint); it now carries the wall's `+x` sidestep. The test also asserts the neighbour really is outside the cone, so the conversion is not what is doing the work.
- `a_crowd_agreeing_with_the_wall_still_adds` — same-side pushes sum, and the sum is still shortened once to the budget.
- `a_crowd_with_no_whiskers_is_untouched` — a crowd on its own behaves exactly as before.

`medsci2-patrol-railing` (the railing *is* the wall-vs-waypoint case) passes, before and after the review fixes.

## Before / after

The pile-up e2e (`medsci2-doorjamb-chase`, "chasing medsci2 hybrids never grind in one spot" — every AI in the level chases at once, so the pile-up is the thing under test) is nondeterministic, so it is a rate:

| branch | passes |
|---|---|
| base `fix/ai-patrol-temporary-unreachable` | **0 / 6** |
| this branch | **2 / 6** |

It never passed on the base. The second assertion in that file (a chasing hybrid takes the open door, not the sliver beside its jamb) is 6/6 on both.

## Media

Two hybrids spawned facing each other in the medsci2 science corridor (x 38, z -26 / -34), then dragged up and down that corridor by teleporting the player between its ends on a fixed schedule, so the pair repeatedly has to swap places in a one-body-wide passage. Identical request sequence on both branches, captured headlessly through the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed timestep, free camera).

head-on sidestep, before vs after

<sub>Left: on the base, once the pair has to cross, the trailing hybrid is left behind down the corridor and never regains contact. Right: with the change it steps around and the two travel abreast.</sub>

same frame, before and after

Gap between the two hybrids over the run, three runs per branch:

pair separation over time

The two branches agree for the first half and separate once the pair has to cross: mean gap over the last 20 samples is 9.4 / 10.5 / 10.9 on the base and 3.7 / 4.7 / 5.9 on this branch (max gap 11.7-12.5 vs 7.2-8.7) - consistent across three runs each, but note this is one scenario, not a deterministic isolation of the fix (see the pass-rate table above).

## Tests

Unit, on the pure vector functions:

- `a_head_on_push_becomes_a_sidestep` — negative-first: a neighbour dead ahead used to yield exactly zero lateral.
- `two_ai_meeting_head_on_yield_to_opposite_sides` — including headings five degrees off anti-parallel that both lean the same way in world space (the review counterexample below); this fails against the first implementation.
- `a_wall_flips_the_side_the_ai_yields_to`, `a_push_off_the_head_on_cone_is_untouched`, `yielding_eases_in_across_the_cone`, `a_wall_outranks_a_crowd_push_into_it`.

`cargo test -p shock2vr`: 1434 passed. `cargo fmt`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean.

**Reduced validation scope (deliberate).** e2e was cut to the pile-up test above — there is no repel/separation e2e to run. `patrol` 4/4, `force-chase` 1/1, `missions` 23/23 and the medsci2 reachability harness (3 runs per branch, `--pass both`) were run on the first commit of this branch before the review fixes: no regression, and on a single mission the harness buckets are too noisy to show this (the one consistent move was idle `no_route` 1/1/1 -> 0/0/0). They were not re-run after the review fixes.

## Review

`/xreview` — both engines, plus the de-dup reviewer. The first implementation did not survive it; three findings, two raised independently by both engines:

- **High [AGREED]** — the side was the one the push already leaned toward, which is only mirrored between a pair when their headings are exactly anti-parallel. Two AIs converging a few degrees off dead-centre (both aiming at the same off-centre doorway) leaned the same way in world space and stayed nose to nose — the very case the change exists to break. The side is now always the same one relative to travel, which *is* opposite between two bodies facing opposite ways whatever their lean, and the counterexample is a regression test.
- **High** — yielding each neighbour's share separately let a neighbour dead ahead and one abreast rotate onto the same axis and cancel, leaving an AI with two neighbours worse off than before the change. The yield now applies once to the whole crowd bias at the call site — which also removed the `crowd_bias` signature change and every test call-site edit with it, for a materially smaller diff.
- **High** — a sidestep aimed at a door jamb was then deleted outright by the new whisker projection, so the fix was a no-op in a doorway, where it is needed most. The whiskers now choose the side rather than erasing it.
- **Medium** — the cone edge was a hard switch that stepped the aim point by roughly half the offset budget in one frame (the same defect the melee fade in this file already documents). Ramped.
- **Medium** — the sidestep was not melee-faded, so an AI closing on the player with an ally ahead got a new lateral push in exactly the situation the fade protects. It now fades with `strength`.
- **De-dup** — three hand-rolled copies of "remove the part of this vector pointing against that direction" (the new yield, the new whisker projection, and the pre-existing `aim_with_bias`); one `drop_opposing` helper replaces all three.

Not taken: gating the yield on the neighbour actually being *oncoming* (needs neighbour velocity or heading, and the melee fade covers the case that motivated it), and hoisting `normalized_horizontal` into `shock2vr::util` to share with the UI and whisker modules (a different subsystem, out of scope here).

Re-reviewed after the lateral-axis fix — both engines plus the de-dup reviewer, on that commit alone:

- **Codex** — no findings at any severity; it independently confirmed the earlier `drop_opposing(crowd, whiskers)` is *not* redundant given the new rule (heading `+z`, crowd `(0,0,1)`, whiskers `(1,0,-1)` is a case only the old rule catches), so both are kept.
- **De-dup `reuse`** — the lateral projection was hand-rolled in twelve lines that are exactly `drop_opposing` along the lateral axis, degenerate cases included. Collapsed to the one call.
- **Medium, raised by both the correctness and de-dup reviewers** — the `(heading.z, 0, -heading.x)` swizzle was written twice, in two modules, and the priority is only correct because both copies agree on handedness. Extracted as `ai_util::lateral_axis`.
- **Low** — `waypoint - position` was recomputed for the yield and again for the blend; hoisted, so the two cannot drift.
- **Not taken:** grading the drop by the whisker's magnitude, on the argument that a far-edge whisker hit should not veto a saturated crowd push. The premise does not hold — `drop_opposing` normalizes its direction, so it has always removed the *whole* opposing projection whatever the whiskers' length, and `a_wall_outranks_a_crowd_push_into_it` asserts that outright veto deliberately. Making it proportional is a change to the established geometry-first rule, not a fix to this one.

Noted for later, out of scope here: a whisker pointing *straight* back down the route (lateral zero, a wall square across the path) is still erased by `aim_with_bias` — the same class of loss, but it predates this branch.

### DIFF
--- a/shock2vr/src/scripts/ai/ai_util.rs
+++ b/shock2vr/src/scripts/ai/ai_util.rs
@@ -734,7 +734,20 @@ pub(crate) fn normalized_horizontal(v: Vector3<f32>) -> Option<Vector3<f32>> {
     (length > 1e-6).then(|| vec3(v.x / length, 0.0, v.z / length))
 }
 
-/// Turn a head-on push into a sidestep of the same strength.
+/// Drop whatever part of `v` points against `direction`, keeping the rest.
+/// Used both to keep a bias from pulling a body backwards down its route
+/// and to stop a crowd push from bidding against the whiskers.
+pub(crate) fn drop_opposing(v: Vector3<f32>, direction: Vector3<f32>) -> Vector3<f32> {
+    match normalized_horizontal(direction) {
+        Some(direction) => {
+            let opposing = (v.x * direction.x + v.z * direction.z).min(0.0);
+            v - direction * opposing
+        }
+        None => v,
+    }
+}
+
+/// Turn a head-on crowd push into a sidestep of the same strength.
 ///
 /// A neighbour squarely ahead pushes straight back down the route, and a
 /// bias pointing backwards is worth nothing: the aim point keeps only its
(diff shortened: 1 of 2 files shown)


tommy-xr:
Re-review follow-up: the geometry-over-crowd rule was applied between the raw vectors, but only the lateral part survives `aim_with_bias`'s forward projection — so a diagonal wall hit and a neighbour off to one side could be perpendicular (neither opposing the other) while their lateral parts cancelled, and the purely-backward sum was erased whole. The wall avoidance was lost exactly when a neighbour was present.

The same `drop_opposing` rule is now applied a second time along the lateral axis, where it survives. Head-on cone and deterministic side choice unchanged; the perpendicular is now shared with `yield_sideways` (`ai_util::lateral_axis`) since the rule depends on both reading the same handedness.

Negative-first on the full composition with exactly those vectors — pre-fix the aim point came back unbent at `(0,0,10)`. `cargo test -p shock2vr` 1434 pass, `-D warnings` check clean, `medsci2-patrol-railing` passes before and after the review fixes.

PR #1355, opened 2026-09-05
fix(ai): one movement-hold reason; the impatience gesture only at a door

tommy-xr:
An AI that stops **on purpose** — standing off from a door leaf still crossing the doorway (#1318), performing an authored turn clip (#1337) — was being handled twice by accident, and neither way was right.

- The door wait ran **after** the steer that reads the stall clock, so the seconds of a hold were already in the clock the wait was supposed to stay clear of. `DOOR_WAIT_TIMEOUT 2.5 < STALL_SECONDS 3.0` proves nothing when the counter it is compared against already contains time.
- A pivot avoided the same clock by **skipping the whole behavior update** — route, whiskers, crowd repel and all — for the length of the clip, just to stop progress accounting.

Now there is one explicit reason. The script publishes a `MovementHold` (`DoorWait`, `Pivot`, or `None`) **before** it steers, and the watchdogs answer by freezing their no-progress clocks while it is up. Nothing else pauses: the behavior is steered right through both holds, so heading, whiskers, crowd repel and the route stay live while the creature stands. A hold *pauses* the clock rather than resetting it — seconds a real wedge had already banked are still there when the hold lifts.

Two watchdogs read it: the path follower's waypoint-progress and displacement clocks (`advance_stall_clocks`), and `PatrolBehavior`'s own 5 s stall clock — which matters precisely because the behavior is no longer frozen: a 4 s turn clip would otherwise spend most of that window and retire the point the AI was turning toward.

**The impatience gesture is now a door gesture only** (user decision, 2026-09-05). It fits inside a wait the AI is already taking; on a stall it added seconds of standing still to a creature trying to get out of a wedge — measured on #1318's own door as ~3.3 s → ~6.6 s, which is a real responsiveness regression whatever it does for eventual reachability. The locked-door give-up keeps its performance, but now goes through the same rate limit, melee gate and scripted-performance gate as the wait (it used to play regardless of what either path had spent), and no longer fires over a pivot in flight.

The hold reason is readable over `GET /v1/ai/paths` (`movement_hold`), so a creature standing at a door is distinguishable from a wedged one — including by the reachability harness, whose `wedged` bucket is exactly this confusion.

Phase-2 item P2 (`~/notes/projects/shock2quest/pathfinding-eval-plan.md`).

## Tests

Negative-first — each new test was confirmed to fail on the pre-change behavior (hold ignored / pivot skipping the steer / give-up gesture unlimited):

- `a_door_wait_never_advances_the_stall_clock`, `a_pivot_never_advances_the_stall_clock`, `a_hold_leaves_the_stall_it_paused_where_it_found_it` (path follower).
- `a_held_patroller_never_stalls` (patrol watchdog).
- `a_pivot_no_longer_freezes_the_behavior` — a patroller standing on its point keeps advancing its route while a 10 s pivot plays.
- `a_door_wait_outranks_a_pivot_as_the_hold_reason`, `a_sustained_stall_draws_no_gesture` (with the door case beside it, so the gesture path is proved still live), `the_locked_door_give_up_respects_the_shared_limit` — which also covers the pivot gate: a turn clip in flight refuses the gesture and leaves the rate limit unarmed.
- `cargo test -p shock2vr`: 1437 passed.
- e2e (scoped to the changed behavior): `medsci2-doorjamb-chase` 2/2 twice; after the review fixes `medsci2-doorjamb-chase` 2/2 twice more, `patrol` 4/4, `medsci2-patrol-railing` 1/1.
- medsci1 `DebugForceChase` headless run: 17 door waits, every one resolved on leaf clearance in 0.02–1.45 s, none reaching `DOOR_WAIT_TIMEOUT` — #1318's hold semantics are intact.

**Evidence correction.** `a_sustained_stall_draws_no_gesture` was not negative-first as first written: it built a world with no pathfinding service, so the “stalled” AI read zero stall seconds and the pre-change code would have kept silent for it too. It now publishes 10 s of no-progress on the channel the path follower writes, and asserts the stall is on the wire before asserting on the gesture — restoring the old stall branch makes it fail. The other claims were re-audited against the pre-change code and stand: the two follower clock tests carry a `MovementHold::None` control that does stall, the patrol one likewise, and `a_pivot_no_longer_freezes_the_behavior` fails on code that skipped the behavior update while a clip played.

## Review

`/xreview` (Claude + Codex). Codex: no findings. The de-dup pass and the Claude reviewer independently raised the same **High**: the patrol watchdog is a second no-progress clock the hold could not reach — fixed here, with a regression test. Also fixed from the review: the give-up gesture no longer stomps a pivot in flight; `update_frustration`'s tail folded into the one `try_frustration_gesture`; a comment that described an anchor the code doesn't take.

Noted, not changed: each AI now takes one extra mutex lock + map insert per frame to publish its hold (negligible on desktop; worth remembering for Quest), and a pivot's leading/trailing edge is published one 16 ms frame late.

## Restack note

This branch is on #1354, which is on #1353 — branched before `feat/ai-turn-clips`' final commit `a824d720`. That commit's `bar_pivot_while_stalled` uses `FRUSTRATION_STALL_SECONDS` and `path_stall_seconds`, both **deleted here** (dead once the stall gesture is gone). Whoever restacks #1353 onto that tip needs to bring them back for it — the failure is a loud compile error, not a silent one.

### DIFF
--- a/shock2vr/src/game_scene.rs
+++ b/shock2vr/src/game_scene.rs
@@ -1091,6 +1091,10 @@ pub struct DebugAiPathEntry {
     pub live_target: Option<[f32; 3]>,
     /// Seconds without progress toward the current waypoint
     pub live_stall_seconds: Option<f32>,
+    /// Why the AI is deliberately standing still, if it is ("DoorWait",
+    /// "Pivot") - a held AI stands on purpose and is not a wedged one, and
+    /// its stall clock is frozen for as long as this is set
+    pub movement_hold: Option<String>,
 }
 
 /// Snapshot of the pathfinding service's monotonic query counters.
--- a/shock2vr/src/mission/mission_core.rs
+++ b/shock2vr/src/mission/mission_core.rs
@@ -12290,6 +12290,10 @@ impl crate::game_scene::DebuggableScene for MissionCore {
                     live_path_len: live.map(|l| l.path_len),
                     live_target: live.and_then(|l| l.target).map(Into::into),
                     live_stall_seconds: live.map(|l| l.stall_seconds),
+                    movement_hold: match service.movement_hold(entity) {
+                        crate::pathfinding::MovementHold::None => None,
+                        hold => Some(format!("{hold:?}")),
+                    },
                 }
             })
             .collect()
--- a/shock2vr/src/pathfinding/mod.rs
+++ b/shock2vr/src/pathfinding/mod.rs
@@ -201,6 +201,29 @@ pub struct AiSteeringDebug {
     pub stall_seconds: f32,
 }
 
+/// Why an AI is deliberately standing still this frame. A hold is a decision,
+/// not a failure: the script publishes it BEFORE it steers, and the path
+/// follower answers by suspending its no-progress accounting, so seconds spent
+/// standing on purpose never spend the patience it keeps for a real wedge.
+/// Steering itself keeps running through a hold - heading, whiskers and crowd
+/// repel stay live.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum MovementHold {
+    /// Free to move
+    #[default]
+    None,
+    /// Standing off while a door leaf this AI opened travels clear
+    DoorWait,
+    /// Performing an authored turn clip - the pose does the turning
+    Pivot,
+}
+
+impl MovementHold {
+    pub fn is_holding(self) -> bool {
+        self != MovementHold::None
+    }
+}
+
 /// Pathfinding service for AI navigation
 ///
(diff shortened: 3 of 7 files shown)


tommy-xr:
Evidence fix: `a_sustained_stall_draws_no_gesture` created no pathfinding service, so the "stalled" AI read zero stall seconds and the pre-change code would have passed it too. It now publishes a real 10 s stall on the follower's channel; restoring the old stall branch makes it fail. The give-up limit test also picks up the pivot gate that commit added. Re-audited the other three claims against the pre-change code - all genuine (each carries an unheld control, and the pivot one fails on code that froze the behavior). Tests only; `cargo test -p shock2vr` 1437 passed, check clean.

PR #1356, opened 2026-09-05
fix(ai): probe the unstick nudge before taking it

tommy-xr:
The stall-recovery unstick nudge is a **teleport**: after two consecutive stalls from the same spot, a pinned creature is moved about two Dark feet toward its retreat point. Its only check was that the destination belonged to a navigation cell — nothing about the wall in between, whether the body fits where it lands, or whether there is a floor under it — and a body *already* off the mesh skipped even that, taking the move unconditionally.

The destination is now probed before it is taken, and a candidate that fails any probe is simply not moved: the body grinds on until the re-path frees it, which is strictly what the pre-nudge code did.

## What the probes establish — and what they do not

They are **rays, not a sweep of the body's capsule**: the physics view a script holds exposes no shape cast, so this is an *approximation* of clearance and geometry that fits between the rays is not seen. What they do establish, measured from the creature's own dimensions (`ai_util::creature_height` and the new `creature_radius`, which resolves the same capsule the body actually collides with — authored sphere dimensions included, not a bounding box):

- **`floor_step`** — a downward ray at both ends; the landing's floor must be within a step (2 Dark feet) of the one the body stands on. The floors are measured **first** and the landing rides that difference, so every other probe is taken where the body would actually stand rather than at the height it is leaving.
- **`path_clear`** — rays from the body to the landing and out *past* it by a body radius, so a wall just beyond refuses the move too.
- **`clearance`** — rays to either side at the landing, one body radius each way.
- **`headroom`** — an upward ray at the landing, **comparative rather than absolute**: a two-foot nudge shares its ceiling with the spot being left, so the landing only has to be no tighter than that spot. Demanding a full standing height would refuse every candidate for a body pinned under a duct or a catwalk — the population the unstick exists for.
- Rays go at the body's origin **and** at the whiskers' own knee fraction of the creature's height (the constant is now shared with them, so the two cannot drift): a low slab passes under a single centre ray.
- The queries are **solid**, so a ray whose origin lies inside a collider — as a pinned body's does — reads a hit at zero range. That is neither a clear segment nor an obstacle: the probe steps past it and judges the **rest** of the segment, so a wall behind the prop the body is embedded in is still seen. A segment spent entirely inside geometry reports *embedded*, which refuses.
- The off-mesh skip is gone. Being off the mesh now waives only the *cell* check — there is no cell to keep such a body in — while the physical probes still apply.

## Was it still needed?

It fires rarely, so the first question was whether #1264's physics fix had made it dead code. It has not: it is rare **and** load-bearing.

| scenario | stalls | pinned | nudges |
|---|---|---|---|
| medsci2, 60 s patrol, no input | 13 | 0 | 0 |
| medsci2, forced chase, 41 s | 1 | 0 | 0 |
| medsci2, forced chase, 35 s (x2) | 1 / 1 | 0 / 0 | 0 / 0 |
| eng1, forced chase, 90 s (~20 AIs) | 5 | 1 | 1 |

**About one nudge per 90 s of a whole-deck chase, none at all in medsci2.** The one eng1 firing is a hybrid pinned on a ledge at `-58.77, -15.37, -49.88`, whose route waypoint sits a metre below it. A/B at that spot, same seed, same scene, position sampled once a second:

| | outcome |
|---|---|
| nudge on | pinned 8 s, freed at t=11 s, walks on |
| nudge off (build-time gate) | pinned for **every one of the 18 samples** it had a path for — never moves a millimetre |

So the re-path alone does not recover it, and cutting the nudge would trade a rare small pop for a creature frozen for the rest of the mission. Kept and hardened. (18 stationary samples are evidence that the re-path does not free it *within the run*, not a proof of permanent freezing.)

After the change the nudge still fires at that ledge and still frees the creature — re-checked at the final commit, same forced chase: 5 stalls, 1 pinned, **1 nudge taken, 0 refused**, `-58.77,-49.88 -> -59.50,-49.84`, the creature then walking on under its own locomotion. It takes the *cell-centre* candidate because the retreat-direction one fails its probes, which is the intended behavior.

## Tests

Unit, on `nudge_is_safe`: `every_failed_probe_refuses_the_nudge` (red when any single clause is dropped) and `a_clear_supported_landing_is_taken`.

Geometry, driving `probe_nudge` itself against a synthetic physics world (a floor slab plus the case's own boxes, one `PhysicsWorld::update` so the broad phase sees them). Each was verified red against the specific behavior it guards:

- `a_wall_behind_the_prop_the_body_is_embedded_in_refuses_the_nudge` — red with the near reading discarded instead of stepped past.
- `a_segment_that_never_leaves_the_geometry_refuses_the_nudge` — a probe spent wholly inside a block is not a clear ray.
- `a_body_embedded_in_a_prop_is_still_nudged_when_the_way_is_clear` — the counterweight: seeing past a prop must not become refusing everyone inside one.
- `a_step_up_alongside_geometry_refuses_the_nudge` — red when clearance is probed at the old height instead of the landing's.
- `a_step_up_under_an_overhang_refuses_the_nudge` — red without the headroom probe.
- `a_landing_no_tighter_than_the_spot_left_is_still_nudged` — red when headroom is absolute rather than comparative.
- `an_open_landing_is_probed_safe_and_rides_its_floor` — the positive control, including a landing half a foot up.

`cargo test -p shock2vr`: 1452 passed. `cargo fmt` and `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean.

e2e, **reduced scope by request** — only the scenarios that exercise stall recovery: `medsci2-patrol-railing` passes (run again after the review fixes). `medsci2-doorjamb-chase`'s first case is flaky on both branches at its marginal 6 s threshold; instrumentation settles the attribution — **zero pinned bodies and zero nudges in medsci2**, so this code path never executes there.

No media: the nudge is a two-foot pop in an unlit engineering corridor, and the change makes an existing move *refusable* rather than adding anything to look at.

## Review

Two rounds of `/xreview` (Claude adversarial reviewer, Codex, and a dedicated de-dup reviewer), plus an external claims re-review between them.

First round, fixed: an accepted landing kept the body's old `y` while `floor_step` allowed a two-foot rise; centre-height rays alone do not see a low slab; a zero-range reading from the body's own embedding would have refused every nudge for the most deeply pinned bodies; the clearance radius was a fixed 2 ft.

Second round — the external re-review's two HIGH findings and what both engines then agreed on:

- **High** — the near-origin reading was discarded *without querying beyond it*, so a wall behind the prop a body is embedded in was invisible. The probe now steps past such readings and judges the rest of the segment.
- **High** — the landing `y` was adjusted *after* the clearance probes ran at the old `y`, so a step-up could land in an overhang. Floors are measured first; clearance and the new headroom probe run at the landing the body would ride to.
- **High [AGREED, both engines]** — with the stepping in place, a ray that consumed its whole segment while still inside geometry returned *clear*. It now returns embedded, which refuses.
- **Medium [AGREED, both engines]** — `creature_radius` re-derived only `live_creature_shape`'s *fallback* half-width, so a creature with authored sphere dimensions was probed at a radius its capsule does not have. Both now share `live_creature_shape`.
- **Medium (Claude)** — an absolute headroom demand would freeze an AI pinned in a low pocket forever. Made comparative.
- De-dup pass: `creature_height`/`creature_radius` now share one definition lookup; the knee fraction is the whiskers' constant rather than a copy of its value; the repeated step-box setup is one test helper. Declined, with reasons: threading a `solid` flag through the shared raycast choke point (a one-query alternative to the stepping loop, but it changes the engine-wide query path for a probe that runs about once a minute), and hoisting a crate-wide `#[cfg(test)]` physics-world module (four call sites in three unrelated modules — its own change, not this one's).

### DIFF
--- a/shock2vr/src/mission/entity_creator.rs
+++ b/shock2vr/src/mission/entity_creator.rs
@@ -1555,7 +1555,9 @@ fn create_physics_representation_with_options(
     }
 }
 
-fn live_creature_shape(
+/// The collision shape a live creature is given: its authored sphere
+/// submodel radii where they exist, and the animation bounding box otherwise.
+pub fn live_creature_shape(
     creature_def: &crate::creature::CreatureDefinition,
     phys_type: Option<&PropPhysType>,
     dimensions: Option<&PropPhysDimensions>,
--- a/shock2vr/src/scripts/ai/ai_util.rs
+++ b/shock2vr/src/scripts/ai/ai_util.rs
@@ -66,29 +66,50 @@ const CREATURE_DEFAULT_HEIGHT: f32 = 6.5 / SCALE_FACTOR;
 /// ...and the body radius to fall back on (half a human's 3.5-foot width)
 const CREATURE_DEFAULT_RADIUS: f32 = 1.75 / SCALE_FACTOR;
 
-/// The creature's height in world units. Fractions of it (rather than fixed
-/// feet) are what let the same reasoning work on a monkey and on a hybrid.
-pub fn creature_height(world: &World, entity_id: EntityId) -> f32 {
+/// The creature's own definition, or None when it has no `PropCreature`.
+fn creature_definition(
+    world: &World,
+    entity_id: EntityId,
+) -> Option<std::sync::Arc<crate::creature::CreatureDefinition>> {
     world
         .borrow::<View<PropCreature>>()
         .ok()
         .and_then(|v_creature| v_creature.get(entity_id).ok().map(|creature| creature.0))
         .and_then(crate::creature::get_creature_definition)
+}
+
+/// The creature's height in world units. Fractions of it (rather than fixed
+/// feet) are what let the same reasoning work on a monkey and on a hybrid.
+pub fn creature_height(world: &World, entity_id: EntityId) -> f32 {
+    creature_definition(world, entity_id)
         .map(|definition| definition.bounding_size.y)
         .unwrap_or(CREATURE_DEFAULT_HEIGHT)
 }
 
-/// The creature's body radius in world units - the half-width its capsule is
-/// built from (`live_creature_shape`'s fallback radius), so a probe asks for
-/// the room this creature actually occupies.
+/// The radius of the capsule this creature actually collides with - the same
+/// resolution its body is built from, authored sphere dimensions included, so
+/// a probe asks for the room the creature really occupies rather than for a
+/// bounding box it may be much narrower than.
 pub fn creature_radius(world: &World, entity_id: EntityId) -> f32 {
-    world
-        .borrow::<View<PropCreature>>()
-        .ok()
-        .and_then(|v_creature| v_creature.get(entity_id).ok().map(|creature| creature.0))
-        .and_then(crate::creature::get_creature_definition)
-        .map(|definition| definition.bounding_size.x.max(definition.bounding_size.z) / 2.0)
-        .unwrap_or(CREATURE_DEFAULT_RADIUS)
+    let Some(definition) = creature_definition(world, entity_id) else {
+        return CREATURE_DEFAULT_RADIUS;
+    };
+    let v_phys_type = world.borrow::<View<PropPhysType>>().ok();
+    let v_dimensions = world.borrow::<View<PropPhysDimensions>>().ok();
+    let shape = crate::mission::entity_creator::live_creature_shape(
+        &definition,
+        v_phys_type
+            .as_ref()
+            .and_then(|view| view.get(entity_id).ok()),
+        v_dimensions
+            .as_ref()
+            .and_then(|view| view.get(entity_id).ok()),
+    );
+    match shape {
+        crate::physics::PhysicsShape::Capsule { radius, .. } => radius,
+        crate::physics::PhysicsShape::Sphere(radius) => radius,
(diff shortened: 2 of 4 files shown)


tommy-xr:
Amended for the re-review's two HIGH findings, plus what `/xreview` then found on top of them.

- A near-origin (embedded) reading is no longer discarded as clear: the probe steps past it and judges the rest of the segment, so a wall behind the prop a pinned body is inside is seen. A segment spent entirely inside geometry refuses.
- Floors are measured first and the landing rides its own floor, so clearance — and a new, comparative headroom probe — are taken where the body would actually stand.
- Heights and radius now come from the creature itself; `creature_radius` shares `live_creature_shape`, so a creature with authored sphere dimensions is probed at the radius its capsule really has.
- The "safe landing" claim is gone from the body. These are rays, not a capsule sweep — no shape cast is reachable from a script's physics view — and the body now says what they do and do not establish.
- Seven geometry tests drive `probe_nudge` against a synthetic physics world, each verified red against the specific behavior it guards (the old tests only fed booleans to `nudge_is_safe`).

eng1 ledge re-checked at the final commit: 5 stalls, 1 pinned, 1 nudge taken, 0 refused, creature freed and back to chasing. `medsci2-patrol-railing` green; `cargo test -p shock2vr` 1452.

PR #1357, opened 2026-09-05
fix(ai): the turn clip hands its yaw over on the player's own transition

tommy-xr:
A pivot is performed as the creature's own authored turn clip (#1337): the clip's **pose** does the visible turning, so the entity holds its heading underneath it and then takes the authored facing change as the pose swings back to neutral. Two things about that handover were guesswork.

**The settle window was the wrong clip's blend, and it was clocked by the wrong thing.** The AI spread the facing change over the *outgoing* turn clip's authored `blend_length`. The animation player blends with the *incoming* clip's — `queue_animation` builds no fade at all when the incoming clip authors none, and a stride clip authors none by design (`zero_blend_clip_still_hard_cuts`). So the pose gave the turn up in one frame while the entity took it over half a second, and the heading ramped **linearly** where the pose cross-fades on a **raised cosine**. The handover now rides the transition the player is actually running, read off the player every frame: the applier reports `AnimationPlayer::blend_alpha_now()` — the very weight `get_transforms` blends the two poses with — as `TurnClipBlend`, and the script takes exactly that fraction of the turn. It integrates nothing of its own, because it cannot: animations update before scripts, so any clock the script keeps runs a frame ahead of the pose it is supposed to cancel out.

**Elapsed time cannot identify a playback.** A completion was accepted as the turn clip's own if it arrived within a quarter-second of when the clip *should* have ended, so a late interruption could still commit an abandoned turn, and a pivot that was never answered fell back on a report timeout. Every `PlayTurnClip` now carries a **token**, and the applier — which owns the animation player and is the only thing that knows whose clip is playing — answers all of them, from the request to the last frame of the fade: `TurnClipStarted`, `TurnClipCompleted`, `TurnClipCancelled` when anything displaces the clip, `TurnClipHandoff` when the following clip starts (`fades` distinguishing a cross-fade from a hard cut), and `TurnClipBlend` for each frame of that fade. A report echoing a token the script has moved past is dropped. That bookkeeping is one type, `TurnClipTracker` (`shock2vr/src/mission/turn_clip.rs`), rather than three call sites in `mission_core`.

That makes cancellation explicit rather than inferred, and it cuts three timeouts (`TURN_CLIP_REPORT_TIMEOUT`, `TURN_CLIP_OVERRUN_MARGIN`, `TURN_CLIP_COMPLETION_SLACK`) and the `TURN_CLIP_SETTLE_SECONDS` fallback. One bounded wait survives, and only because a creature can legitimately queue nothing after its pivot: `TURN_CLIP_HANDOFF_TIMEOUT`, after which the pivot is **dropped**, not committed — with no following clip the player is still holding the turn clip's final frame, so taking the facing change there would turn the creature twice.

**The script guesses at nothing.** Damage, a door to stand off from and a behavior swap are no longer treated as preemptions: if nothing actually displaced the clip its pose is still turning the body and the turn is still owed. Only the applier's report ends a pivot — and every path that can take an entity's animation player away (`ChangeModel`, `ClearModel`, entity removal) now answers the pivot it orphans, so a held heading can never outlive its clip.

Ledger step P3 (`~/notes/projects/shock2quest/pathfinding-eval-plan.md`).

## Before / after

Same scenario as #1337 — medsci2, `DebugForceChase`, a chasing hybrid reversing through `ogstu180`, free camera locked to the creature — captured **one sim frame at a time** across the seam so the handover is resolved at 60 Hz.

before/after at the seam

<sub>Top row, before (`fix/ai-movement-hold`): the clip's last frame, then +1, +2, +10, +25 frames. The pose swaps to the run cycle and the creature is suddenly **back-on** — a 171 degree flip in one frame — and then rotates back over the next half second. Bottom row, after: the same five frames, and the facing never leaves the direction the clip ended on.</sub>

Entity yaw, per sim frame, from the clip's final frame:

| | before | after |
|---|---|---|
| pose swaps to the following clip | frame 87 | frame 88 |
| authored turn fully taken | frame 114 | frame 89 |
| frames the entity turns under a pose that is no longer turning | **28 (467 ms)** | **1 (17 ms)** |
| yaw per frame across that window | 5.7 deg | 171 deg, once |
| following clip's authored blend | 0 (`ogsrunl`) | 0 (`ogsrunl`) |

The whole 171 degrees now arrives on the frame the pose gives it up, which is what the sheet shows: no rendered frame in which the facing disagrees. Before, the entity's yaw was up to 171 degrees away from the facing the pose was drawing, for 28 consecutive frames.

after

before

<sub>1.5x slow motion (every 2nd sim frame at 20 fps), windowed on the seam. Clip selection is random per query, so the two runs are the same request sequence rather than the same creature at the same spot. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

One pre-existing artefact is visible in both and is untouched here: the pose *cut* at the seam (the largest single-frame image difference in either run, 6.4 vs 6.5 on a normalized frame diff) is the zero-blend stride clip hard-cutting from the turn clip's final pose. Giving the follow-up an actual cross-fade is a separate change; the point of this one is that whatever fade the player runs, the yaw now rides it.

### A fade that actually runs

The shipping follow-up is a stride clip, which hard-cuts — so the case where a fade *does* run was captured too, by injecting `Damage` during the pivot so the clip that follows it is a wound reaction (`ogprecsw`, floored at a 0.5 s blend). Same scene, one sim frame per step, reading the pose weight the renderer uses (`/v1/entities/{id}/animation` `blend.alpha`) beside the entity's yaw:

| frame | pose weight | fraction of the turn taken |
|---|---|---|
| 108 (fade starts) | 0.0000 | 0.0000 |
| 114 | 0.0955 | 0.0625 |
| 120 | 0.3455 | 0.2939 |
| 123 | 0.5000 | 0.4462 |
| 129 | 0.7939 | 0.7508 |
| 135 | 0.9755 | 0.9592 |
| 138 (fade ends) | 1.0000 | 1.0000 |

Across all 31 frames the yaw equals the pose weight of the frame before it to within **0.005 of the turn (0.86 deg)** — a constant one-frame offset, which is the entity transform's publication lag and shows identically in the zero-blend capture (there the whole -171.01 deg lands in a single frame, one frame after the hard cut). Never a lead, and no drift across the fade.

## Tests

Negative-first, each shown failing before the change that makes it pass:

- `the_yaw_never_leads_the_pose_across_a_real_fade` — a REAL `AnimationPlayer`, the REAL applier-side tracker and the REAL script, run in the mission loop's own order (animations, then reports, then the script, then its effects) over a follow-up clip that actually fades. Restore the script-side clock and it fails on the first frame of the fade: the pose is 0.0027 through it and the entity has taken 0.0109 of the turn.
- `a_clip_displacing_the_settle_takes_the_rest_of_the_turn_at_once` — fails if the applier stops tracking the pivot at the handoff (the settle then eases on across a fade that has been replaced).
- `the_yaw_follows_the_players_blend_curve` — fails with a linear ramp (the old behavior) at every sampled point.
- `a_completion_carrying_a_stale_token_is_ignored` — fails without the token check.
- `a_preempting_clip_drops_the_pending_turn` — fails if a cancel commits the turn.

Also new: `a_zero_blend_follow_up_hands_the_turn_over_at_once`, `taking_a_hit_does_not_abandon_a_turn_that_is_still_playing`, `a_pivot_nothing_follows_still_ends`, `a_pivot_with_no_clip_falls_back_to_steering`; in `dark`, `blend_alpha_now_is_the_weight_the_fade_has_reached` (with the load-bearing fact: a stride clip queued over a 500 ms clip is at full weight immediately — there is no fade to ride) and `blend_alpha_is_a_clamped_raised_cosine`.

- `cargo test -p shock2vr`: 1444 passed. `cargo test -p dark`: 162 passed.
- e2e, scoped to the scenarios that exercise pivoting: `medsci2-doorjamb-chase` 2/2, `medsci2-patrol-railing` 1/1 — doorjamb re-run 2/2 after the amendment below.
- Pathfinding is untouched (nothing under `shock2vr/src/pathfinding/` or `dark/src/mission/` is in the diff).

## Review

`/xreview`, both engines (Claude adversarial reviewer, Codex CLI) plus the de-dup reviewer, run twice — once on the original change and once on the amendment.

First pass, fixed:

- **High, [AGREED]** — the handover timeout committed a turn the pose was still holding. When nothing follows the pivot the player keeps the turn clip's final frame on screen, so adding the authored turn there reproduced the exact double-turn this work exists to remove. It now drops the pivot.
- **High, [AGREED]** — cancelling a pivot on damage or on a door wait cleared the script's state without stopping the clip: the pose kept performing the about-face while the entity was steered independently, and the completion it eventually reported was ignored. Neither is a preemption; both cancels are gone, along with the ones in `enter_death` and `force_alertness_state` (which play a clip, so the applier reports the preemption itself). The applier is now the sole authority.
- **Medium** — `ChangeModel` and `ClearModel` remove or rebuild the animation player without going through the animation applier, so no completion could ever arrive for a turn clip they orphaned. With the elapsed-time bound on `Playing` gone, that would have held the creature's heading for the rest of the level. Both answer the pivot now.
- **Low** — a `PlayTurnClip` that resolved to no clip left the previous playback entry behind, so a later clip reported a handover for a dead token (harmless only because of the token check). The entry is taken whenever an apply concerns a turn at all.
- **De-dup, `[S3-index, S4-read]`** — the debug animation endpoint's `alpha` field reported linear progress through the fade, not the pose's weight (they differ by up to 0.2). It reads the player's own weight now, so the endpoint and the renderer agree.
- **De-dup, `[S2-clones]`** — the new tests repeated a five-line "ask, answer, play out" pivot preamble; one `start_pivot` helper.

## Amendment — a second review pass

Two findings against the first pass, both fixed here:

- **High — the handoff frame was counted twice.** Animations update before scripts, so the script's own wait advanced on the frame the following clip was applied, and the settle it seeded advanced again on the next: **two dt of yaw against one dt of pose**, a gap that never closed for the rest of the fade (about 9 degrees of a 170 degree turn at mid-fade). Seeding the settle from that wait — which is what the first pass did — only papered over one of the two ticks. The script now keeps no clock: the applier reports the pose's own weight every frame, and the yaw is that number. The test above drives a real animation player through a real fade, which is the only way to see this: the previous test asserted the lead into existence by advancing nothing.
- **Medium — ownership ended at the first handoff.** The playback entry was dropped as soon as the following clip started, and `Settling` carried no token and ignored cancellation, so a wound or behavior clip landing mid-settle went unnoticed and the yaw kept riding a fade the player had already replaced. `TurnClipTracker` keeps the pivot through the settle and reports the displacement; `Settling` answers it like every other state — by **taking the rest of the turn in that frame**. `play_animation` fades from the follow-up clip's *own* pose rather than the blended one on screen (its documented limit), so the `turn * (1 - alpha)` the pose was still holding vanishes on the displacing frame: landing the remainder is what keeps the visible facing continuous, where dropping it would snap the creature back by that remainder. This is the opposite of the handover-timeout case, and for the opposite reason — there the player is still holding the turn clip's final frame, so the pivot is dropped.

Also from that pass's `/xreview`:

- **De-dup, `[S3-index, S4-read]`** — the debug animation endpoint recomputed the cross-fade weight inline; it calls `blend_alpha_now()` now, so there is one definition of the pose's weight and the endpoint, the renderer and the AI cannot drift apart.
- **De-dup, `[S1-names, S4-read]`** — with the curve carried frame by frame, the handoff's `blend` seconds was only ever squashed to `> 0.0`. It is a `fades: bool`, and `blend_remaining_seconds` (whose precision nothing read any more) is gone: one accessor, `blend_alpha_now`.
- **Low (Codex)** — `TurnClipBlend` is dispatched once per frame of a fade and would have pushed the damage/signal history out of the 256-entry message trace; it joins the other high-frequency payloads the trace excludes.
- **Medium (Claude)** — the snap-back above: the first cut of this amendment dropped a displaced settle where it stood. Fixed as described, and the test flipped with it.
- **Low (Claude)** — `DebugCycleHitboxPose` queued clips straight onto every creature's animation player, replacing a pivot's fade without the pivot hearing about it — the very class of bug this amendment fixes. It reports through the tracker now.

Noted, not changed: a pivot that times out waiting for a follow-up leaves its `Completed` entry in the tracker until that entity's next clip, which then reports a handover for a token the script has already dropped. Token matching makes it inert, and the entry is one per entity; bounding it would need exactly the clock this change exists to remove.

Noted, not changed:

- `TurnClip::Settling` is exercised on the *minority* path. Every follow-up the AI queues after a pivot is a stride clip authoring no blend, so the shipping case is a single-frame commit; a nonzero fade only comes from the clips that `play_animation` floors at 0.15 s (a wound reaction, a behavior swap, a handback landing on a completed pivot) — which is what the capture above injects to exercise it. Keeping the eased path is what makes the zero case *correct rather than lucky*, and it is what stops working the moment a follow-up gets a blend.
- Giving the turn clip's follow-up a real cross-fade (so the pose stops hard-cutting at the seam at all) is the obvious next step, and would make the eased handover the common path rather than the rare one.

### DIFF
--- a/dark/src/motion/animation_player.rs
+++ b/dark/src/motion/animation_player.rs
@@ -258,24 +258,13 @@ impl AnimationPlayer {
         new_player
     }
 
-    /// How long the cross-fade currently running still has to go, in seconds
-    /// (0 when nothing is fading) - read immediately after queueing a clip,
-    /// that is the whole window the fade will take. This is the blend the
-    /// player ACTUALLY applies - `queue_animation` takes the incoming clip's
-    /// authored length (often zero), `play_animation` floors it - so a caller
-    /// that must move in step with the fade reads it here rather than
-    /// re-deriving it.
-    pub fn blend_remaining_seconds(&self) -> f32 {
-        self.blend_state
-            .as_ref()
-            .map(|blend| (blend.duration - blend.elapsed).max(0.0))
-            .unwrap_or(0.0)
-    }
-
     /// How much of the pose the INCOMING clip owns right now: the very weight
     /// `get_transforms` blends with, so anything that has to move in step with
     /// the fade (the AI's turn-clip yaw handoff) reads the pose's own number
-    /// rather than integrating a clock of its own. 1.0 when nothing is fading.
+    /// rather than integrating a clock of its own. 1.0 when nothing is fading -
+    /// which, read straight after queueing, is also how a caller learns that
+    /// the clip it just started hard-cuts (`queue_animation` takes the INCOMING
+    /// clip's authored blend length, and a stride clip authors none).
     pub fn blend_alpha_now(&self) -> f32 {
         match &self.blend_state {
             Some(blend) if blend.duration > f32::EPSILON && blend.elapsed < blend.duration => {
@@ -1008,45 +997,24 @@ mod tests {
         }
     }
 
-    /// What the turn-clip yaw handoff reads: the fade the player is really
-    /// running. `queue_animation` takes it from the INCOMING clip - and a
-    /// stride clip authors none - so a caller that assumed the outgoing
-    /// clip's blend would settle over a window nothing is fading across.
+    /// What the turn-clip yaw handoff reads every frame: the weight
+    /// `get_transforms` blends with, so the facing the entity takes and the
+    /// facing the pose gives up are the same number. Read straight after
+    /// queueing, it is also what tells a caller whether there is a fade at all -
+    /// `queue_animation` takes the INCOMING clip's authored length, and a
+    /// stride clip authors none however long the clip it interrupts did.
     #[test]
-    fn blend_remaining_seconds_reports_the_fade_actually_running() {
+    fn blend_alpha_now_is_the_weight_the_fade_has_reached() {
         let mut outgoing = (*clip_with_root_motion()).clone();
         outgoing.blend_length = Duration::from_millis(500);
         let player = AnimationPlayer::queue_animation(&AnimationPlayer::empty(), Rc::new(outgoing));
-        assert_eq!(
-            player.blend_remaining_seconds(),
-            0.0,
-            "nothing to fade from"
-        );
-
-        // A stride clip queued after it: no fade at all, however long the
-        // clip it interrupts authored.
-        let stride = AnimationPlayer::queue_animation(&player, clip_with_root_motion());
-        assert_eq!(stride.blend_remaining_seconds(), 0.0);
-
-        // ...and one that does author a blend fades for exactly that long,
-        // counting down as it runs.
-        let mut incoming = (*clip_with_root_motion()).clone();
-        incoming.blend_length = Duration::from_millis(200);
-        let fading = AnimationPlayer::queue_animation(&player, Rc::new(incoming));
-        assert!((fading.blend_remaining_seconds() - 0.2).abs() < 1e-6);
-        let (fading, _, _, _) = AnimationPlayer::update(&fading, Duration::from_millis(50));
-        assert!((fading.blend_remaining_seconds() - 0.15).abs() < 1e-6);
-    }
-
-    /// The same fade as a pose weight: what `get_transforms` blends with, and
-    /// what the AI's yaw handoff is told every frame so the facing it takes
-    /// and the facing the pose gives up are the same number.
-    #[test]
-    fn blend_alpha_now_is_the_weight_the_fade_has_reached() {
-        let player =
-            AnimationPlayer::queue_animation(&AnimationPlayer::empty(), clip_with_root_motion());
         let (player, _, _, _) = AnimationPlayer::update(&player, Duration::from_millis(100));
         assert_eq!(player.blend_alpha_now(), 1.0, "nothing is fading");
+
+        // A stride clip queued after it: a hard cut, however long the clip it
+        // interrupts authored.
+        let stride = AnimationPlayer::queue_animation(&player, clip_with_root_motion());
+        assert_eq!(stride.blend_alpha_now(), 1.0, "no fade to ride");
 
         let mut incoming = (*clip_with_root_motion()).clone();
         incoming.blend_length = Duration::from_millis(200);
--- a/dark/src/motion/animation_player.rs
+++ b/dark/src/motion/animation_player.rs
@@ -272,6 +272,19 @@ impl AnimationPlayer {
             .unwrap_or(0.0)
     }
 
+    /// How much of the pose the INCOMING clip owns right now: the very weight
+    /// `get_transforms` blends with, so anything that has to move in step with
+    /// the fade (the AI's turn-clip yaw handoff) reads the pose's own number
+    /// rather than integrating a clock of its own. 1.0 when nothing is fading.
+    pub fn blend_alpha_now(&self) -> f32 {
+        match &self.blend_state {
+            Some(blend) if blend.duration > f32::EPSILON && blend.elapsed < blend.duration => {
+                blend_alpha(blend.elapsed / blend.duration)
+            }
+            _ => 1.0,
+        }
+    }
+
     pub fn set_additional_joint_transform(
         player: &AnimationPlayer,
         joint_idx: u32,
@@ -580,8 +593,7 @@ impl AnimationPlayer {
 
         if let Some(blend) = &self.blend_state {
             if blend.duration > f32::EPSILON && blend.elapsed < blend.duration {
-                let t = (blend.elapsed / blend.duration).clamp(0.0, 1.0);
-                let alpha = blend_alpha(t);
+                let alpha = self.blend_alpha_now();
                 // update() keeps from_frame in range (loops wrap, one-shots
                 // clamp), and the fractional position advances smoothly
                 // during the fade.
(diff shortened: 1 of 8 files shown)


tommy-xr:
Amended for the two re-review findings (commits `43e10961`, `68ef152b`, `0c96c43f`).

**High — the handoff frame was counted twice.** Animations update before scripts, so the script's own wait advanced on the frame the following clip was applied and the settle it seeded advanced again on the next: two dt of yaw against one dt of pose, ~9 degrees of a 170 degree turn at mid-fade, and the gap never closed. The script keeps no clock now — the applier reports the pose's own cross-fade weight (`AnimationPlayer::blend_alpha_now()`, the number `get_transforms` blends with) every frame as `TurnClipBlend`, and the yaw is that number.

**Medium — ownership ended at the first handoff.** `TurnClipTracker` (new, `shock2vr/src/mission/turn_clip.rs`) keeps the pivot through the settle; `Settling` carries its token and answers a displacement — by landing the rest of the turn in that frame, because the displacing clip takes the turn clip's pose with it (`play_animation` fades from the follow-up's own pose, not the blended one).

Tests, both negative-first: `the_yaw_never_leads_the_pose_across_a_real_fade` drives a real `AnimationPlayer`, the real tracker and the real script in the mission loop's own order — restore the script clock and it fails on the first frame of the fade (pose 0.0027, entity 0.0109). `a_clip_displacing_the_settle_takes_the_rest_of_the_turn_at_once` fails if tracking stops at the handoff.

Re-captured on medsci2 at one sim frame per step, with a nonzero fade as well as the shipping zero-blend one (a `Damage` injection makes the follow-up a wound clip): across all 31 frames of a 0.5 s fade the yaw equals the pose weight of the frame before it to within 0.005 of the turn (0.86 deg) — a constant one-frame transform-publication offset that shows identically in the zero-blend case. Never a lead.

`cargo test -p shock2vr` 1444, `-p dark` 162, `-D warnings` clean, e2e `medsci2-doorjamb-chase` 2/2. `/xreview` re-run on the amendment: both engines' findings applied (the snap-back above was the Claude reviewer's Medium).

PR #1362, opened 2026-09-05
fix(physics): a rising door leaf no longer lifts the creature under it

### LINKED ISSUE #1255: AI walks into a rising door leaf and gets carried upward
Seen in PR #1245's demo GIF (medsci2 Science-wing security door, OG-Pipe t705 under DebugForceChase), but pre-existing and generic to any raising door an AI opens.

Trace: the AI sends TurnOn to the door from ~4.8 units out (`animated_monster_ai.rs` door-open logic) and keeps walking. The leaf is a kinematic body moved by `set_next_kinematic_translation` each frame from `std_door`; the AI is a dynamic capsule whose root-motion write only preserves physics y-velocity. The leaf sweeps up through the capsule and contact resolution lifts it: hybrid y tracks the leaf 0.10 → 1.47 with vertical velocity pinned at the leaf speed (+2.28), then it hangs at ~1.5 for ~2.3 s after the leaf has cleared, then drops to the floor on the far side.

Two odd parts: (1) the lift itself; (2) the ~2.3 s hover after the leaf is well above it — suggests the AI body isn't under normal gravity / has a contact with the frame.

Fix direction: mirror the original's door handling — an AI approaching a door it just opened waits (steers away / holds) until the leaf has cleared actor height before continuing, rather than walking into the sweep. Alternatively exclude the leaf from actor collision while it is below clearance height. Also check why the body hovers after the leaf is gone.

tommy-xr:
An AI standing where a door leaf rises was carried up with it. At medsci1's Security Door a grunt walking through the doorway rode the leaf from its resting `y = 0.099` to `y = 1.719`, hung there for about three quarters of a second after the leaf had stopped, then dropped back to the floor and carried on.

Nothing about the contact is wrong; the friction the solver is allowed to apply through it is. The leaf is a kinematic body, and an actor pressed against its **face** makes a contact whose normal is horizontal. Rapier's friction constraint works to erase the relative *tangential* velocity between the two surfaces — and for a rising leaf that relative velocity is entirely vertical, so the capsule is dragged up to the leaf's own speed. (It is not `tangent_velocity`, which is the conveyor-belt opt-in and stays zero here; the post-step `linvel.y` goes from 0 to exactly the leaf's 2.4 u/s in a single step.) The **hover** is the same friction seen from the other side: once the leaf stops, the actor is still pressed against it and slips down its face at ~0.12 u/s until it breaks free.

So a Rapier contact hook drops friction, and only where it can do that:

- on **side** contacts alone — above a normal `y` of 0.5 the surface is under the actor and carries it through the contact *normal*, which this never touches, so lifts and platforms keep working;
- only when the mover travels **mostly vertically** — a sideways-sliding leaf transfers horizontal motion, which is a different mechanism with its own handling (`recover_live_creatures_swept_off_support`), left untouched;
- only on **actor** colliders, the only ones that ask for the hook.

The leaf still pushes along the contact normal, which is what shoves an actor standing in the doorway out of its way — no post-step Y clamp, and the leaf's collision is never disabled.

Ledger step P6 (`~/notes/projects/shock2quest/pathfinding-eval-plan.md`).

Fixes #1255.

## Before / after

medsci1, the rising Security Door on the corridor at `x = -18.4`. A grunt is spawned on the near side and left to stand up; the player crosses to the far side and shuts the door behind them; `DebugForceChase` sends the grunt through. Per-physics-step trace of leaf `y` / creature `y` / creature vertical velocity, deterministic fixed-timestep, same request sequence on both branches:

| frame | leaf y | before: creature y (vy) | after: creature y (vy) |
|---|---|---|---|
| 150 | 0.70 | 0.099 (0.00) | 0.099 (0.00) |
| 190 | 2.30 | **0.468 (+2.20)** | 0.099 (0.00) |
| 210 | 3.10 | **1.235 (+2.28)** | 0.099 (0.00) |
| 230 | 3.90 | **1.692 (−0.76)** | — leaf already parked, grunt through the doorway |
| 260 | 4.10 (parked) | **1.631 (−0.12)** — hovering | 0.099 (0.00) |
| 300 | 4.10 | 0.273 (−5.18) — falling | 0.099 (0.00) |

Summary, from the same runs:

| | before | after |
|---|---|---|
| resting y | 0.099 | 0.099 |
| peak y while the leaf travels | 1.719 | 0.105 |
| lift | **1.620** | **0.006** |
| unsupported hover | ~45 frames at ~1.65 | none |
| frame the grunt is through the doorway | 350 | 90 |

Platforms still carry: `moving-terrain-carry.e2e.test.ts` covers both a command1 tram passenger over a full leg and a grav-lift rider up and back down in lockstep, and both pass.

**`contact_count` was a lie.** `/v1/physics/bodies/:id` hardcoded it to `0`, which is exactly the field that tells an unsupported hover apart from a body at rest — it reported zero contacts while the narrow phase held two. It now reports the real active-contact count, and alongside it **`contacts`**: the `body_id` of each body actually being touched. "It touches something" becomes "it touches *that*", which is what separates a creature pressed against a particular leaf from one standing near it. Both are computed on the single-body detail path only, and the SDK grew `game.physics.body(id)` to read them.

**Not covered.** Corpses and ragdoll limbs are `SELECTABLE`, not `ACTOR`, so they carry no hook — a corpse leaning on a rising leaf is still dragged up. Out of scope for #1255, noted so it isn't mistaken for fixed.

## Tests

Both physics units press a capsule against the leaf **by construction**, so the contact the fix is about is guaranteed rather than hoped for, and both are negative-first (measured with the hook's body disabled, everything else identical):

- `a_rising_kinematic_leaf_does_not_lift_a_creature_against_its_face` — without the hook the capsule rides the leaf from `y = 0.999` to `1.603`; with it, under the 0.05 contact tolerance.
- `a_closing_kinematic_leaf_pushes_a_creature_clear_instead_of_through_the_floor` — the other direction, and the case of a creature that walked in under a leaf parked open before it shuts. Without the hook the capsule squeezed out sideways is flung up the descending leaf's face to `y = 2.02` and hangs at `1.47`, half a unit off the floor, after the leaf has shut; with it, it is pushed clear, never rises above its resting height, and ends standing at `0.999`. Its floor penetration bottoms out at 0.58 into the 1.0-thick floor slab either way — the leaf is kinematic and the port's door script has no obstruction/reopen path, so the pushout is the only thing keeping the creature out of the deck.
- `a_body_at_rest_reports_the_body_it_is_resting_on` pins the new contact report itself: the pair must read the same way round from both bodies.
- `a_rising_kinematic_platform_still_carries_the_creature_on_it` passes either way and is the guard against over-fixing: it fails if the hook ever starts eating support contacts.
- `cargo test -p shock2vr`: 1449 passed. `cargo fmt`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean.
- e2e: `moving-terrain-carry` 2/2. SDK `npm test` clean.

**The `door-leaf-lift` e2e is gone, and the reason is worth stating.** Its "the grunt reached the leaf" precondition was a Z-distance check, which a door stand-off satisfies without any contact. Replacing it with the real thing — is the grunt's body in the new `contacts` list of the leaf's body while the leaf moves — showed the run never touches the leaf at all: the AI opens the door from its stand-off and walks through the gap. Rerun with the hook disabled, the scenario does not reproduce the lift either (two runs ended with the grunt killing the player, no carry seen). Staging the contact by hand — shutting the leaf on the grunt in the doorway, then reversing it — did reproduce it in-game (contact with the *rising* leaf, dragged upward at 2.77 u/s), but landed only about one run in three. A test that passes on the build it is meant to catch is not evidence, so it is deleted rather than kept as reassurance; the proof is the two units above, where the contact cannot fail to happen.

## Review

`/xreview` (Claude + Codex + a de-dup pass), twice. Findings that changed the code:

- **The hook was reading last frame's leaf velocity.** Rapier runs `detect_collisions` — and therefore `modify_solver_contacts` — *before* `interpolate_kinematic_velocities`, so `linvel()` inside the hook is the previous step's. The frame a leaf starts moving is exactly the frame that matters, so the hook derives the pending move from `next_position() - position()` over the timestep.
- **The closing test's launch bound accepted the launch.** It allowed a full unit above rest — more than the peak it was meant to reject — and stopped sampling when the leaf stopped, which is *before* a thrown creature tops out. It now samples through the settle and bounds the peak at 0.25, failing at 2.02 without the hook.
- **The closing test's pushout was a coin flip.** Creature and leaf were both centred on `x = 0`, so which way the squeezed capsule left was solver tie-breaking, not geometry — and the assertion took the absolute value, hiding it. The leaf is offset by 0.2 and the escape direction is asserted with its sign.
- **`contacts` promised more than it reports.** Level geometry is inserted as colliders with no rigid body, so a creature standing on a mission floor raises `contact_count` and contributes nothing to `contacts`. Documented at each layer, and a unit test now pins the pair reading the same way round from both bodies.
- **The e2e's contact precondition** — the whole investigation above; the test is gone.
- **SDK type contracts**: the detail response names the linear velocity `linear_velocity` where the list calls it `velocity`, so `PhysicsBodyDetail` `Omit`s the inherited field instead of promising one the endpoint never sends; `body()` reports the endpoint's `null` for an unknown id as an error.
- Not taken: a shared helper for the "other side of this contact pair, then its parent body" idiom, which `recover_live_creatures_swept_off_support` also spells out. Five lines, and extracting it would edit the recovery path this change deliberately leaves alone.

## Media

before/after side-by-side loop: grunt riding the rising leaf up on the left vs. staying on the floor and walking through on the right

<sub>Side-by-side loop of the same request sequence on both branches (medsci1 Security Door, `DebugForceChase`). Left/BEFORE: the grunt is carried up the rising leaf's face and hangs before dropping. Right/AFTER: the grunt stays on the floor throughout and walks through the doorway. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

still comparison at the leaf's peak travel: before shows the grunt lifted above the door frame, after shows it still on the floor

<sub>Same moment (leaf near peak travel) on both branches.</sub>


tommy-xr:
Top of the linearized phase-2 chain: `feat/ai-turn-clips` → #1353 → #1354 → #1355 → #1357 → #1356 → #1362. #1356 was reparented from #1354 onto #1357, so this PR moved with it; all four commits replayed with no conflicts.

Verified here at the tip:

- `cargo fmt --check` clean
- `cargo test -p shock2vr -p dark`: 1447 / 151 / 11 pass
- e2e `medsci2-doorjamb-chase` x2 (2/2 each) and `medsci2-patrol-railing` x1 — all pass

`a824d720`'s behavior survived the restack: the single-division `creature_height`, and the stall→pivot-cooldown bar (now thresholded on the path follower's own `STALL_SECONDS` rather than the constant #1355 deletes) with its test.

tommy-xr:
Evidence pass on the review findings.

`/v1/physics/bodies/:id` now also reports `contacts` — which bodies a body is touching — and asserting that instead of Z-proximity showed the e2e's precondition was never met: the grunt does not touch the leaf on a fixed build, and with the hook disabled the scenario no longer reproduces the lift either (staging the contact by hand did: contact with the rising leaf, dragged at 2.77 u/s, but only ~1 run in 3). A test that is green on the build it exists to catch is anti-coverage, so it is deleted and the proof moved to the units, where the contact is guaranteed by construction.

The closing direction (and a creature standing under a leaf parked open, which is the same setup) is now a negative-first unit: without the hook the capsule is flung to y=2.02 and left hanging at 1.47 after the leaf shuts; with it, pushed clear and standing at 0.999. `moving-terrain-carry` 2/2, `cargo test -p shock2vr` 1449 passed.

PR #1367, opened 2026-09-05
fix(ai): a stalled crossing is excluded for the AI that stalled, not the level

tommy-xr:
When an AI's steering stalls mid-route it blacklists the crossing it could not walk, so its re-path routes around the obstacle instead of grinding on it (#481). That blacklist was **level-wide**: one `Mutex<HashMap<(from,to), expiry>>` handed to *every* AI's query, and unlike the stalled-cell report beside it — a cost penalty — a blacklisted crossing is a **hard cut** in A*. With a 30-second TTL, as long as an engagement, the set only ever grew over a pass. One creature wedged in a doorway removed that crossing for the other twenty, and where crossings are few — a narrow bridge between navigation islands — that left the rest of the level with no route at all.

Exclusions are now keyed by the AI that reported them and only that AI's queries carry them. Other AIs are still steered off the obstacle by the **shared cell penalty**, which is a cost and can never make a goal unreachable — occupancy is meant to be global, and stays so. An AI's set is bounded by the 30-second TTL alone and is dropped when the entity despawns, so nothing ratchets across a pass.

The steering comment at the report site claimed the exclusion was already "per AI"; it had been level-wide since the sharing landed. It is now true.

## What this does and does not move

Measured live on hydro2 and ops4 with `--experimental nav_bridges` + `DebugForceChase`, 30 s per pass, from one binary with the old level-wide behaviour behind a temporary switch. The pathfinding worker was instrumented to re-run every shortfall (a non-`Full` outcome) twice in the same frame: once with **all** exclusions lifted, once with only the **querying AI's own** — so "who cut this route" is answered per query rather than inferred.

| arm | hydro2 shortfalls | route existed with exclusions lifted | cut by *another* AI's exclusion | max exclusions on one query | AI path outcomes at pass end | `no_route` |
|---|---|---|---|---|---|---|
| level-wide (before) | 37 / 37 | 22 / 21 | 1 / 0 | 21 | 16 Full, 4 Partial, 2 Failed | 90 |
| per-AI (this PR) | 35 / 36 / 36 | 20 / 20 / 21 | **0** (by construction) | **4** | 17 Full, 3 Partial, 2 Failed | 87 |

ops4, two runs per arm: shortfalls 20/18 → 18/19, route-existed 5/2 → 2/4, `no_route` 55/52 → 52/53.

So the ratchet is gone and no query is ever cut by a stall it did not suffer — but the **shortfall count barely moves**, and that is the finding worth recording: in ~20 of hydro2's ~36 shortfalls the AI is severed by its **own single** exclusion, on a crossing that is its only way through. The level-wide sharing was not the dominant driver; the **hard cut itself** is.

A third arm settles it. Keeping the per-AI scoping and making an exclusion a heavy **cost** instead of a cut (the same shape as the stalled-cell penalty) gives hydro2 **15 shortfalls, 0 of them with a route available**, 21 Full / 0 Partial / 1 Failed, `no_route` 45 — the remaining 15 are one authored-disconnected creature that fails from t=0.32 s with an empty exclusion set. ops4 lands the same way (15 shortfalls, 0 recoverable, `no_route` 45).

That change is **deliberately not in this PR**: it trades a parked AI for a possibly grinding one where a pocket has a single exit — the exact case #481 chose parking for — and it would make #1353's temporary-unreachable machinery inert, since an exclusion could then never be why a goal is unreachable. It wants its own PR, with the medsci1 capture-pocket and door-jamb scenarios as the evidence.

## Tests

Negative-first: with the entity filter neutered (the old shared read, same keys), `a_stalled_ais_exclusion_leaves_every_other_ais_route_intact`, `an_ais_exclusions_are_capped_at_its_own_oldest`, `blocked_link_expiry_reports_when_the_exclusions_lapse` and `worker_applies_blocked_links` all fail; with it, they pass.

- `a_stalled_ais_exclusion_leaves_every_other_ais_route_intact` — X's report severs X's only route; Y still routes through it, and the shared *cell* penalty is what Y sees instead.
- `worker_applies_blocked_links` — the reporter's query comes back `Partial` with a deadline to wait out; the bystander's comes back `Full` with none.
- `an_ais_exclusions_outlive_only_their_ttl` — a lapsed report is dropped on the AI's next read, its live one kept, and its pruning leaves another AI's set alone.
- `blocked_link_expiry_reports_when_the_exclusions_lapse` (#1353's deadline) — extended: another AI's exclusions are not this query's deadline. #1353's patrol retry path is unchanged and its tests are untouched.
- `prune_drops_steering_with_the_path_record` — a despawned AI's exclusions go with it (previously asserted to survive), while a surviving AI keeps its own.
- `stats_report_the_live_blacklists` — the count stays level-wide; two AIs excluding one crossing are two live entries.

`cargo test -p shock2vr` 1450 passed. `cargo fmt`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean. `cargo bn path bench medsci2.mis --seed 424242` unchanged (200/0 found, inflation 1.67, 45.8 waypoints — A* itself is untouched). e2e: `medsci2-patrol-railing` 2/2, `patrol` 4/4.

## Review

`/xreview`, both engines plus the de-dup pass. Codex: no findings at any severity. Applied from the Claude reviewer and the de-dup pass, which agreed on the shape:

- **Medium (both)** — the flat composite key `HashMap<(u64,(u32,u32)), f32>` forced an entity filter in four places; it is now `HashMap<u64, HashMap<(u32,u32), f32>>`, the shape the sibling per-entity maps (`ai_paths`, `ai_steering`, `movement_holds`) already use. Reporting, reading and pruning are direct lookups and `prune_ai_paths` is a one-line `retain` like its neighbours.
- **Medium (scope)** — the per-AI cap and its oldest-eviction are gone. The TTL and the despawn prune already bound the set (a stall can only report every 3 s), and evicting the oldest handed back a crossing the AI still could not walk. That removed the tie-break non-determinism flagged separately.
- **Medium** — `merge`: `blocked_links` and `blocked_link_expiry` were the same lock/prune/filter body with different projections; one private `live_links_for` serves both.
- **Medium** — `prune_ai_paths`' doc still said blocked crossings were world facts that expire only on their TTL, which this change makes false. The stat's meaning (per AI, so one crossing can count twice) is now stated in the Rust doc and the SDK type.

**Noted, and the reason it is not a blocker.** The reviewer's strongest objection is that a *fresh* AI arriving at the same physical blocker no longer inherits a cut, only the shared cell penalty (a cost, 10 s) — the case the sharing was introduced for (medsci1's stationary-NPC pocket). Two things bound it: the penalty is refreshed on **every** stall, so it stays live for as long as anything is actually blocked there, and the worst case for a new arrival is one 3 s stall before it holds its own exclusion — against a level-wide cut that measurably disconnected hydro2 for everyone. A 60 s medsci1 forced chase (no bridges) shows no divergence between the arms — 13 vs 11 `Partial`, 1 vs 3 `Failed`, and exclusions applied to one query 29 → 7.

Declined: narrowing the retry deadline to the crossings the failed search actually touched (`blocked_link_expiry` still returns the max over the AI's own set, so a patrol can over-wait) — that is #1353's own follow-up and lands there.

### DIFF
--- a/shock2vr/src/game_scene.rs
+++ b/shock2vr/src/game_scene.rs
@@ -1111,7 +1111,9 @@ pub struct DebugPathfindingStats {
     pub stressed_retries: u64,
     pub no_route: u64,
     /// Cell crossings currently excluded because an AI's steering stalled
-    /// on them (see `PathfindingService::report_blocked_link`)
+    /// on them, summed over every AI holding one - each exclusion applies
+    /// only to the AI that reported it (see
+    /// `PathfindingService::report_blocked_link`)
     pub blocked_links: usize,
     /// Cells currently penalized because an AI stalled inside them
     pub blocked_cells: usize,
--- a/shock2vr/src/pathfinding/async_queries.rs
+++ b/shock2vr/src/pathfinding/async_queries.rs
@@ -29,7 +29,7 @@ pub struct PathQueryRequest {
     pub start: Vector3<f32>,
     pub goal: Vector3<f32>,
     pub movement_bits: MovementBits,
-    /// Mission time (seconds) at submit, for expiring the entity's
+    /// Mission time (seconds) at submit, for expiring the entity's own
     /// steering-reported blocked crossings (see
     /// `PathfindingService::report_blocked_link`)
     pub now_seconds: f32,
@@ -80,12 +80,13 @@ impl AsyncPathfinding {
             .spawn(move || {
                 while let Ok(request) = rx.recv() {
                     let mut outcome = AiPathOutcome::Full;
-                    // Crossings ANY AI's steering reported as physically
-                    // blocked (stall mid-route) are excluded, and cells an
-                    // AI is stalled in are made expensive, so re-paths -
-                    // including fresh arrivals' - route around the obstacle
-                    // instead of piling into it
-                    let avoid = service.avoidance(request.now_seconds);
+                    // Crossings THIS AI's steering reported as physically
+                    // blocked (stall mid-route) are excluded from its own
+                    // search, and cells any AI is stalled in are made
+                    // expensive, so the re-path routes around the obstacle
+                    // instead of grinding on it - without cutting the
+                    // crossing out from under every other AI
+                    let avoid = service.avoidance(request.entity, request.now_seconds);
                     let path = service
                         .find_path_avoiding(
                             request.start,
@@ -115,6 +116,7 @@ impl AsyncPathfinding {
                     let exclusion_expires_at = (outcome != AiPathOutcome::Full)
                         .then(|| {
                             service.exclusion_expiry_for_unreached_goal(
+                                request.entity,
                                 request.start,
                                 request.goal,
                                 request.movement_bits,
@@ -261,16 +263,16 @@ mod tests {
 
     #[test]
     fn worker_applies_blocked_links() {
-        // Cell 0 -> goal in cell 2, but an AI reported the 1 -> 2 crossing
-        // blocked: the worker must return a PARTIAL route ending at cell 1
-        // instead of the full route through the obstacle - for EVERY AI,
-        // not just the reporter.
+        // Cell 0 -> goal in cell 2, but entity 11 reported the 1 -> 2
+        // crossing blocked: ITS query must come back with a PARTIAL route
+        // ending at cell 1 instead of the full route through the obstacle
+        // it could not walk - and every other AI's must be untouched.
         let service = Arc::new(PathfindingService::new(Arc::new(
             crate::pathfinding::tests::three_cell_db(
                 dark::mission::path_database::PathCellFlags::empty(),
             ),
         )));
-        service.report_blocked_link(1, 2, 0.0);
+        service.report_blocked_link(11, 1, 2, 0.0);
         let async_pf = AsyncPathfinding::spawn(service.clone());
         assert!(async_pf.submit(PathQueryRequest {
             entity: 11,
@@ -287,7 +289,14 @@ mod tests {
             "partial route must end at cell 1's center, before the blockage"
         );
 
-        // The exclusion is shared: another entity's query avoids it too
+        assert_eq!(
+            response.exclusion_expires_at,
+            Some(super::super::BLOCKED_LINK_TTL_SECONDS),
+            "the shortfall is the reporter's own exclusion, so it can wait it out"
+        );
+
+        // The exclusion belongs to the AI that stalled: another entity's
+        // query still crosses, and has nothing to wait for.
         assert!(async_pf.submit(PathQueryRequest {
             entity: 12,
             start: cgmath::vec3(1.0, 0.0, 1.0),
@@ -296,7 +305,12 @@ mod tests {
             now_seconds: 1.0,
         }));
         let response = wait_for_result(&async_pf, 12).expect("worker must respond");
-        assert_eq!(response.outcome, AiPathOutcome::Partial);
+        assert_eq!(
+            response.outcome,
+            AiPathOutcome::Full,
+            "one AI's stall must not cut the crossing out from under another"
+        );
+        assert_eq!(response.exclusion_expires_at, None);
     }
 
     #[test]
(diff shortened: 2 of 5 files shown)


tommy-xr:
## Phase-2 chain restacked — every branch now sits on its parent's tip

Bottom-up rebase pass (rerere off). Old -> new tips:

| PR | branch | old | new |
| --- | --- | --- | --- |
| (base) | `feat/ai-turn-clips` | a824d720 | a824d720 |
| #1353 | `fix/ai-patrol-temporary-unreachable` | dfe5c456 | dfe5c456 (already current) |
| #1354 | `fix/ai-repel-head-on` | 5882b6ac | **8543c10f** |
| #1355 | `fix/ai-movement-hold` | fe620ae1 | **5fc163e6** |
| #1357 | `fix/ai-turn-clip-handoff` | 0c96c43f | **0d3a4515** |
| #1356 | `fix/ai-unstick-nudge-safety` | 86f34099 | **c104f556** |
| #1362 | `fix/door-leaf-lift` | a43f24fe | **a96ebbcf** |
| #1367 | `fix/ai-stall-exclusion-per-ai` | 976cf446 | **c929e196** |

`range-diff` is all `=` for #1354/#1355/#1357/#1356/#1362; every PR still shows only its own commits and no base moved.

### Conflicts, and how they were resolved

**1. #1367 over #1353's R6 (the one predicted).** R6 replaced the level-wide `blocked_link_expiry(now)` with `exclusion_expiry_for_unreached_goal(...)` + a private `blocked_link_expiries(now)`; #1367 made the same map per-AI. Resolved so both survive:

- `exclusion_expiry_for_unreached_goal` takes the query's own `entity` and reads only that AI's expiries — the route-scoped deadline stays R6's, the scope becomes #1367's.
- `blocked_link_expiries` is gone: #1367's review commit already introduced `live_links_for(entity, now)` as the one lock/prune/filter reader, so `blocked_links` and the deadline both read through it (the de-dup finding that commit was making, applied to R6's caller as well).
- `blocked_link_expiry` (singular) is dropped — R6 had already deleted the function it was per-AI-ifying.
- Tests: both sets kept. #1353's three deadline tests now query as `STALLER`; #1367's "another AI's exclusions are not this query's deadline" assertion moved onto `exclusion_expiry_for_unreached_goal(BYSTANDER, ...)`, since the function it used to assert on no longer exists. The worker test `worker_reports_a_deadline_only_for_an_exclusion_caused_shortfall` reports its three crossings for **both** querying entities, exclusions now being per-AI.

**2. A silent one that `cargo check -p shock2vr` cannot see.** #1355's R4 commit (fe620ae1) added a test constructing `TurnClip::Playing { turn, remaining, blend }` — but #1357, which sits *above* #1355, had already reduced `Playing` to `{ token, turn }`. The rebase applied without a textual conflict and left #1357 and everything above it failing to compile *in test code only*. Fixed at #1357 with one commit — `0d3a4515 test(ai): the give-up test builds a pivot in the applier's shape` — so #1357 has 6 commits now; #1356/#1362/#1367 were then re-rebased onto it. (Lesson for the next pass: validate with `cargo check -p shock2vr --all-targets`, not the lib alone.)

**3. #1356 over #1354's R3** — expected, did not happen. `blend_biases`/`aim_with_bias`/`yield_sideways`/`lateral_axis` (R3) and `probe_nudge`/`nudge_is_safe` (R2) merged cleanly; verified all present.

### Validation at the top of the stack (c929e196)

- `cargo fmt --check` clean; `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean; `cargo check -p shock2vr --all-targets` clean.
- `cargo test -p shock2vr -p dark`: **1465 / 151 / 11 pass, 0 failed**.
- `cargo bn path bench medsci2.mis --seed 424242`: **200/0, inflation 1.67, turn 944 deg** — unchanged.
- e2e: `medsci2-doorjamb-chase` **2/2**, `medsci2-patrol-railing` **1/1**, `patrol` **4/4**.

PR #1373, opened 2026-09-06
feat(vr): glove look back, with an affordance light on the cuff and fingertips

tommy-xr:
The VR hands wear the glove's own colour map again instead of the bare-skin
recolour, and the sleeve tube (`hand_forearm.rs`) is gone - the hand ends at the
glove's own cuff. The forearm HUD panels are untouched.

In its place the glove can light up. `SkinnedMaterial` gains an optional
emissive-mask sampler plus a `vec3` tint (`finalColor += mask.r * tint`), behind
a uniform-coherent branch so every other skinned mesh pays nothing; the mask
(`assets/vr_glove_emissive.png`, 256px) is generated from the model by
`tools/make_vr_glove_emissive.py` - a band round the wrist cuff and a pad per
fingertip, so the state reads from either side of the hand. A `HandLight` enum
(Off / Green / Amber / Red) picks the tint per hand.

Nothing sets the light yet: `VirtualHand::render` passes `Off` and no
interaction behaviour changes. Wiring it to reach/grab/frob state is the next
slice.

`debug_gloves` grows a third row showing all four states.

Verified: `cargo test -p shock2vr`, warnings-free check, `vr-held-model` +
`vr-forearm-readouts` e2e, and render checks of `debug_gloves --vr`,
`debug_minimal --vr` and `debug_melee` (an ordinary skinned creature, to confirm
the shader change is inert for it).

## Visuals

debug_gloves row 3, front and back: off, green, amber, red

<sub>`debug_gloves --vr`, bottom row: Off, Green, Amber, Red. Left, palm side (fingertip pads + cuff); right, back of hand (the stripes down the fingers and the wrist strap).</sub>

glove fist animation

<sub>Middle row's rightmost glove looping open→fist over a 3-second blend cycle.</sub>

The mask is authored at 1024px to stay hand-paintable; the texture loader expands it to RGBA on upload, so downscaling on load (not shrinking the source asset) is the saving available if Quest VRAM ever becomes a concern.

## Before → after

before/after: bare-skin sleeved hand vs glove colour map, sleeve removed

<sub>`debug_minimal --vr`: before, bare-skin hand with a sleeve tube; after, the glove's own colour map with the sleeve removed - hand ends at the cuff, forearm HUD panel unchanged.</sub>

PR #1376, opened 2026-09-06
feat(vr): the glove's light and pose read one resolved affordance

tommy-xr:
Slice 2 of the VR hands round-2 campaign. Stacks on #1373.

The hand's ray already decides what the trigger and squeeze act on. That same hit now resolves one affordance per hand per frame — `None` / `Grabbable` / `Frobbable` / `Blocked` / `Failed` — and drives two channels:

- **light = eligibility**: Off / Green / Amber, plus a ~250 ms Red pulse when an attempt is refused. Hysteresis holds a hover 6 frames so the light can't strobe at a collider edge.
- **pose = action**: the hand leans toward the authored `fist` (grab) or `point` (press) pose, weighted by how close the target is. Applied so a real pull always shows through — `fist` only ever adds curl, and the point prompt leaves the index finger exactly where the trigger put it.

The coloured raycast-hit cube and `color_from_state` are gone; the light replaces them.

Each hand's state is reported at `/v1/info` → `player.hand_affordance.{left,right}`, with an e2e test that aims a hand at a pickup and at nothing.

Left: nothing in reach. Middle: a wrench in reach (green, grip). Right: a hybrid in reach (green, index extended to frob).

states

Approach, in `debug_melee`: the light comes up and the fingers close as the hand enters reach.

approach

Not captured on screen: Amber (`Blocked`) and Red (`Failed`) — no debug scene has a locked target within reach, and neither state is reachable in one without staging a mission. Both are covered by unit tests, and #1373's `debug_gloves` row already renders all four light colours.

No haptics: the codebase has no haptic path to pulse.

### DIFF
--- a/runtimes/debug_runtime/src/commands.rs
+++ b/runtimes/debug_runtime/src/commands.rs
@@ -921,6 +921,10 @@ pub struct PlayerInfo {
     /// Climb state: whether the player is on a ladder/hand hold at all, and
     /// (VR) which hands hold what. See `shock2vr::vr_climb`.
     pub climb: shock2vr::game_scene::DebugClimbState,
+    /// What each VR hand could do with whatever it is pointing at - the state
+    /// that lights the glove and pre-shapes its fingers. See
+    /// `shock2vr::hand_affordance`.
+    pub hand_affordance: shock2vr::game_scene::DebugHandAffordances,
 }
 
 /// Input state snapshot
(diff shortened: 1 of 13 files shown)


tommy-xr:
Closing as adapted/superseded by #1412, per the project owner's direction to continue the Astra interaction stack.

The feedback states are adapted onto Astra's existing hand target and input decisions, with time-based refusal pulses and the production door lock/keypad rule. Automatic preshaping is intentionally not adopted; Astra's current analog and authored poses remain the chosen behavior.

The replacement PR includes both-hand visual evidence, actual production-button refusal checks, and existing grip regressions. Headset brightness/readability is still pending; this closure does not claim headset validation of the original PR. Branch retained for downstream reuse candidates.

PR #1168, opened 2026-08-25
feat(hud): psi and overload meters on the wielded psi amp

tommy-xr:
## Summary
- Add the on-amp psi pool bar and hold-to-overload meter, under the `ambient_meters` experimental flag (stacked on #1091's on-weapon ammo tag / glance-gated wrist health prototype).
- New `hud::psi_amp_panel` module (readout + layout, mirroring `hud::ammo_panel`), hung off the wielded psi amp's live `RuntimePropTransform` via `hud::ambient_meters::amp_psi_meter_root`, tuned to sit beside the amp's lit indicator strip.
- Draws in **both** presentations - VR hangs it directly off the amp; flat premultiplies the same root transform by the viewmodel's FOV `squish` so it stays pinned to the amp instead of floating off it.
- Closes a flat/VR parity gap: the hold-to-overload charge was gated to flat-only in game logic (`PsiAmpScript`, not just the HUD) with a comment that VR should keep cast-on-pull "until a VR meter exists". Now that a VR meter exists, the gate is `is_flat || ambient_meters_enabled` (a new `GlobalAmbientMetersEnabled` Unique resource next to the existing `GlobalPresentationMode`), so VR only gets the real hold-to-overload gesture where a meter can show it - never blind.
- Kept the flat bio-monitor's existing psi bar (owner decision) - this adds the on-amp meter alongside it.

## Visuals

on-amp psi meter charging

<sub>Flat presentation, `debug_psi`: the small red psi-pool bar and green/yellow hold-to-overload bar sit right on the wielded amp, next to its lit indicator strip, alongside the existing center-screen overload meter. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

on-amp psi meter steady state

<sub>Steady state (no charge in progress): just the psi-pool bar.</sub>

The same root transform (premultiplied by the flat viewmodel's FOV `squish`) drives the VR presentation - verified manually via `cargo dbgr --mission debug_psi --experimental ambient_meters --vr` with the amp grabbed and a free-camera framing: the meter lands in the same relative spot on the amp, and holding the trigger fills the same overload bar.

## Test plan
- [x] `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`
- [x] `cargo fmt`
- [x] `cargo test -p shock2vr --lib` (1143 passed)
[...]

PR #1378, opened 2026-09-06
feat(vr): a health watch on the left wrist, the ammo gauge on the right

tommy-xr:
wrist watch draining

<sub>The left glove's watch as the player takes damage: the bar drains and the number follows (100 -> 20). Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

both wrists

<sub>Left: the watch. Right: the ammo gauge with the debug pistol held (12 rounds, STD).</sub>

## Before -> after

before/after

<sub>Top: the retired BIOFULL forearm panel (health 100 + psi 80), same scene on the base ref. Bottom: the wrist watch that replaces it.</sub>

The two VR forearm panels are gone. Each glove now wears a small canvas at the
wrist instead: a **watch face** on the left (health only — number + bar) and the
**ammo gauge** on the right.

- **Health only on the watch.** Psi stays on the psi amp / cyber interface, so
  the watch is the one number worth a glance mid-fight.
- **The game's own art, cropped.** `ImageKind::Crop` samples a sub-rectangle of
  a shipped bitmap through the UV path the hologram tiles already use, so the
  watch is the compact bio plate's *health row* (BIO.PCX) and the right wrist is
  the ammo gauge's *well* (AMMOFULL.PCX) — no second, hand-cut copy of the art.
  The crops stop where the art's cyan colour key starts, which ordinary UI
  images do not drop.
- **One layout, as before.** Both canvases are emitted by the shared
  presentation-agnostic layout (`hud::readouts`, `hud::ammo_panel`) at the same
  rects the cyber interface and the flat HUD place, shifted by the crop — a
  wrist cannot drift from either (AGENTS.md §3). `hud/wrist.rs` only decides
  where the plate hangs on the hand.
- **Placement.** One rotation about the hand's X lays the canvas on the back of
  the wrist, just past the glove's cuff so #1373's emissive ring still reads.
  Canvas up points at the fingers; the same offset and basis serve both hands.
  Every wrist canvas is the same width (~5.9 cm) and keeps its own aspect.
- An unwielded right wrist draws nothing at all, matching flat, which drops the
  whole gauge rather than showing an empty one.

Retired: `create_forearm_hud_with_overlays` / `create_forearm_hud_panel`,
`forearm_pose`, `forearm_panel_transform`, `FOREARM_OFFSET`,
`HUD_PANEL_WIDTH/HEIGHT`, and the BIOFULL/AMMOFULL forearm quads.
`hud/virtual_arms.rs` → `hud/wrist.rs`; `create_arm_hud_panels` →
`create_wrist_hud_panels` (callers in `interaction.rs` and `scenes/debug_hud.rs`
otherwise unchanged).

Supersedes #1091 and #1168 (leaving both open for the owner to close).
#1230 (draft) is superseded too — the panels it repositions no longer exist.

## Tests

- Unit: the watch is the interface's health row shifted by its crop, element for
  element; the wrist ammo canvas is the panel layout shifted by *its* crop, all
  inside the plate; the crop holds the health row and not the psi row; the plate
  faces out of the back of the hand; both wrists share a width and keep aspect;
  an empty canvas hangs nothing.
- e2e `vr-wrist-readouts` (renamed from `vr-forearm-readouts`): the wrist
  canvases exist under render source `player_hands` and go quiet in use mode;
  wielding a gun adds the ammo gauge; the health readout follows the player's
  hit points. Also re-ran `vr-held-model`, `vr-hand-affordance`,
  `vr-interface-readouts`, `bio-ammo-full`, `ammo-readout` — all green.


tommy-xr:
Adapted in #1436, stacked on Astra glove feedback #1412. Reused the compact cropped-art approach, with health plus psi on both glove backs, per-hand weapon ammo underneath for dual wielding, and mounting derived from the calibrated glove and final visual hand pose. Authored forearms such as the psi amp need a dedicated mount and are explicitly deferred. Closing this competing implementation as superseded; preserving its branch for reference.

Issue #1262, opened 2026-09-03
e2e: 8 tests fail on main (d693a5a4)


tommy-xr:
Found while running the full SDK e2e suite for #1248 (comparison reruns on main at d693a5a4). Failing on main, independent of that branch:

- creature loot: Watts yields the 12451 disc (known)
- Earth hackable keypad (known)
- hydro3: Korenchkin log transcript (known)
- use mode exposes AMMOFULL ammo-cycle for an empty multi-ammo weapon
- MedSci saved mission Wrench keeps canonical 9-damage VR melee (fresh Wrench emits 0 Damage msgs, expected 1)
- eng1: a locked grav-lift button only works once its unlock trap fires
- VR authored melee damages only during a trigger-gated physical contact window (breakable pane)
- the flat Developer screen launches a debug scene (launches `cs1.avi` instead of `debug_minimal`), and the VR variant via controller ray

CI does not run e2e, so these accumulated silently. Details in https://github.com/tommy-xr/shock2quest/pull/1248#issuecomment-5520636904

PR #1380, opened 2026-09-06
feat(vr): the glove holds a wielded gun, and the baked arm goes away

tommy-xr:
Slice 4 of the VR interaction pass, stacked on #1378.

A VR-wielded 25AE gun no longer draws the hand and forearm baked into its view model. The player's own tracked glove is drawn on it instead, so the hand they see is the hand they are moving. The psi amp and the melee `_h` rigs are unchanged.

the glove holding a wielded pistol, index finger following the trigger

<sub>One trigger pull and release on a VR-wielded pistol. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

still

## Before to after

pistol: baked hand and sleeve, plus a floating spare hand, become the glove

The pistol's baked arm *and* the spare reload hand at the left edge are gone, and the glove lands where the baked hand was.

assault rifle: the baked rest hand becomes the glove

## Stripping the arm — by material, not by island size

Every arm in the gun set is on an `ND-arm*` material and nothing else is, so `is_first_person_arm_material` is the whole classifier. The largest-arm-island rule tried in #1023 was wrong because "one arm island = one hand" is false — `atek_h` splits its firing arm into a sleeve and a hand, so that rule kept the sleeve and deleted the grip.

Material table, from `cargo run -p shock2vr --example gun_hand_islands` (added here):

| model | arm materials (stripped) | weapon materials (kept) | baked hand islands |
| --- | --- | --- | --- |
| `atek_h` | `ND-arm.psd` 710, `ND-arm_atek.psd` 440 | `ND-ammo1.psd`, `ND-bulletcasing`, `ND-atek.psd` | 3 — sleeve 445, firing hand 440, spare hand 265 (0.9 units behind the grip) |
| `ar15_h` | `ND-arm.psd` 795 | `ND-ar15.psd`, `ND-ar15_il.psd`, `ND-bulletcasing` | 1 — a rest hand over the receiver |
| `sg_h` | `ND-arm_atek.psd` 791 | `ND-shotgun.psd` | 1 — rides the pump |
| `lasehand` | `ND-arm.psd` 733, `ND-arm_las.psd` 771 | `ND-laserp.psd`, `ND-laserp_il.psd`, `ND-laserp_t.psd`, `SCAN_.png` | 2 |
| `empgun_h` | `ND-arm.psd` 791 | `ND-empgun.psd`, `ND-empgun.psd1` | 1 |
| `gren_h` | — | `ND-glaunch.psd` | 0 |
| `sfg_h` | — | `ND-sfg.psd` | 0 |
| `fsn_h` | — | `ND-FSN.psd`, `ND-FSN_il.psd` | 0 |
| `al_h` | — | `ND-al.psd` | 0 |
| `viro_h` | — | `ND-viro.psd` | 0 |
| `amp_h` | — | `ND-amp_h.psd`, `ND-amp_il.psd` | 0 — the arm is modelled into the amp |

No hand uses a non-`ND-arm` material, and the amp is the reason the psi-amp exception exists at all: nothing isolates its forearm, so there is nothing to strip and nothing for a glove to replace. The strip runs in its own importer view (`VrHeldGunModel`, its own cache bucket — see the `type_id` trap), so vhots, sub-objects and the flat presentation are untouched.

## Where the glove goes

On the tracked hand, at the view model's own scale. `HAND_MODEL_POSITIONING` already carries a hand-fitted grip per model, tuned in #1023 *with the baked hand drawn* — which is exactly the alignment the glove wants, and the before/after below shows the glove landing where that hand was.

Inheriting the placement from the art instead — each baked hand carries a `HandFrame`, so the glove's wrist could go on the baked wrist for free — was implemented, measured and dropped. Three reasons, all in the table above and in `debug_weapons` captures: the baked hand is often not the trigger hand (`ar15_h` bakes only a rest hand on the receiver, `sg_h`'s rides the pump); the meshes are not life size; and `HandFrame`'s wrist is estimated by stepping back a hand's length from the far end of the point cloud, so it lands mid-palm on an open hand.

The one number that is needed is the size: the set is authored for a fixed flat camera, so `atek_h`'s weapon spans 0.66 world units (0.51 m against a real pistol's 0.20) and its baked hand is exaggerated to match. A life-size glove on it reads as a doll's hand, so the glove joins the weapon at 2.5x. That is a presentation correction for the glove only — drawing the view models at life size is the real fix and a change of its own, since it moves every grip offset, muzzle vhot and magazine anchor.

## Contact sheet

Every gun `debug_weapons` stocks, held in the right hand, arm stripped - and the psi amp keeping its own:

contact sheet, right hand

<sub>`sfg_h`'s tile is blocked by scene clutter at that framing - the known `sfg_h` capture-angle gotcha, not a render defect.</sub>

Left hand: the model mirrors and the glove mirrors with it.

contact sheet, left hand

## Tests

- `dark`: the arm classifier over every material the shipped gun set uses, and the strip over `atek_h`'s real table (both arm slots go, all three weapon slots stay).
- `vr_config`: guns / melee / amp partition the view models; the glove placement is a pure scale, so it composes inside the one `Handedness::mirror`.
- `vr-held-model.e2e`: wielded `atek_h` draws 5 objects, not 7 — verified to fail on the base ref (`expected: 5, actual: 7`) — and a `player_hands` draw lands within 0.5 units of the weapon, so the glove is on the gun rather than left at the controller.

Ran `vr-held-model`, `vr-muzzle-origin`, `vr-hand-affordance`, `vr-wrist-readouts`, `vr-eject-clip`, `vr-clip-insert-reload`, `vr-fire-mode-button`, `vr-contextual-buttons`, `vr-weapon-reload`, `vr-interface-readouts`: all green. `vr-weapon-handedness` fails identically on the base ref (pre-existing, #1262).

## Not in this slice

- **Rifles are held where their art says.** `ar15_h`/`sg_h`/`empgun_h` seat the controller at the receiver or the pump, because that is where their one baked hand is and where #1023 fitted the grip. It looked intentional while the baked rest hand was drawn; with a glove there it reads as holding a rifle by the magwell. Re-fitting those grips to the pistol grip moves the weapon relative to the hand (aim, magazine anchors), so it is its own change.
- **The view models are ~2.5x life size.** Corrected for the glove, not fixed at source.
- Two-hand / support grip, and the over-large weapons (`fsn_h`, `al_h`, `sfg_h`) whose bodies swallow the glove.
- No per-model glove offset table: the placement is now entirely the existing grip entry, and a second per-model table would only be able to decouple the glove from the weapon. The captures showed no residual error worth one.


tommy-xr:

Glove-held weapon models and removal of the authored weapon hands are covered by #1382.

PR #1381, opened 2026-09-06
feat(vr): hold a gun at life size, by its pistol grip

tommy-xr:
Slice 4b of the VR interaction pass, stacked on #1380.

A VR-wielded gun is drawn at **life size**, and the glove goes back to life size with it. #1380 stripped the baked arm and put the player's own glove on the gun, but scaled the glove up 2.5x to meet a view model authored several times life size — a doll's hand on a cannon, from the wrong end. This shrinks the gun instead, and refits every grip onto the model's actual pistol grip.

one full orbit of a life-size pistol held in the glove

<sub>One turn around a VR-wielded pistol, trigger pulled through the middle. Every side shows the same thing: the grip inside the fist, the barrel out of the hand, the gun no longer than the hand holding it. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

still

## Before → after

pistol: a 51 cm gun and a giant glove become a 20 cm gun in a real hand

Same camera, same distance. Before, the glove and the gun together overflow the frame; after, the whole pistol sits in a life-size hand. The glove is the scale reference in every capture here — it is drawn at 1.0 and is ~19 cm.

assault rifle: and the grip moves from the receiver to the pistol grip

## The scale — 0.4, from the pistol

`gun_scale` (`dev_params::GUN_WIELD_SCALE`, default **0.4**) is a uniform scale on the wielded gun. It is set from `atek_h`: its weapon geometry spans 0.663 world units — 0.51 m against a real pistol's 0.20 — so 0.4 draws it at 0.20 m. One value serves the whole set, because the set is exaggerated by about one factor (`cargo run -p shock2vr --example gun_hand_islands` prints the spans; nine of eleven view models match their own world model to within 3%):

| model | span (units) | at 0.4 | real class |
| --- | --- | --- | --- |
| `atek_h` | 0.663 | **0.20 m** | pistol |
| `lasehand` | 0.744 | 0.23 m | pistol |
| `al_h` | 1.042 | 0.32 m | — |
| `viro_h` | 2.028 | 0.62 m | — |
| `gren_h` | 2.330 | 0.71 m | grenade launcher |
| `ar15_h` | 2.360 | 0.72 m | carbine (M4 ~0.76) |
| `sg_h` | 2.419 | 0.74 m | short shotgun |
| `fsn_h` | 2.506 | 0.76 m | — |
| `empgun_h` | 2.780 | 0.85 m | rifle |
| `sfg_h` | 7.840 | — | a stray part 5 m off the model; the body is ~1.1 units |

Not 0.5 to split the difference with melee, and melee is left alone here: `melee_scale` 0.7 is a compromise between two *differently* exaggerated things (the baked arm and the weapon), which the guns no longer have — their arm is stripped. If 0.7 melee now reads oversized beside life-size guns, that is a follow-up on `melee_scale`, not a reason to fudge this one.

## How it composes — one scale about the grip

The scale has two halves and they are read from one recorded value:

- `Effect::ChangeModel` bakes `gun_wield_model_transform` (mirror × scale) into the geometry **and its muzzle vhots**, then records the scale on `RuntimePropVrGunWield`.
- `gun_wield_adjustments`, `magazine_anchor_from_entity` and `reload::clip_insert_radii` multiply the authored hand-local grip offset, the clip anchor and the clip-insert zone's radii by that same recorded scale.

Together that is a uniform scale **about the grip point**: the grip stays on the tracked hand and the muzzle stays where the barrel is, at any scale. Storing the value (rather than re-reading the dev param on each half) is what keeps them from disagreeing while the knob moves; like `melee_scale`, a change lands on the next grab.

The scale deliberately does *not* go into the entity transform: projectile velocity is `root_transform * (0,0,magnitude)`, so a scale there would slow every shot by 60%.

The zone scales for the same reason the anchor does. A life-size pistol is only ~0.26 units long; an unscaled 0.25-unit enter radius would swallow the whole weapon and every per-model anchor with it, degrading "bring the clip to the magwell" into "bring the clip near the gun". Scaled, the gesture keeps exactly the tolerance it had relative to the gun (~7.6 cm at 0.4).

## Grips — refit onto the pistol grip

#1380's grips were #1023's, fitted to where each model's *baked* hand sat. That hand is a rest hand on `ar15_h`'s receiver and rides `sg_h`'s pump — invisible while it was the hand being drawn, wrong the moment the player's glove replaced it. Each grip is now placed from the model's own side-on silhouette (added to the `gun_hand_islands` probe) and corrected against `debug_weapons` captures.

| model | grip the fist is on |
| --- | --- |
| `atek_h` | pistol grip, magazine hanging below the fist |
| `lasehand` | grip below the body |
| `ar15_h` | pistol grip, behind the magwell (was: the receiver) |
| `sg_h` | small of the stock, behind the trigger guard (was: the pump) |
| `empgun_h` | pistol grip below the receiver |
| `gren_h` | the stock behind the drum (was: inside the drum) |
| `fsn_h` | the trigger handle under the body |
| `sfg_h` | lower-rear corner of the housing — no authored grip exists |
| `al_h` | the body lump — no authored grip exists |
| `viro_h` | the stalk behind the head — no authored grip exists |

The last three (plus `gren_h`'s far end) are the oversized/organic cases: the primary hold is placed and the far end is left alone, as agreed.

## Contact sheet

Every gun `debug_weapons` stocks, right hand:

contact sheet, right hand

Left hand — the model mirrors and the grip mirrors with it:

contact sheet, left hand

<sub>Green/blue wireframes are the `clip_zone` dev overlay (the clip-insert zone and the model origin), turned on so the placement is measurable rather than guessed.</sub>

The sheet's `ar15_h` tile hides its own evidence — the fist covers the pistol grip and the magwell in front of it reads like an empty grip. Close up, with the same wield:

ar15_h close up: the pistol grip is inside the fist, the magazine is the box in front of it

Aim is unchanged by the shrink — four rounds down `debug_weapons`' 12-unit lane land on the wall, on the lane's centre line:

bullet holes on the test wall

## Tests

- `vr_config::the_gun_wield_scale_keeps_the_grip_in_the_hand_and_the_muzzle_on_the_barrel`: over every gun model and both hands, at 0.4 / 1.0 / 1.25, the authored grip point lands exactly on the tracked hand and the muzzle stays proportionally down the barrel. Verified to fail when the offset stops being scaled with the geometry.
- `reload::the_magazine_zone_needs_a_closer_approach_than_it_needs_to_hold`: now runs at both the authored size and a life-size wield, so the hysteresis survives the scaling.
- `vr-muzzle-origin.e2e`: the laser's muzzle vhot scales with the wield (read from the live `gun_scale`, not the default); the psi amp stays unscaled, which is what pins the gun/amp partition end to end.
- `vr-clip-insert-reload.e2e`: keeps its original negative — a clip carried to the mirror of the anchor across the model origin must still miss — because the zone shrank with the gun. Its radius is now read through the live `gun_scale` rather than hardcoded.
- Run green: `vr-held-model`, `vr-muzzle-origin`, `vr-clip-insert-reload`, `vr-eject-clip`, `vr-fire-mode-button`, `vr-weapon-reload`, `vr-contextual-buttons`, `vr-hand-affordance`, `vr-wrist-readouts`, plus `vr-melee-contact` as an untouched-path check.

## Known follow-up

FX spawned in the weapon's own frame — the `GunFlash`-linked muzzle flash and ejected casing (`weapon_script::create_muzzle_flash`) — are *placed* at the scaled vhot but drawn at their own authored size, so they now read relatively larger against a life-size gun. Ten consecutive single-frame captures of a VR pistol shot show no flash on screen at all, so there is nothing to see yet; scaling attached FX means plumbing a scale through `CreateEntityOptions`, which also feeds the created entity's physics, and is left as its own change.

### DIFF
--- a/shock2vr/examples/gun_hand_islands.rs
+++ b/shock2vr/examples/gun_hand_islands.rs
@@ -27,21 +27,47 @@ const GUNS: &[&str] = &[
     "viro_h", "amp_h",
 ];
 
-/// The longest axis of the geometry a mesh's polygons actually reference, in
-/// world units - the one number that says how big a model is drawn.
-fn longest_span(mesh: &ss2_bin_obj_loader::SystemShock2ObjectMesh) -> f32 {
+/// Bounds and centroid of the geometry a mesh's polygons actually reference -
+/// the one pass every readout below is derived from. `None` for a mesh whose
+/// polygons reference nothing.
+struct MeshBounds {
+    lo: [f32; 3],
+    hi: [f32; 3],
+    centroid: [f32; 3],
+}
+
+fn mesh_bounds(mesh: &ss2_bin_obj_loader::SystemShock2ObjectMesh) -> Option<MeshBounds> {
     let mut lo = [f32::INFINITY; 3];
     let mut hi = [f32::NEG_INFINITY; 3];
+    let mut sum = [0.0f32; 3];
+    let mut count = 0.0f32;
     for polygon in &mesh.polygons {
         for index in &polygon.vertex_indices {
             let vertex = mesh.vertices[*index as usize];
             for (axis, value) in [vertex.x, vertex.y, vertex.z].into_iter().enumerate() {
                 lo[axis] = lo[axis].min(value);
                 hi[axis] = hi[axis].max(value);
+                sum[axis] += value;
             }
+            count += 1.0;
         }
     }
-    (0..3).map(|axis| hi[axis] - lo[axis]).fold(0.0, f32::max)
+    (count > 0.0).then(|| MeshBounds {
+        lo,
+        hi,
+        centroid: [sum[0] / count, sum[1] / count, sum[2] / count],
+    })
+}
+
+/// The longest axis of a mesh, in world units - the one number that says how
+/// big a model is drawn.
+fn longest_span(mesh: &ss2_bin_obj_loader::SystemShock2ObjectMesh) -> f32 {
+    let Some(bounds) = mesh_bounds(mesh) else {
+        return 0.0;
+    };
+    (0..3)
+        .map(|axis| bounds.hi[axis] - bounds.lo[axis])
+        .fold(0.0, f32::max)
 }
 
 /// The mesh's XY silhouette (barrel along -X, +Y up) on a coarse grid, with the
@@ -53,8 +79,9 @@ fn print_silhouette(mesh: &ss2_bin_obj_loader::SystemShock2ObjectMesh) {
     const COLS: usize = 78;
     const ROWS: usize = 24;
 
-    let mut lo = [f32::INFINITY; 2];
-    let mut hi = [f32::NEG_INFINITY; 2];
+    let Some(MeshBounds { lo, hi, .. }) = mesh_bounds(mesh) else {
+        return;
+    };
     let mut edges = Vec::new();
     for polygon in &mesh.polygons {
         let corners = polygon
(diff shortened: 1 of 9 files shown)


tommy-xr:

Weapon grip alignment and editable uniform item scale are covered by #1382 and the Explorer workflow.

PR #1383, opened 2026-09-07
feat(vr): fingers curl until they touch what the hand holds

tommy-xr:
Slice 5: the glove's fingers curl until they touch what the hand holds.

Every pickup template is `PhysType SPHERE`, so the contact proxy is the item's **render mesh** — a new `VrContactMesh` importer hands back its triangles with the baked first-person arm stripped, and `hand_glove` transforms them into the glove's own hand space. For each finger `hand_fit` sweeps the same open→fist blend the glove already poses with, stops at the first curl where a phalanx capsule (whole chain, not just the tip) reaches a triangle, and bisects for the contact point; a `GripFamily` gives each finger a floor/cap on that blend (guns take `Trigger` by name — index on the trigger, the rest wrap — everything else is measured off the mesh's hand-space box: thin → `Pinch`, wider than the palm → `Broad`, else `Cylindrical`). The answer is cached per (model, hand, wield scale) and read back at `/v1/info` as `player.hand_affordance.{left,right}_grip`; the trigger still adds curl on top of the index, so a real pull always wins.

**The fit only runs on a model the wield actually seats** (a `HAND_MODEL_POSITIONING` entry that *moves* it, not merely turns it), and a finger the item is already inside at the open pose keeps the authored wrap. Both matter: a plain world pickup is drawn centred on the wrist, so anything measured off it describes the missing placement rather than the grip — those keep today's generic wrap until the seat data lands (slice 6).

### Contact sheet

`debug_grips` (new): one test item at a time in the right hand, at the transform the fit measured it in and the grip it solved. `grip_item` dev param cycles; framed with `/v1/camera`.

grips

contact sheet

| item | family | result |
| --- | --- | --- |
| pistol `atek_h` | trigger | **pass** — fist on the grip, index out on the trigger, ring fitted to 0.68 |
| shotgun `sg_h` | trigger | **pass** — thumb closes fully (nothing in its way), index 0.45 |
| fusion cannon `fsn_h` | trigger | **pass** — no fitted finger; the grip is inside the fist, so the wrap stands |
| worm launcher `al_h` | trigger | **pass** — pinky fitted to 0.60 on the organic body |
| coffee mug `mug` | — | **not fitted**: no seat, drawn centred on the wrist |
| magazine `magci` | — | **not fitted**: same; also renders as a wall-sized rack, not a held magazine |
| clip `ammoss` | — | **not fitted**: same (the world model is a half-metre ammo box) |
| basketball `hamball` | — | **not fitted**: same |

Honest summary: of the owner's test table only the guns are seated well enough to fit today, and three fingers across the four of them actually land on a surface — the rest of each gun's hand is the authored wrap, because a wield deliberately buries the grip inside the closed fist. The four plain pickups cannot be judged until they have a hand seat. The wrench is untouched (skinned melee rig, no glove).

### Perf

Release, `debug_grips`: **188–423 µs** per solve on 742–2106-triangle meshes. One-off per (model, hand, scale) — the cache stores the miss too, so a held item costs nothing per frame. Triangles are prefiltered by each finger's swept AABB and the coarse sweep is posed once.

### Tests

`hand_fit` unit tests on synthetic meshes (negative-first: the ungoverned fist closes through a handle, the fitted grip does not), family auto-pick from the box, the seated-through-the-hand case, cache key, trigger composition. `vr-finger-fit.e2e.test.ts` asserts a held pistol reports the trigger family with the index trailing the wrapping fingers. Ran `vr-finger-fit`, `vr-held-model`, `vr-hand-affordance`, `vr-muzzle-origin` (11 pass); full unit suite green; `RUSTFLAGS="-D warnings"` clean.

Follow-up, not done here: the geometry helpers duplicate what `parry` already does (`SharedShape::trimesh` + `query::contact`, as `debug_hitbox` uses for capsules) — worth replacing once the fit has settled.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -489,6 +489,7 @@ For debugging visual/rendering changes without a full interactive session:
    | `debug_ragdoll`          | Test ragdoll physics                         |
    | `debug_hitbox`           | View fitted hitbox shapes vs ragdoll colliders across poses |
    | `debug_gloves`           | Test VR hand/glove rendering                 |
+   | `debug_grips`            | One held test item at a time in the VR glove, posed by the finger fit (`grip_item` dev param) |
    | `debug_hand_poses`       | 25AE authored hand poses, anchored at a common hand origin |
    | `debug_teleport`         | Test VR teleport locomotion                  |
    | `debug_joint_constraint` | Test physics joint constraints               |
--- a/dark/src/importers/model_importer.rs
+++ b/dark/src/importers/model_importer.rs
@@ -371,6 +371,39 @@ pub static VR_HELD_GUN_MODELS_IMPORTER: Lazy<
     AssetImporter<SystemShockContentModel, VrHeldGunModel, ()>,
 > = Lazy::new(|| AssetImporter::define(load_model, process_vr_held_gun_model));
 
+/// The triangles of a held item's render mesh, in model space, with any baked
+/// first-person arm stripped.
+///
+/// The contact proxy the VR finger fit closes the glove against: every pickup
+/// template is `PhysType SPHERE`, so the physics collider says nothing about
+/// the shape of a mug, a magazine or a pistol grip - only the drawn surface
+/// does. Stripping the arm matches what a VR wield actually draws
+/// ([`VrHeldGunModel`]); leaving it in would stop the fingers on a hand that
+/// is not there.
+///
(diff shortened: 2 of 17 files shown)


tommy-xr:

Prepared contact fitting is covered by #1375, with per-model authoring and preview in #1379. We are retaining that fitting pipeline.

PR #1385, opened 2026-09-07
feat(vr): seat a plain pickup in the hand, and author every grip as data

tommy-xr:
Slice 6: a plain pickup gets a seat in the hand, and where every model sits becomes authored data.

`HAND_MODEL_POSITIONING` is gone. `assets/vr_grips.json` replaces it, keyed by lowercased model name; a unit test captures the old table's 28 entries and asserts every one still resolves to the same offset and turn. Anything the file leaves out is **measured off the model's own render-mesh box** (`hand_seat.rs`): the longest axis lies across the palm, the shortest faces it, and the surface rests on the palm skin - so the four test pickups, which #1383 could not fit because they were drawn centred on the wrist, are now in the hand and the fingers have something to close on.

### Profile format, in five lines

```json
"atek_h": { "rotation_deg": [0, -90, 0], "offset": [0, 0.073, 0.02], "mirror": "flip_x" }
```

`rotation_deg` XYZ Euler applied Z·Y·X (hand space: +X thumb side, +Y out of the back of the hand, -Z along the fingers); `offset` where the model origin goes; `family` (`cylindrical`/`pinch`/`broad`/`trigger`) and `fingers` `[thumb, index, middle, ring, pinky]` override what the box and the contact fit decide; `scale` is a held-geometry-only size; `mirror` (`same`/`flip_x`) is how the left hand derives the grip. Every field is optional and resolves independently - omit `offset` and the drop is measured against your authored turn.

### Scale finding

Measured with the new `grip_measure` example (world-space span of each model, metres):

| model | span | vs real | verdict |
| --- | --- | --- | --- |
| `mug` | 0.148 x 0.135 x 0.093 | ~1.3x a mug | authored `scale: 0.65` |
| `magci` | 0.354 x 0.021 x 0.478 | ~1.7x a magazine | authored `scale: 0.57` |
| `ammoss` | 0.229 x 0.119 x **0.517** | ~5x a clip - the model is a half-metre ammo crate | authored `scale: 0.23` |
| `hamball` | 0.35 dia | ~1.45x a basketball | authored `scale: 0.69` |
| `atek_w` | 0.070 x 0.282 x 0.457 | ~2.4x a pistol | untouched |

So the #1383 sheet's "several times larger than the hand" is **not** a viewmodel exaggeration like the `_h` guns' - it is SS2's world models, which run ~1.3-2x life size across the board (and `ammoss` is simply the wrong object: a crate standing in for a clip). Same conclusion, same fix as `gun_scale`: a held-only scale composed into the held transform about the seat, applied to the render and to the contact fit alike. Physics and the dropped model are untouched, so an item does change size between the floor and the hand - the trade the campaign already accepted for guns.

### Contact sheet

`debug_grips`, framed close: palm side and thumb side per item.

grips

contact sheet

| item | seat | family | verdict |
| --- | --- | --- | --- |
| coffee mug `mug` | measured + `scale` | cylindrical | **pass** - upright, hand-sized, handle turned toward the fingers, fingers curled onto it. The fingers do not pass *through* the handle: the box cannot see it, and the contact fit stops on the outside. An authored turn read worse than the measured one, so the mug ships with a scale only. |
| magazine `magci` | measured + `scale` | pinch | **partial** - seated flat against the palm with thumb and index engaged, at a believable magazine size. It reads as carried against the palm, not pinched at its edge; an edge pinch needs an authored seat, which is follow-up work. |
| clip `ammoss` | measured + `scale` | cylindrical | **pass** - clip-sized now, sitting in the fingers. |
| basketball `hamball` | measured + `scale` | broad | **pass** - rests against the palm, fingers splayed almost open. Exactly the "bigger than the hand" branch. |
| pistol `atek_h` | profile | trigger | **pass** - unchanged from #1381/#1383. |
| shotgun `sg_h` | profile | trigger | **pass** - unchanged. |
| fusion cannon `fsn_h` | profile | trigger | **pass** - unchanged. |
| worm launcher `al_h` | profile | trigger | **pass** - unchanged. |

Six of eight pass; the magazine is honest partial, and the mug's handle is the known limit of a bounding box.

### Tuner

`GET /v1/vr/grip?model=X` reports the resolved seat, family, scale and whether it came from the file (`profile`), the model's own box (`heuristic`), or nothing yet (`unseated`). `POST /v1/vr/grip` merges fields into that model's entry and re-seats live (the fit cache is keyed on a profile generation). `POST /v1/vr/grip/save` writes the registry back to `assets/vr_grips.json` with sorted keys, and refuses while the profiles have not loaded so an empty registry can never erase the file. Keyed by **model**, not by hand - the registry is process-global, so no game-loop round trip is needed and it works in `debug_grips` where nothing is "held"; `/v1/info` gained `player.hand_affordance.{left,right}_model` so a caller can ask a hand what it holds. Mirrored in the SDK as `game.vrGrips` (`get` / `set` / `clear` / `save` / `heldModel`).

### Tests

`hand_seat` unit tests on synthetic boxes (a rod lies across the palm, a slab pinches thin-side-down, a ball's surface lands on the palm, the seating turn is never a reflection); `vr_grips` parse/save round trip, resolution precedence, mirroring, generation bump; `the_profile_file_reproduces_the_grips_it_replaced` pins the migration; `the_palm_anchor_matches_the_glove_rig` re-measures the palm constant off the shipped GLB. `vr-grip-profiles.e2e.test.ts` holds each test item in `debug_grips`, asserts a gun's seat comes from the file and the mug's from its box, nudges via `/v1/vr/grip`, checks the readout moved, and saves to a temp path so it never touches the asset. Ran `vr-grip-profiles`, `vr-finger-fit`, `vr-held-model`, `vr-hand-affordance`, `vr-muzzle-origin`, `vr-clip-insert-reload`: 16 pass. Full unit suite green (1555); `RUSTFLAGS="-D warnings"` clean.

### Notes

- `dark`'s new text importer was dropped in review: `engine::assets::text_importer::TEXT_IMPORTER` already existed.
- One unrelated line in `selection-bounds.e2e.test.ts` (`as const`) - the SDK did not typecheck on the base ref, so nothing e2e could build without it.
- Not done: an edge pinch for the magazine; a handle-aware seat for the mug; two-hand grips (slice 9).


tommy-xr:

Prepared pickup poses and editable per-model grip data are covered by #1375 and #1379, with copy/mirror support in #1390. We are retaining that resource schema and editor.

PR #1388, opened 2026-09-07
fix(vr): pickups sit against the open palm, then the fingers close on them

tommy-xr:
Slice 6b: a pickup sits against the **open** palm, and then the fingers close on it.

#1385 put plain pickups at the right size, but its sheet showed a closed fist *beside* the item. The seat was not off by a nudge — it was on the wrong axis. `hand_seat` documented hand space as "+Y out of the back of the hand, palm faces -Y" and dropped every item along -Y. Measured off the shipped glove rig, -Y runs **across the knuckles**: the palm faces roughly -X and its plane is oblique to every hand axis. So an item landed off the pinky side, the open fingers were already through it, and `hand_fit` — which by design yields nothing for a finger already in contact — handed all five back the canned wrap.

The numbers, off `debug_grips`, before and after:

| item | before | after |
| --- | --- | --- |
| `mug` | thumb 1.0 index 1.0 middle 1.0 ring 0.9 pinky 0.9 | 0.0 / 0.15 / 0.18 / 0.25 / 0.28 |
| `ammoss` | 1.0 / 1.0 / 1.0 / 0.9 / 0.9 | 0.0 / 0.29 / 0.56 / 0.57 / 0.23 |
| `magci` | 0.55 / 1.0 / 0.55 / 0.55 / 0.55 | 0.0 / 0.02 / 0.55 / 0.55 / 0.55 |
| `hamball` | 0.4 / 0.4 / 0.4 / 0.4 / 0.10 | 0.4 / 0.01 / 0.04 / 0.02 / 0.04 |

Every `1.0`/`0.9` above is `GENERIC_GRIP` echoed back — the fist in the old sheet.

### The palm frame

`PALM_ANCHOR` (one point) becomes a frame measured off the open glove's own metacarpals: `PALM_CENTRE`, `PALM_NORMAL`, `PALM_CURL_AXIS`, and `PINCH_POINT`. `hand_glove`'s `the_palm_frame_matches_the_glove_rig` re-derives all four from the shipped GLB and fails if the rig moves under them.

Two things the frame settled that guessing had got backwards:

- **No nudge toward the knuckles.** A grip "sits under the base of the fingers" — but on this rig the open fingers' first phalanges run *below* the palm plane there, so seating that way puts the item into the fingers. Centred on the palm, the mug and the clip go from a 0.01–0.2 curl to a real wrap.
- **The pinch point is not at the open pose.** Open, the thumb and index are 7 cm apart and one behind the other, so a slab landed against the index while the thumb swung past. Searched over the pair's curl range they converge to 1.3 cm at thumb 0.15 / index 0.25 — and the direction between the pads *there* is the palm normal to within a few degrees, which is what makes "thin side to the palm" the right turn rather than a guess.

A measured seat also gets the **open** hand as its fit fallback, not `GENERIC_GRIP`: the seat deliberately puts the item against the open palm, so a finger already touching has finished closing. The thumb is always that finger — it lies on whatever the palm carries — and answering it with a 0.85 curl is what put a closed thumb next to an open hand. Keyed on the family (`Trigger` is the one family nothing measures its way into), so authoring a pickup's offset in the tuner cannot silently change how its fingers land.

### Before / after

Same camera, same scene, only the seat differs.

before/after

### Contact sheet

`debug_grips`, palm-on and from the thumb. The harness itself was part of the bug: it rolled the hand by a hardcoded `Ry(90)·Rz(90)` chosen under the old assumption, which against the measured frame points the palm at the floor — so the one place this work is eyeballed was framing the back of the hand. The roll now comes from `hand_seat::palm_frame`.

grips

contact sheet

| item | family | verdict |
| --- | --- | --- |
| coffee mug `mug` | cylindrical | **pass** — in the palm, fingers curled onto the body, nothing through it. Held by the body, not the handle: see below. |
| clip `ammoss` | cylindrical | **pass** — the clearest win. Was a fist with the clip behind it; now the fingers wrap it. |
| magazine `magci` | pinch | **pass** — a real edge pinch now, held at its near corner and standing away from the hand, which is what #1385 called honest-partial. |
| basketball `hamball` | broad | **pass** — the palm on the ball, fingers splayed almost open. A hand cannot wrap a basketball and no longer pretends to. |
| belt card `scipass` | pinch | **pass** — at the fingertips, face outward, readable. #1387's "same look as the mug at the reader" is gone. |
| the four guns | trigger | **unchanged** — their seats are authored, and `the_profile_file_reproduces_the_grips_it_replaced` still pins all 28 migrated entries. |

### Not done

**The mug's handle.** The brief asked for a handle-aware seat, with an authored profile as the fallback. Neither shipped, and the reason is that `mug.bin` is a *single* sub-object and a single connected island — the handle is welded to the body, so `split_connected` returns one component and there is no sub-feature to find. Authoring it instead means inferring the handle's axis from bounding-box asymmetry alone, which is a guess I could not check cheaply. The mug is meanwhile no longer a fist beside a mug: it is cradled in the palm with the fingers on it and the handle outboard. Left as follow-up, and it needs the mesh rather than its box.

**The thumb reads open on anything the palm carries.** Geometrically it is at contact before it curls, and the sheet shows it lying along the item rather than sticking out — but it is a visible change for every unprofiled pickup and every rotation-only profile, so it is called out rather than buried.

### Tests

Unit: the palm frame is orthonormal and right-handed; a rod lies across the palm with its near face on the skin; a ball's surface lands on the palm; a slab stands on edge with its mid-plane on the pinch line; the seating turn never reflects, per family; only `Trigger` falls back on the canned wrap. Plus `a_seated_handle_is_somewhere_every_finger_can_close_onto_it`, which fits a seated cylinder against the *real* glove rig twice with opposite fallbacks — a finger that answers the same either way measured the item — and fails on the old seat with "Index closed to 1 without stopping on the handle". 1575 pass.

e2e: `vr-finger-fit` gains a `debug_grips` case asserting the mug's and the clip's wrapping fingers each stop on their own surface (both answer 1.0/0.9 on the base ref). `vr-grip-profiles` asserted a measured seat by the sign of its hand Y, which was the bug written into a test; it now projects onto the palm normal and checks the origin clears the palm skin. Ran `vr-finger-fit`, `vr-grip-profiles`, `vr-held-model`, `vr-body-anchors`, `vr-hand-affordance`: 13 pass.

`RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime` clean.

### Review

`/xreview` (Claude + Codex). Both engines independently caught that `grip_item`'s range still stopped at the 8th item, so adding the belt card pushed `al_h` out of reach — fixed. Also applied: the harness roll above; the fallback re-keyed from the seat's provenance to the family (Codex: authoring an unchanged offset would otherwise change the pose); the rig tripwire no longer reads a missing joint as the origin; the pinch search no longer re-poses the index once per thumb sample; dead `Trigger` arms in the seat removed; the e2e's near-unfalsifiable sign test strengthened. Duplication pass found no re-implementation of the new geometry; its test-helper extraction notes span modules this PR does not otherwise touch and are left alone.

### DIFF
--- a/shock2vr/src/dev_params.rs
+++ b/shock2vr/src/dev_params.rs
@@ -245,7 +245,7 @@ dev_params! {
     GUN_WIELD_SCALE = float("gun_scale", "Gun scale", 0.4, 0.1, 1.5, 0.05),
     /// Which of `debug_grips`' test items to show in the hand. Ignored
     /// everywhere else; the scene clamps it to its own list.
-    GRIP_ITEM = float("grip_item", "Grip test item", 0.0, 0.0, 7.0, 1.0),
+    GRIP_ITEM = float("grip_item", "Grip test item", 0.0, 0.0, 8.0, 1.0),
     /// Enables the detached debug ("free") camera. This is the *gate*, not
     /// the camera's own on/off: while it is false the toggle input is not
     /// even read, so a stray `Alt+V` (or controller chord) during normal play
--- a/shock2vr/src/hand_glove.rs
+++ b/shock2vr/src/hand_glove.rs
@@ -347,7 +347,18 @@ impl GloveRenderer {
             fist: &self.fist,
             to_hand: glove_model_to_hand(),
         };
-        let amounts = hand_fit::fit(&mut rig, &mesh, family, GENERIC_GRIP);
+        // What a finger the fit can't measure falls back to, which depends on
+        // *why* it can't. An authored seat puts the item inside the closed hand
+        // on purpose (a gun's grip lives in the fist), so a finger already
+        // touching at the open pose keeps the generic wrap. A measured seat
+        // puts the item against the *open* palm on purpose, so a finger already
+        // touching it - the thumb lies on anything the palm is carrying - has
+        // finished closing, and curling it further would drive it through.
+        let resting = match grip.source {
+            crate::vr_grips::GripSource::Heuristic => FingerAmounts::default(),
+            _ => GENERIC_GRIP,
+        };
+        let amounts = hand_fit::fit(&mut rig, &mesh, family, resting);
         let solve = started.elapsed();
 
         tracing::debug!(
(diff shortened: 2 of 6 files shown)


tommy-xr:

Pickup seating is handled by the prepared contact-grip pipeline in #1375, the expanded fixtures in #1377, and the editable overrides in #1379.

PR #1389, opened 2026-09-07
feat(vr): a hip pouch that hands you a clip for the gun you hold

tommy-xr:
The right hip is now an ammo pouch. With a gun in one hand, an empty grip at the
other hip takes out a clip for that gun's **selected** ammo; carrying it to the
magazine loads the gun through the existing insert gesture, unchanged.

- `BodyAnchor::Pouch` mirrors the belt card on the opposite hip, same radius,
  disjoint from the belt and both shoulders at every stance (test extended).
- What the hip shows, what the glove pre-lights and what the grip commits to all
  read one rule: the taking hand must be empty and the **other** hand must wield
  a gun the reserve can serve. Green when a clip is there, amber when the gun is
  held but nothing compatible is left.
- The pouch draws the **top clip of the reserve stack itself**, so it is visible
  exactly when a grip would produce something; an empty pouch draws nothing and
  the amber light is what says so.
- Withdrawing never fabricates rounds. A stack a clip would empty is handed over
  as it stands (the strip withdraw, unchanged); a bigger one is split - the clip
  is minted first and the reserve debited only once it exists.
- Releasing a clip in the zone deposits it through the shoulder-stow path.
- Readout: `/v1/info` -> `player.body_frame.pouch` (position, `available`, clip
  template + rounds), plus the SDK type.

Left out, deliberately: a returned clip lands as its own backpack stack rather
than merging back into the one it came from - that is the existing stow deposit's
behaviour and rounds are conserved either way. A full hand carrying a non-clip
past the hip reads amber rather than being ignored, so the pouch says it will not
take it.

Tests: unit coverage for the zone, the availability rule and the withdrawal
accounting (a clip's worth, a short stack, a mission-local archetype's size, an
empty pouch, a debit that never goes negative); `vr-ammo-pouch.e2e.test.ts` walks
the refusal, the withdraw, the reload, the put-back and the stack-splitting
withdraw. Re-ran `vr-clip-insert-reload`, `vr-eject-clip`, `vr-body-anchors`,
`vr-hand-affordance`, `vr-grip-profiles` - all green.

ammo pouch demo

<sub>An empty hand at the right hip takes a clip out of the reserve and carries it into the gun: the hip's clip disappears, the glove lights green as it closes, and the pistol goes 0 -> 12 when the clip reaches the magazine. Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep), framed with the free camera.</sub>

| the clip on the hip | taken into the hand |
| --- | --- |
| pouch on the hip | clip in hand |

<sub>Left: the pistol wielded in the right hand, the free glove out to the side, and the reserve's top clip riding the right hip. Right: after the grip - the hip is empty, the clip is in the fingers, and the glove reads green. The LEFT hip is bare in both because the belt card only exists once a credential has been collected, and `debug_weapons` has none.</sub>

### DIFF
--- a/shock2vr/src/ammo_pouch.rs
+++ b/shock2vr/src/ammo_pouch.rs
@@ -1,0 +1,32 @@
+//! The ammo pouch: the right hip hands out a clip for the gun you are holding.
+//!
+//! Like the belt card this is **presentation** - there is no pouch object and
+//! nothing extra to save. What is drawn is the top clip of the reserve stack
+//! the pouch would actually hand over, resolved from the inventory every frame,
+//! so the hip is occupied exactly when a grip there would produce something.
+//! An empty pouch simply is not drawn, and the glove's amber light is what says
+//! the gun is held but the reserve has nothing for it.
+
+use cgmath::Matrix4;
+
+use crate::body_frame::BodyFrame;
+
+/// The clip sitting in the pouch this frame.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PouchClip {
+    /// The clip archetype the withdrawal would hand over.
+    pub template_id: i32,
+    /// Its model, so the hip shows the ammo the player actually carries.
+    pub model: String,
+    /// Rounds in that clip.
+    pub rounds: i32,
+}
+
+/// The clip riding the right hip, upright and facing the way the body does. Its
+/// size comes from the same `vr_grips` profile the hand uses, so hip and hand
+/// can never disagree about how big a clip is.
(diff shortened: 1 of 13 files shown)


tommy-xr:
Superseded by #1448 on the active Astra stack: selected-ammo withdrawal, real reserve splitting, physical reload compatibility, and pouch return are covered there, including merge-first returns and full-backpack behavior. The new front belt/pouch and rigid holster use original assets. Pouch-specific emissive/pre-light feedback remains a follow-up on that stack. Closing this alternate implementation; preserving its branch and discussion as reference.

PR #1394, opened 2026-09-07
feat(vr): a second hand grips the held weapon, and two hands aim it

tommy-xr:
A second hand can take hold of the weapon the first is holding - anywhere on
it - and the two together aim it.

- **Support grab.** An empty hand whose palm meets the other hand's held item
  latches onto it on a fresh grip edge. The test is the item's own surface: its
  render triangles for a gun or pickup, its fitted contact cuboid for a melee
  wield (the `_h` melee rigs are skinned and report no triangles). The latch is
  taken once, in the item's model space - re-picking the nearest point each
  frame would let the item crawl through the hand as the aim swung it.
- **Authored seats are a preference, not a gate.** `sg_h`'s pump, `empgun_h`'s
  foregrip and `ar15_h`'s magwell are authored in `assets/vr_grips.json`, read
  off each model's own baked arm island; a palm within 8 cm snaps to one, and
  takes that seat's grip family. The oversized weapons bake no support hand, so
  every grip on them is a free latch.
- **Two-hand aim.** The primary hand keeps position and the trigger. The
  placement rotation is the wrist turned by the shortest arc that puts *the
  point the support hand has hold of* onto that hand - so it is the identity
  while the hands are still where the one-hand pose put them, and moving the
  support hand swings the item to follow it. Aiming a fixed hand-local forward
  instead would only be right for a gun, whose barrel happens to run down the
  wrist; a wrench's shaft does not. One transform either way, so the muzzle, the
  magazine anchors and the melee drive's kinematic target all follow. Hands
  closer than 10 cm keep the last axis; attach and release ease over six frames.
- **Support pose.** The glove closes on the item placed at the point it latched,
  cached per model and grip point rounded to the centimetre.
- **What ends the grip:** opening the off-hand (tracked or not - the controller
  reports its grip through a lost pose, so the button is the reliable half), the
  primary dropping the item, or either hand being claimed elsewhere. A lost
  *pose* alone is held for the body anchors' own 12-frame window, with the aim
  axis frozen meanwhile.
- Readout: `/v1/info` `player.two_handed` and `player.two_hand`, plus SDK types.

The primary releasing still drops the item - no hand transfer in this slice.

Known, and left for the melee/aim slice that reads this state: the glove's
finger fit for the *primary* hand is solved against the item at its own seat, so
a large aim excursion turns the item under fingers that were fitted to it
upright. The item-relative solve above makes that zero at attach and only as
large as the hands make it.

## Visuals

two-hand grip demo

<sub>`debug_melee --vr`: the wrench held in one hand; the second hand grips it (the weapon does not jump - attaching in place is the identity); then the support hand travels and the weapon swings so the gripped point follows it; finally both hands swing it together.</sub>

shotgun support hand snapped to the pump

<sub>`debug_weapons --vr`: the support hand snaps to the shotgun's authored pump - `two_hand.snapped = true`, `support_point = [-0.382, 0.014, -0.052]`.</sub>

fusion cannon held two-handed

<sub>`debug_weapons --vr`: with no authored anchor the support hand latches wherever it touched the body - `two_handed = true`, `snapped = false`.</sub>

<sub>Captured headlessly via the debug runtime (`/v1/step` + `/v1/screenshot`, deterministic fixed-timestep), framed with the free camera.</sub>

### DIFF
--- a/assets/vr_grips.json
+++ b/assets/vr_grips.json
@@ -53,7 +53,17 @@
       0.123,
       -0.469
     ],
-    "mirror": "flip_x"
+    "mirror": "flip_x",
+    "support": [
+      {
+        "offset": [
+          -0.299,
+          0.025,
+          -0.008
+        ],
+        "family": "cylindrical"
+      }
+    ]
   },
   "ar15_w": {
     "rotation_deg": [
@@ -124,7 +134,17 @@
       0.083,
       -0.473
     ],
-    "mirror": "flip_x"
+    "mirror": "flip_x",
+    "support": [
+      {
+        "offset": [
+          -0.386,
+          0.019,
+          -0.035
+        ],
+        "family": "cylindrical"
+      }
+    ]
   },
   "fsn_h": {
     "rotation_deg": [
@@ -250,7 +270,17 @@
       -0.025,
       -0.693
     ],
-    "mirror": "flip_x"
+    "mirror": "flip_x",
+    "support": [
+      {
+        "offset": [
+          -0.382,
+          0.014,
+          -0.052
+        ],
+        "family": "cylindrical"
+      }
+    ]
   },
   "sg_w": {
     "rotation_deg": [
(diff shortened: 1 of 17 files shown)


tommy-xr:

Two-hand support is covered by #1384 and #1396, with editable support poses (#1386), support regions (#1400, #1404), and physical-contact handling (#1402). We are retaining that solver and authoring workflow.

Issue #1398, opened 2026-09-07
fix(save): QuickLoad in a debug scene panics on the scene name


tommy-xr:
Triggering `QuickLoad` while a `debug_*` scene is running panics the runtime:

```
thread 'main' panicked at shock2vr/src/mission/mod.rs:64:32:
mission debug_weapons not found in the mounted data
```

The quicksave records the scene's name as if it were a level file, and the load hands it to the level loader, which has no such `.mis`. Repro: `cargo dbgr --mission debug_weapons --port 8080`, then `QuickSave` + `QuickLoad` over `/v1/input/action`.

Found while writing an e2e that wanted to check body-anchor state across a save; the test had to drop that leg. Either refuse a quicksave in a debug scene, or make the load restore the scene by name.

PR #1399, opened 2026-09-07
feat(vr): a hip holster for the ready weapon

tommy-xr:
One weapon rides the dominant thigh, below the belt: open a hand full of gun or
melee there and it docks, close an empty hand there and it comes back to that
hand.

- **Anchor.** `BodyAnchor::Holster` on the dominant side, thigh height, with
  `holster_height` / `holster_lateral` / `holster_forward` / `holster_side` dev
  params to tune seated vs standing and to flip sides. A 12 cm zone, clamped to
  stay clear of the hips whatever the height knob says - the disjointness test
  fails without that clamp.
- **State.** The weapon goes into the pack exactly as a shoulder stow does, so it
  is the **same entity** with its own magazine and condition; `QuestInfo` records
  only the archetype, which is what carries the slot through save/load and level
  changes. A slot whose weapon is gone reads empty.
- **Rules.** Weapons only - guns and melee, not the psi amp, not a pickup. One
  slot: a second weapon pre-lights Amber and is refused. An *empty* holster
  claims nothing at all, like an unserved pouch, so a hand resting by the thigh
  can still grab and frob past it.
- **Worn.** The weapon's world model in an authored `holstered_deg` pose (default
  muzzle down, grip up) at its own `holstered_scale` - a separate number from the
  in-hand `scale`, because the holster wears the world model while a VR hand
  wields the already-shrunk `_h` one. Pistol, shotgun, assault rifle and wrench
  are authored; anything else takes the default.
- **Readout.** `/v1/info` `player.body_frame.holster` `{position, occupied}`, plus
  the SDK type.

Known limits, all deliberate:

- **A refused dock is an ordinary release, not a weapon stuck in the hand.** The
  brief asked for the weapon to stay in hand; a hand whose grip is open still
  holding a gun, which then falls when the player walks away, is worse VR than a
  plain drop, and the shoulder already lets a blocked release drop. The
  affordance goes Amber *before* the hand opens, which is the signal the player
  acts on.
- **The slot names a kind of weapon, not one weapon** - the price of storing an
  archetype so it survives a save. Lose the holstered pistol and the thigh
  empties; pick up another pistol later and it fills again. `last_stowed_weapon`
  has always worked this way; documented on the field.
- **No save/load leg in the e2e.** `QuickLoad` panics in every debug scene
  (#1398, filed) and `debug_weapons` is the only scene with a weapon bench, so
  the round trip is a `quest_info` unit test instead.
- **Long guns hang past the thigh.** A holstered shotgun or assault rifle reaches
  roughly knee height. Nothing clips the floor standing, and a `holster: false`
  profile flag exists to retire a model later; none is set. Six of the ten
  weapons have no authored `holstered_deg` yet and hang whichever way their own
  axes fall - the world models do not share a convention (the shotgun runs
  opposite to the pistol).

Tests: unit (zone disjointness at every stance including the negative-first
clamp, head-tilt invariance, dock/draw edges, occupied and non-weapon refusals,
the holstered pose and size, the save round trip) + `vr-hip-holster.e2e.test.ts`.
Ran the neighbouring VR suites alongside it - body anchors, ammo pouch, held
model, hand affordance, two-hand grip, grip profiles, clip insert, eject clip:
20/20 green. Cross-engine review findings addressed: holster/shoulder
exclusivity, the `scale` overload, the empty-holster dead zone, the
unconditional deposit, the same-frame double dock, and three shared helpers.

## Visuals

hip holster demo

<sub>VR right hand carries the pistol to the dominant thigh, opens - the weapon docks on the hip and the hand is empty - then closes at the hip to draw it again. Captured headlessly via the debug runtime (`debug_weapons --vr`, `/v1/step` + `/v1/screenshot`, deterministic fixed-timestep).</sub>

pistol holstered on the hip

<sub>The docked pistol on the thigh: grip up and forward, muzzle down, hand free.</sub>


tommy-xr:
Superseded by #1440 in the Astra interaction stack. The replacement adapts the shotgun/wrench worn-model axis corrections here and implements dedicated per-entity holster slots, Pack Rat second-slot capacity, retained-item refusal, and tested save/load and level transitions. Preferred-leg selection, editor controls, haptics and further worn-pose tuning are documented follow-ups in projects/astra-vr-thigh-holsters.md. Closing this alternate implementation without deleting its branch so its remaining profile work stays available.

PR #1502, opened 2026-09-11
fix(scripts): implement DelayGrenade so thrown grenades detonate

### LINKED ISSUE #1501: fix(ai): grenade hybrids damage nobody - DelayGrenade is unimplemented, so thrown grenades never detonate or despawn
## Problem

Grenade hybrids (`OG-Grenade`, template -176) damage nobody, at any range. The thrown grenades also never despawn: they pile up, slide forever, and sink through the floor.

The throw path is **not** at fault. Instrumented headlessly in `rec1.mis` with per-frame entity set-diffing:

- The AI plays real throw clips (`oggshot2`, `oggshot3`) and emits one `Grunt Grenade` (-1600) partway through each, about every 3 s.
- The projectile launches correctly and arcs at the player:

```
f0  pos [22.35, 2.85, -86.86]  vel [1.41,  7.06, -6.35]  speed 9.60   player [22.57, 3.94, -87.84]
f5  pos [22.47, 3.40, -87.38]  vel [1.41,  6.25, -6.35]  speed 9.02
f6  pos [22.46, 3.47, -87.34]  vel [-0.49, 3.89,  2.21]  speed 4.50   <-- bounces off, reverses
f24 pos [22.31, 4.19, -86.68]  vel [-0.49, 0.95,  2.21]  speed 2.46
```

Then nothing. It bounces off the player and lies there.

## Cause

`Grunt Grenade` (-1600) carries `PropCollisionType { BOUNCE | FULL_COLLISION_SOUND }` - notably **not** inheriting `SLAY_ON_IMPACT` from its parent `Monster Projectiles` (-672). It is authored to land and detonate on a fuse, not to explode on contact, and it links `Corpse -> HE Explosion (-3933)` for the blast.

The fuse is its `DelayGrenade` script, which in this codebase maps to `UnimplementedScript`:

```rust
// shock2vr/src/scripts/mod.rs:1140
"delaygrenade" => Box::new(UnimplementedScript::new(&script_name)),
```

So nothing ever slays the grenade -> no corpse -> no explosion -> no damage. Over ~2000 stepped frames (~33 s) there was not a single entity removal and no `HE Explosion` ever appeared.

## Second-order effects

Because nothing slays them, thrown grenades accumulate. Five were alive simultaneously at the end of one 33 s run, and their physics bodies are `dynamic`, `blocks_player: true`, never sleeping, with mass **1.13e-5** - effectively massless, so they slide indefinitely rather than settling, and several tunnelled through the floor:

```
58   [17.31,    2.73, -85.92]   sliding, never sleeps
839  [29.27,    1.23, -87.11]   below the floor
563  [22.56, -153.51, -86.17]   fell out of the level, still falling
```

The near-zero mass is worth a look on its own - `PropPhysAttr` on -1600 authors `mass: 1.0, density: 1.0` with a 0.03 radius, so the runtime is deriving mass from the radius somewhere rather than taking the authored value.

## Fix

Implement `DelayGrenade`: fuse on creation, then slay itself so the existing `Corpse -> HE Explosion` path spawns the blast. Check Telliamed's script reference for the authored fuse duration and whether it also reacts to impact or damage. The explosion's damage then flows through the normal corpse/stim route, so no new damage plumbing should be needed.

## Verification

Reproduce and verify with the debug runtime in `rec1.mis` - the only mission that places an `OG-Grenade`. Put the player adjacent, step, and assert player HP drops and that grenade entities are removed rather than accumulating. Worth an SDK e2e scenario given how the bug hides (a 0.5 s poll misses the projectile entirely; it takes per-frame set-diffing to see it at all).

tommy-xr:
Fixes #1501. Stacked on #1500 (independent code, but that PR is what lets a grenade hybrid engage at contact range at all).

## Problem

Grenade hybrids (`OG-Grenade`, -176) damaged nobody, at any range.

The throw path was never at fault. Instrumented per-frame in `rec1.mis`, the AI plays its throw clips (`oggshot2`, `oggshot3`) and emits a correctly arced `Grunt Grenade` (-1600) about every 3 s:

```
f0  pos [22.35, 2.85, -86.86]  vel [1.41,  7.06, -6.35]  speed 9.60   player [22.57, 3.94, -87.84]
f6  pos [22.46, 3.47, -87.34]  vel [-0.49, 3.89,  2.21]  speed 4.50   <-- bounces off, reverses
```

Then nothing, forever. The projectile is authored `BOUNCE | FULL_COLLISION_SOUND`, deliberately **not** inheriting `SLAY_ON_IMPACT` from its parent `Monster Projectiles` (-672), so the generic collision path never ends it. Ending it is `DelayGrenade`'s job - and that mapped to `UnimplementedScript`:

```rust
// shock2vr/src/scripts/mod.rs:1140
"delaygrenade" => Box::new(UnimplementedScript::new(&script_name)),
```

No slay -> no `Corpse -> HE Explosion` (-3933) -> no damage. Nothing removed them either: over ~2000 stepped frames there was not a single entity removal, five grenades were alive at once, sliding forever and sinking through the floor.

## Change

Implement `DelayGrenade`, per its documented behaviour - *dies when bumped-in to, but with delayed activation*. It arms on a short timer, then a contact slays it and the authored corpse explosion does the rest. No new damage plumbing: the blast is data.

The fuse is script state (`script_state_key`/`save_state`/`restore_state`), so a grenade in flight when the game is saved neither re-arms from scratch nor expires early.

Two judgment calls, both called out in the code:

- **Arm delay 50 ms.** Enough to clear the throw frame, well under the ~100 ms a point-blank throw takes to reach the player, so a direct hit still detonates. A grenade arriving before arming just lands and waits to be bumped - the same script behaviour a heartbeat later.
- **30 s expiry, ours and not authored.** Only entity contacts raise `Collided` (world geometry carries no entity id, `physics_events.rs:43`), so a grenade landing on the floor waits to be walked into rather than going off on landing. Without an expiry a miss lives forever - which is exactly what the bug report observed. Expiry destroys rather than detonates: no blast in an empty room long after the fight.

## Verification

**Unit** (4 new): arms then detonates on contact; does not detonate still in the hand; expires quietly without a blast; fuse survives save/restore.

**In-game** (headless debug runtime, `rec1.mis` - the only mission placing an `OG-Grenade`):

| | Before | After |
| --- | --- | --- |
| Close range (~1.5), 3 throws | player HP 30 -> 30 | player HP **30 -> 6** |
| Medium range (~9), 5 throws | player HP 30 -> 30 | player HP **30 -> 0 (death)**, 6-8 dmg/blast |
| `HE Explosion` (-3933) | never appeared | **8/8 throws**, at the grenade's last position |
| Grenades alive at once | 5 after 33 s, none ever removed | **max 1**; 8 removed by detonation, 1 by expiry |
| Forced miss | drifts forever | destroyed at dt = 30.00 s, **no** explosion |

`cargo test -p shock2vr --lib`: 1710 passed. `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`: clean. Existing `close-range-projectile` e2e (muzzle clearance / close-range damage) passes, along with `ai-alertness`, `ally-line-of-fire` and `chase-last-seen`.

No screenshots: the visible result is the shipped `HE Explosion` FX, unchanged by this PR - the fix is that it now gets spawned at all.

## Worth a maintainer's call

**Point-blank, the hybrid kills itself in 3 throws** (6 HP self-damage per blast against its own 15 HP). This is not an arming failure - the grenade clears the thrower and detonates on the *player*; the thrower is simply inside its own authored blast radius, and took zero self-damage on all four medium-range throws. Whether retail hybrids are meant to be this suicidal at contact range is a data/balance question the harness can't settle, so I've left the authored radius alone rather than inventing a self-damage exemption.

## Not covered

The "lies on the floor until something walks into it" path could not be exercised in-game: a missed grenade falls *through* the floor (it reached y = -4098 before expiring) rather than resting on it. That tunneling is pre-existing and noted in #1501 - the bodies are authored `mass: 1.0` but come out at 1.13e-5, so the runtime is deriving mass from the 0.03 radius somewhere instead of taking the authored value. Out of scope here; the expiry at least reaps them.

### DIFF
--- a/shock2vr/src/scripts/ai/ai_util.rs
+++ b/shock2vr/src/scripts/ai/ai_util.rs
@@ -575,8 +575,6 @@ pub fn fire_ranged_projectile(
 
         let transform = root_transform.0;
         let position = joint_transform.transform_point(point3(0.0, 0.0, 0.0));
-        let muzzle_transform = transform * Matrix4::from_translation((position + forward).to_vec());
-        let muzzle_position = muzzle_transform.transform_point(point3(0.0, 0.0, 0.0));
         let Some((target_entity, target_position)) = world
             .borrow::<UniqueView<PlayerInfo>>()
             .ok()
@@ -584,6 +582,19 @@ pub fn fire_ranged_projectile(
         else {
             return Effect::NoEffect;
         };
+
+        // The muzzle stands a unit ahead of the firing joint so the shot
+        // clears the shooter's own body. An AI with no melee weapon fires
+        // right down to contact range, where that unit can reach PAST the
+        // target - spawning the shot behind it, aimed back at the shooter,
+        // and a contact-fused grenade at the thrower's own feet. Keep the
+        // muzzle short of the target instead.
+        let joint_position = transform.transform_point(position);
+        let distance_to_target = (target_position - joint_position.to_vec()).magnitude();
+        let muzzle_offset = muzzle_offset_for_distance(distance_to_target);
+        let muzzle_transform =
+            transform * Matrix4::from_translation((position + forward * muzzle_offset).to_vec());
+        let muzzle_position = muzzle_transform.transform_point(point3(0.0, 0.0, 0.0));
         if !has_line_of_fire_from(
             entity_id,
             muzzle_position,
@@ -651,6 +662,22 @@ pub const MONSTER_FOV_HALF_ANGLE: f32 = 60.0;
 /// whether to keep swinging, and a swing may only connect inside it.
 pub const MELEE_ATTACK_RANGE: f32 = 8.0 / SCALE_FACTOR;
 
+/// How far ahead of the firing joint an AI's muzzle sits, so a shot clears
+/// the shooter's own body.
+const MUZZLE_CLEARANCE: f32 = 1.0;
+
+/// The muzzle is pulled back to at least this far short of the target, so a
+/// point-blank shot still spawns between shooter and target rather than
+/// behind it.
+const MUZZLE_TARGET_MARGIN: f32 = 0.25;
+
+/// How far ahead of the firing joint to put the muzzle when the target is
+/// `distance_to_target` away: the full clearance normally, pulled in to stay
+/// short of the target at point-blank range, and never behind the shooter.
+fn muzzle_offset_for_distance(distance_to_target: f32) -> f32 {
+    MUZZLE_CLEARANCE.min((distance_to_target - MUZZLE_TARGET_MARGIN).max(0.0))
+}
+
 ///
 /// Melee Contact
 ///
@@ -697,12 +724,7 @@ pub fn melee_contact_attack(world: &World, entity_id: EntityId, physics: &Physic
         return Effect::NoEffect;
     }
 
-    let Some((weapon_template_id, _)) =
-        get_first_link_with_template_and_data(world, entity_id, |link| match link {
-            Link::Weapon => Some(()),
-            _ => None,
-        })
-    else {
+    let Some(weapon_template_id) = melee_weapon_template(world, entity_id) else {
         return Effect::NoEffect;
     };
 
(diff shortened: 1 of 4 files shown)


tommy-xr:
Follow-up filed: #1506 — use the authored `AIRCProp` stand-off distances and back off instead of firing point-blank.

The point-blank self-kill noted above is, per the authored data, a positioning problem rather than a blast-radius one. `P$AIRCProp` gives each ranged creature a `minimum_distance` and `ideal_distance` (grenade hybrid: 0 / 20), and those decide where an AI is willing to stand and fight. An AI manoeuvring toward its ideal distance is not standing in its own blast, so the self-damage should mostly resolve there rather than needing a self-damage exemption here.

PR #1570, opened 2026-09-17
fix(audio): play the launch schema when an AI fires a projectile

tommy-xr:
Creature ranged fire was completely silent. `fire_ranged_projectile` — the path
hybrids, droids, midwives and monkeys use — returned only the projectile's
`CreateEntity`, while the turret path beside it (`fire_ranged_weapon`) already
pushed a `play_positional_sound`. That is why turrets are audible and creatures
are not.

The schemas were there all along; nothing queried them. Dark keys creature fire
on `event:launch` plus the **projectile** archetype's class tag, not on the
player's `event:shoot`/`weapontype` query:

| shooter | query | schema |
| --- | --- | --- |
| player shotgun | `event:shoot` + `weapontype:shotgun` + `ammotype:pellet` + `weaponmode:0` | `f_sg2` |
| shotgun hybrid | `event:launch` + `enemyammotype:ogslug` | `fire_og_shotgun` → `ogsgun` |

Nothing in `shock2vr/` emitted `event:launch`, so this was silent for *every* AI
projectile: `ogslug`, `oggren`, `midwifeshot`, `droidspark`, `droidlaserbolt`,
`droidfusionshot`, `ninjashot`, `ovshot`, `monkeyshot`, `shodanshot`.

The projectile entity does not exist yet at launch, so the tags come from
`GlobalTemplateClassTags` rather than a live `PropClassTag`. A projectile
archetype with no class tags plays nothing, as before.

## Verification

Live, on the headless debug runtime — eng1 Blue Monkey (template 781) in
`AIBehavior: RangedAttack`:

```
Playing clip: f_mkshot   tags: [event=launch, enemyammotype=monkeyshot]
```

Repro: launch `eng1.mis`, find the monkey, confirm line of sight with
`/v1/physics/raycast`, teleport the player to standoff range, send one
`Damage` message to aggro it, then step until `AIBehavior` settles on
`RangedAttack`. (Ranged AIs stay idle until damaged, and cycle
`Chase`/`BackOff` without firing if the player is too close.)

Unit test `firing_plays_the_projectile_archetype_launch_schema` asserts the
shotgun hybrid's query is `event:launch` + `enemyammotype:ogslug`. Verified
negative: stubbing the helper to `NoEffect` fails it.

`cargo fmt`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`,
and `cargo test -p shock2vr --lib` (1876 passed) are clean.

No visuals: this change is audio-only.


tommy-xr:
## e2e: no regression against `main`

The SDK e2e suite has ~57 pre-existing failures on a clean `main` (`d9fd8023`) — 780 tests,
712 pass, 57 fail — so the absolute count is not the signal.

My first run on this branch was invalidated by my own mistake (I checked out another branch
and edited sources while it was still running, so the runtimes it launched straddled two
branches). Re-running cleanly on the sibling melee branch gave 780 / 711 / 58, and every
test that differed from the `main` set was chased down individually and reproduces on `main`
without any of my changes, or passes on retry:

| test | full `main` run | `main`, isolated | branch, isolated | verdict |
| `Hydro3 Baby Arachnids can be pushed aside and hit by normal weapons` | pass | **fail** | **fail** | pre-existing, order-dependent |
| `OG-Shotgun gives ground when approached along a clear walkway` | pass | **fail** | **fail** | pre-existing, order-dependent |
| `a VR swing lands on the limb it struck, and bills the creature once` | pass | — | pass on retry | flaky |

Note `OG-Shotgun gives ground` touches ranged AI and fails on `main` too, so it is unrelated
to this change despite the overlap in subject matter.

PR #1571, opened 2026-09-17
fix(audio): play the impact schema when a flat melee swing lands

tommy-xr:
Hitting anything with a melee weapon in flat mode was silent. Damage landed
normally, so only the sound was missing.

`HeldMeleeWeapon` (`scripts/melee_weapon.rs`) is the only code that emits a
melee impact sound, and it runs off real physics contacts — it returns nothing
when the presentation is not VR. Flat melee instead resolves its hit by raycast
in `flat_melee_hit`, which returned `Effect::Send { Damage }` and nothing else.
So flat had no impact-sound path at all, for any weapon and any target.

Play the weapon's collision schema at the hit point, tagged with what was
struck — the same `play_impact_sound` call every other contact path already
makes (`internal_fast_projectile`, `internal_collision_type`, `throwing`).

This is `flat_melee_hit`, the single handler for *any* melee weapon (it is
reached from the `TRIGGER1` animation flag whenever the weapon has no
projectile links), so it covers the laser rapier and crystal shard too, not
just the wrench.

The material tag comes from whatever was hit, so the blow already sounds
different per enemy — nothing here hardcodes a sample:

| target | inherited material | schema |
| --- | --- | --- |
| hybrid (`Hybrids`) | `Material FleshTarget` | `hwrefle*` |
| protocol/maintenance droid (`Robots`) | `Material MetalTarget` | `hwremet*` |
| plain metal prop | `Material Metal` | `hmetmet*` |

## Measured before/after

Driven headlessly in `debug_melee` (wrench, both presentations):

| | vs hybrid | vs wall |
| --- | --- | --- |
| flat, before | silent | silent |
| VR | `hwrefle1/4` (flesh) | `hmetmet3` (metal) |

VR was already correct on both; flat was silent on both.

## Not in scope

A flat swing at bare world geometry stays silent: `flat_melee_hit`'s raycast
covers `ENTITIES | HITBOX | SELECTABLE` only, so a wall produces no hit to
sound. Widening the melee raycast changes what counts as a hit, so that is a
follow-up PR.

## Verification

Unit test `a_flat_melee_hit_plays_the_weapon_impact_schema` asserts the query
is `event:collision` + `weapontype:wrench` + `material:fleshtarget`. Verified
negative: stubbing the sound to `NoEffect` fails it.

New e2e `tools/shock2-sdk/test/flat-melee-impact-sound.e2e.test.ts` swings the
shipped Wrench at Earth's Training Droid through production inventory
selection and asserts the played-sound log carries a `weapontype:wrench` /
`material:metaltarget` collision sound. (The existing melee sound e2e tests
cover VR against static geometry only.)

`cargo fmt`, `RUSTFLAGS="-D warnings" cargo check -p shock2vr -p desktop_runtime -p debug_runtime`,
and `cargo test -p shock2vr --lib` (1876 passed) are clean.

No visuals: this change is audio-only.


tommy-xr:
## e2e: no regression against `main`

Full SDK e2e suite, run separately on a clean `main` (`d9fd8023`) and on this branch:

| | tests | pass | fail |
| --- | --- | --- | --- |
| `main` | 780 | 712 | 57 |
| this branch | 780 | 711 | 58 |

The suite has ~57 failures on `main` itself, so the absolute count is not the signal — the
set difference is. Three tests failed on the branch but not in the `main` run, and all three
were chased down individually:

| test | full `main` run | `main`, isolated | branch, isolated | verdict |
| `Hydro3 Baby Arachnids can be pushed aside and hit by normal weapons` | pass | **fail** | **fail** | pre-existing, order-dependent |
| `OG-Shotgun gives ground when approached along a clear walkway` | pass | **fail** | **fail** | pre-existing, order-dependent |
| `a VR swing lands on the limb it struck, and bills the creature once` | pass | — | pass on retry | flaky |

Both stubborn failures reproduce identically without this diff, so neither is caused by it.
(Both are sensitive to running alone vs. mid-suite, which is what shuffled the aggregate
counts.) The baby-arachnid one asserts flat wrench damage, so it was worth ruling out
directly rather than assuming.

